PR #5, opened 2024-07-21
[WIP] Assignments table and methods 

bailliekova:
## Description

- Item 1

## Reviewers

- Primary:
- Secondary:

## Checklist


backend/app/main.py
```diff
@@ -53,3 +56,42 @@ async def db_is_alive(session: Session = Depends(get_session)):
         raise HTTPException(
             status_code=status.HTTP_503_SERVICE_UNAVAILABLE, detail="DB is unreachable"
         )
+
+
+@app.post("/create_document")
```
raphaellaude, line 61: I: Add HTTP response status code.
```suggestion
@app.post(
    "/create_document",
    status_code=status.HTTP_201_CREATED,
    response_model=DocumentPublic # Will need to create this model
)
```

And it sounds like we will go the UDF route to create the document, with the client hitting the DB when starting a new doc.

backend/app/main.py
```diff
@@ -53,3 +56,42 @@ async def db_is_alive(session: Session = Depends(get_session)):
...
+            f"""
+            CREATE TABLE assignments_{document_id} PARTITION OF assignments
+            VALUES IN ('{document_id}')
+        """
+        )
+    )
+    return doc
+
+
+@app.post("/update_assignments")
```
raphaellaude, line 80: I: PATCH rather than POST since we're doing partial updates and (TBD) update document from `document_id` with `Assignments` as just `(geoid, zone)` tuples? 
```suggestion
@app.patch("/update_assignments/{document_id}") # TBD will we need a response model?
```
See [examples in docs](https://fastapi.tiangolo.com/tutorial/body-updates/) in case helpful!

backend/app/models.py
```diff
@@ -36,3 +36,15 @@ class GerryDBTableBase(TimeStampMixin, SQLModel):
 class GerryDBTable(GerryDBTableBase, table=True):
     uuid: str = Field(sa_column=Column(UUIDType, unique=True))
     name: str = Field(nullable=False, unique=True)
+
+
+class Document(TimeStampMixin, SQLModel):
```
raphaellaude, line 41: Great to create the Document table. We'll likely add many more attributes as we go.
```suggestion
class Document(TimeStampMixin, SQLModel, table=True):
```

raphaellaude (changes requested):
Just memorializing some of the things we briefly discussed! Super psyched to see this coming together

backend/app/models.py
```diff
@@ -36,3 +36,15 @@ class GerryDBTableBase(TimeStampMixin, SQLModel):
...
+    document_id: str | None = Field(sa_column=Column(UUIDType, unique=True))
+
+
+class Assignments(SQLModel, table=True):
+    # this is the empty parent table; not a partition itself
+    document_id: str = Field(foreign_key="document.document_id", primary_key=True)
+    geo_id: str = Field(primary_key=True)
```
raphaellaude, line 48: Also did we want to add indices to each of these?
```suggestion
    document_id: str = Field(foreign_key="document.document_id", primary_key=True, index=True)
    geo_id: str = Field(primary_key=True, index=True)
```
Or just on `geo_id` since the table is partitioned?

backend/app/sql/create_assignment_partition_sql_udf.sql
```diff
@@ -0,0 +1,51 @@
+CREATE FUNCTION create_assignment_partition_sql(document_id text)
+       RETURNS text LANGUAGE plpgsql AS $$
+BEGIN
+  RETURN format('CREATE TABLE assignments_%s PARTITION OF assignments
```
raphaellaude, line 4: I think we'll need to double quote assignments_%s since uuids have dashes in them

backend/app/models.py
bailliekova: just geoid; an index on document_id is redundant with the partition.

backend/app/main.py
```diff
@@ -53,3 +60,47 @@ async def db_is_alive(session: Session = Depends(get_session)):
...
+
+@app.post(
+    "/create_document",
+    response_model=DocumentPublic,
+    status_code=status.HTTP_201_CREATED,
+)
+async def create_document(session: Session = Depends(get_session)):
+    # To be created in the database
+    results = session.execute(text("SELECT create_document();"))
+    document_id = results.one()[0]  # should be only one row, one column of results
```
bailliekova, line 73: @raphaellaude I tried to return the whole document object in the one udf, but it was coming back with unlabeled columns so accessing it as doc.document_id was failing. Gonna keep futzing with it bc imo silly to do another trip to the database for this, but I think this is ok to merge as-is.

backend/app/main.py
```diff
@@ -53,3 +60,47 @@ async def db_is_alive(session: Session = Depends(get_session)):
...
+    if not doc.document_id:
+        raise HTTPException(
+            status_code=status.HTTP_500_INTERNAL_SERVER_ERROR,
+            detail="Document creation failed",
+        )
+
+    return doc
+
+
+@app.patch("/update_assignments")
```
raphaellaude, line 87: (O) for later but would love to figure out if it makes sense / is possible to remove the duplicate document_ids from `AssignmentsCreate` e.g. by using dynamic SQL

raphaellaude (approved):

PR #6, opened 2024-07-21
Assignment population totals

raphaellaude:
## Description

- Total pop endpoint, joining pops based on `gerrydb_table` stored in `Document`
- `PATCH` endpoint to update document, for now useful for updating the `gerrydb_table`
- Added coverage report to GHA
- Updated `create_document` to take a `gerrydb_table`. Because we haven't applied this migration in prod it's fine to just update this in place but if we're altering UDFs in the future will have to handle UDF replacement in migration.

Moved to #5:
- ~Move tests to dedicated directory~
- ~Refactor `gerrydb` schema name as constant~

## Reviewers

- Primary: @bailliekova 
- Secondary: @mariogiampieri 

## Checklist


raphaellaude:
Getting an error with the partitioned table's constraints:
```
self = <sqlalchemy.dialects.postgresql.dml.OnConflictDoUpdate object at 0x104f46c60>
constraint = None, index_elements = None, index_where = None
set_ = {'zone': Column('zone', Integer(), table=<excluded>, nullable=False)}
where = None

    def __init__(
        self,
        constraint: _OnConflictConstraintT = None,
        index_elements: _OnConflictIndexElementsT = None,
        index_where: _OnConflictIndexWhereT = None,
        set_: _OnConflictSetT = None,
        where: _OnConflictWhereT = None,
    ):
        super().__init__(
            constraint=constraint,
            index_elements=index_elements,
            index_where=index_where,
        )

        if (
            self.inferred_target_elements is None
            and self.constraint_target is None
        ):
>           raise ValueError(
                "Either constraint or index_elements, "
                "but not both, must be specified unless DO NOTHING"
            )
E           ValueError: Either constraint or index_elements, but not both, must be specified unless DO NOTHING

.venv/lib/python3.12/site-packages/sqlalchemy/dialects/postgresql/dml.py:290: ValueError
```

backend/app/main.py
```diff
@@ -53,3 +62,68 @@ async def db_is_alive(session: Session = Depends(get_session)):
...
+    return doc
+
+
+@app.patch("/update_assignments")
+async def update_assignments(
+    data: AssignmentsCreate, session: Session = Depends(get_session)
+):
+    stmt = insert(Assignments).values(data.model_dump()["assignments"])
+    stmt = stmt.on_conflict_do_update(
+        constraint=Assignments.__table__.primary_key, set_={"zone": stmt.excluded.zone}
```
bailliekova, line 107: thanks for fixing this bit with the explicit constraint!

backend/app/main.py
```diff
@@ -53,3 +62,68 @@ async def db_is_alive(session: Session = Depends(get_session)):
...
+    return results
+
+
+@app.get("/api/document/{document_id}/total_pop", response_model=list[ZonePopulation])
+async def get_total_population(
+    document_id: str, session: Session = Depends(get_session)
+):
+    stmt = text("SELECT * from get_total_population(:document_id)")
+    result = session.execute(stmt, {"document_id": document_id})
+    return [ZonePopulation(zone=zone, total_pop=pop) for zone, pop in result.fetchall()]
```
bailliekova, line 129: I like that the ZonePopulation model doesn't include the document id.

backend/app/sql/total_pop_udf.sql
```diff
@@ -0,0 +1,24 @@
...
+    sql_query TEXT;
+BEGIN
+    SELECT gerrydb_table INTO gerrydb_table_name
+    FROM document
+    WHERE document.document_id = $1;
+
+    sql_query := format('
+        SELECT
+            assignments.zone::TEXT AS zone,
+            SUM(blocks.total_pop)::BIGINT AS total_pop
```
bailliekova, line 14: (o) we should think about how we want to handle unmatched geoids- if we're going to cast nulls to zeros upstream and we aren't doing anything on the backend with the nulls, we should just do that in the sql query. If we still need nulls for geoids with missing population information, then this is fine as written.

backend/app/sql/total_pop_udf.sql
```diff
@@ -0,0 +1,24 @@
+CREATE OR REPLACE FUNCTION get_total_population(document_id UUID)
+RETURNS TABLE (zone TEXT, total_pop BIGINT) AS $$
+DECLARE
+    gerrydb_table_name TEXT;
+    sql_query TEXT;
+BEGIN
+    SELECT gerrydb_table INTO gerrydb_table_name
```
bailliekova, line 7: (o) I don't 100% understand why gerrydb tables are document-specific. Can we have a shared one that just joins directly on geoid? or is this a performance thing where each gerrydb table only has rows for the geoids in a given doc?

.github/workflows/test-backend.yml
```diff
@@ -39,6 +39,11 @@ jobs:
         with:
           python-version: "3.12"
 
+      - name: Install GDAL
```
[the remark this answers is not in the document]
raphaellaude: 😆

backend/app/sql/total_pop_udf.sql
raphaellaude: Ohhh good point – for now the right answer is probably to cast nulls as zeros in sql query.

Raises another question: do we need a constraint on inserted `geoids` that they're in the document's `gerrydb_table`? This would probably be slow / cumbersome but we probably want to catch unmatched geoids before we get here. unmatched `geoids` could come from: (1) user manually edits blob sent over wire, (2) gerrydb view lost/corrupted, (3) gerrydb updated by us

backend/app/sql/total_pop_udf.sql
raphaellaude: Not document-specific! Multiple documents might use the same gerrydb table – this just tracks which table to use for the document. We'll want to use the "pick your geography" flow on the frontend to inform which gerrydb table we assign to the document.

I set it up this way so that the MGGG folks could easily support new metrics / endpoints w/o having to worry about potentially breaking changes / migrations to a single monolithic `GerryDBTable` with all the geoids. Thoughts?

raphaellaude:
TY for merging in #5 @bailliekova ! I'll probably leave this PR open to address the null populations in the UDF then merge after you have.

backend/app/alembic/versions/48a91c29aae9_get_pop_totals_udf.py
```diff
@@ -0,0 +1,39 @@
...
+revision: str = "48a91c29aae9"
+down_revision: Union[str, None] = "556843a0c3bf"
+branch_labels: Union[str, Sequence[str], None] = None
+depends_on: Union[str, Sequence[str], None] = None
+
+
+def upgrade() -> None:
+    # ### commands auto generated by Alembic - please adjust! ###
+    op.drop_column("gerrydbtable", "id")
+    with open(f"{SQL_DIR}/total_pop_udf.sql") as f:
```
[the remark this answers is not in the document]
bailliekova: yeah, I didn't see this til after I did my script and we did the exact same thing, which makes me feel like it's an easy enough pattern. I like that we're storing the raw SQL; this feels fine to me.

bailliekova:
@raphaellaude I lost the pop total endpoint along the way in my branch, sorry, that'll need to come back in here.

backend/app/sql/total_pop_udf.sql
bailliekova: hm, where do candidate geoids come from in a document? IMO ideally this is a one-time validation we do when something changes in either the geoids we make available to paint or on gerrydb load.

backend/app/sql/total_pop_udf.sql
raphaellaude: Yeah stronger contract makes sense – rn the CLI just loads any layer in a gpkg lol

backend/app/sql/total_pop_udf.sql
raphaellaude: For now I `SUM(COALESCE(total_pop), 0))`

bailliekova (approved):

Issue #18, opened 2024-08-01
Paint map UI


mariogiampieri:
Better controls for choosing a zone, choosing a color, seeing results
- [ ] Zone selector based on color as currently
- [ ] Add / update label for zone
- [ ] View totals (numeric or graphical) for pop updates

Issue #23, opened 2024-08-04
Propagation of map styles from `colors`


mariogiampieri:
Currently, `colors.ts` has an array of hex codes that are used to generate the map ui color picker. There is a separate and verbose `ExpressionSpecification` in `layers.ts` which uses a hard-coded copy of that array. I would like to have that `ExpressionSpecification` be generated dynamically so we don't have to have two copies of those colors if/when they change over time.

PR #22, opened 2024-08-02
[draft] URL handling

mariogiampieri:
## Description

- WIP handling of `urlParams`; currently adds and updates x/y/z on the url/querystring 
- Waiting to finish until adding `document_id` + doing lookup by doc id is feasible- will then also connect x/y/z queries to map doc
- Ultimately not required for midpoint but it would be nice!

## Reviewers

- Primary: @raphaellaude 
- Secondary: @bailliekova 

## Checklist

- [ ] Added/Updated related documentation (if applicable).
- [ ] Added/Updated related unit tests (if applicable).

PR #24, opened 2024-08-04
Feat/map controls

mariogiampieri:
## Description

In order to be able to test / interact with the map:
- Map modes are available via radio cards. `Erase` is currently disabled but I added it here so we can demonstrate where it will be.
- Active zone is accessed via color button instead of dropdown. This gets us closer to the current UI.
- Map styles are updated. This isn't as dynamic as I'd like it to be but I don't think it's as important as other tasks we need for the demonstration so I opened #23 . 
- add reset map functionality. we will need to square how this is synced with the backend but will address that in #22 . 

## Reviewers

- Primary: @raphaellaude 
- Secondary:

## Checklist

image


app/src/app/constants/layers.ts
```diff
@@ -13,14 +12,59 @@ export const DEFAULT_PAINT_STYLE: ExpressionSpecification = [
...
 export const ZONE_ASSIGNMENT_STYLE: ExpressionSpecification = [
-  // based on zone feature state, set fill color
   "case",
-  ["==", ["feature-state", "zone"], 1],
-  "#0099cd",
-  ["==", ["feature-state", "zone"], 2],
-  "#ffca5d",
-  ["==", ["feature-state", "zone"], 3],
-  "#00cd99",
+  ["==", ["feature-state", "zone"], "#dd4425"],
```
raphaellaude, line 16: O/question: curious if we can get the hex code from the radix color objects in order to directly tie the palettes from `app/src/app/constants/colors.ts` to this.

app/src/app/page.tsx
```diff
@@ -1,6 +1,6 @@
 "use client";
 import { MapComponent } from "./components/Map";
-import SidebarComponent from "./components/Sidebar";
+import SidebarComponent from "./components/sidebar/Sidebar";
```
raphaellaude, line 3: good idea

app/src/app/components/sidebar/ResetMapButton.tsx
```diff
@@ -0,0 +1,18 @@
...
+
+  const handleClickResetMap = () => {
+    mapStore.setFreshMap(true);
+  };
+
+  return (
+    <button
+      onClick={handleClickResetMap}
+      className=" m-3 p-1.5 bg-white rounded-lg shadow-md"
+    >
```
raphaellaude, line 14: I: Use radix themes [Button](https://www.radix-ui.com/themes/docs/components/button) for consistency + accessibility.

app/src/app/components/sidebar/MapModeSelector.jsx
```diff
@@ -0,0 +1,66 @@
...
+  return (
+    <Box>
+      <RadioCards.Root
+        defaultValue="default"
+        value={activeTool}
+        onValueChange={handleRadioChange}
+        columns={{ initial: "1", sm: "3" }}
+      >
+        {activeTools.map((tool) => (
+          <Flex key={`${tool.mode}-flex`} direction="column" width="100%">
```
raphaellaude, line 36: I: by default `RadioCards` are flex column. Don't think you need this. See [example](https://www.radix-ui.com/themes/docs/components/radio-cards)

app/src/app/components/sidebar/MapModeSelector.jsx
```diff
@@ -0,0 +1,66 @@
...
+            <RadioCards.Item
+              value={tool.mode}
+              id={tool.mode}
+              disabled={tool.disabled}
+            >
+              {tool.label}
```
raphaellaude, line 42: O: Low effort nice to have would be adding some [icons](https://www.radix-ui.com/icons)

app/src/app/constants/layers.ts
raphaellaude, line 16: H: Don't the zones need to be integers?

app/src/app/components/sidebar/ColorPicker.jsx
```diff
@@ -0,0 +1,96 @@
...
+  if (!colorArray) return null;
+  const handleRadioChange = (value) => {
+    console.log(
+      "setting accumulated geoids to old zone",
+      selectedZone,
+      "new zone is",
+      value
+    );
+    setZoneAssignments(selectedZone, accumulatedGeoids);
+    setSelectedZone(value);
```
raphaellaude, line 33: H: We're setting the selected zone to a hex color? Strongly thing we should set the zones to integers. The colors could get reshuffled and mess this up. Worst case if you're using integers the map will just look different. And the backend is expecting integer zones.

raphaellaude (changes requested):
Only one blocking comment – we should go back to storing the zones as integers. UI is pretty–nice work!

app/src/app/constants/layers.ts
mariogiampieri: set dynamically now, much less brittle

mariogiampieri: addressed per below

app/src/app/components/sidebar/ColorPicker.jsx
raphaellaude: Any ints will work. Can do 1-indexed or 0-indexed

raphaellaude (approved):
Lgtm!

app/src/app/components/sidebar/MapModeSelector.jsx
mariogiampieri: good call, removed

app/src/app/components/sidebar/MapModeSelector.jsx
mariogiampieri: added!

mariogiampieri:
addressed those non-breaking comments above and ready for another review- talk soon!

raphaellaude (approved):
Lgtm!

mariogiampieri:
closes #18 and #23

PR #35, opened 2024-08-14
Midpoint feature additions

mariogiampieri:
## Description

- Fixes (some) problems caused by restructuring `Document` in the store, most notably sending assignments to the server. there is definitely still work to be done there but keeping this small so that doesn't block charts or otherwise sending data

## Reviewers

- Primary:
- Secondary:

## Checklist


app/src/app/components/Map.tsx
```diff
@@ -32,14 +32,19 @@ export const MapComponent: React.FC = () => {
   const hoverFeatureIds = useHoverFeatureIds();
   const document = useCreateMapDocument();
```
mariogiampieri, line 33: fyi this mutation object being conflated with the store object seems to be a big source of confusion later on. I renamed the references to the store entry as `storeDocument` but we may instead want to rename this `mutateDocument` or something

mariogiampieri:
I had been working on charts- I'm going to open a new branch against this to keep going

raphaellaude:
https://github.com/user-attachments/assets/8daf9bb3-0f77-4f5a-800d-5397e885538f

Alrighy, I have a decent number of the basics working reliably here, fetching populations from the server for now. The map reset doesn't work properly.

mariogiampieri:
things seem to work- i think we should revisit some of the most recent changes re: mutations and querying in particular but i think this is fine share as a progress update. i think technically one of you have to review

raphaellaude (approved):
got the okay from @mariogiampieri so merging!

raphaellaude:
> things seem to work- i think we should revisit some of the most recent changes re: mutations and querying in particular but i think this is fine share as a progress update. i think technically one of you have to review

I wanted to chat through this. I find that colocating the mutations and queries is much simpler than putting those in `apiHandlers`, which seems to be the tanstack-style

PR #43, opened 2024-08-16
Get total voting age population, handle missing pop col

### LINKED ISSUE #42: Update total population endpoint to return `SUM(gerrydbview.total_vap)` when `total_pop` is not present
Currently documents linked to gerrydb views with a `total_vap` (rather than `total_pop`) column error because `total_pop` is hardcoded in the `/api/document/{document_id}/total_pop` endpoint.

Hot fix this bug by updating `get_populations` UDF to pull desired column from `INFORMATION_SCHEMA`. If no population column is present, endpoint should return 404 rather than internal server error.

raphaellaude:
## Description

- Fixes #42: Update total population endpoint to return `SUM(gerrydbview.total_vap)` when total_pop is not present.
- Sends 404 instead of 500s when either the document isn't associated with a gerrydbview or if neither `total_pop` or `total_vap` are present in the gerrydb view

## Reviewers

- Primary: @bailliekova @mariogiampieri 
- Secondary:

## Checklist


raphaellaude:
Confirmed this works on all views we've received :tada:

bailliekova (approved):

PR #58, opened 2024-08-17
Clean up docs

raphaellaude:
## Description

- Just cleaning up a bit before more people start to look at the repo!

## Reviewers

- Primary: @mariogiampieri 
- Secondary: @bailliekova 

## Checklist


README.md
```diff
@@ -1,35 +1,34 @@
...
-TK
+The redesign aims to principally to address three key pain points in the Districtr application’s performance and maintainability:
+1. Slow tile rendering
+1. Cumbersome use of tiles as global state for tile rendering and most metric calculation
+1. Complexity and poor interoperability in architecture without slow copies
 
-### Pipelines
+And two key feature additions
+1. Block “shattering”
+1. A headless CMS
```
mariogiampieri, line 25: should be "2" and the CMS is out of our scope right? if so we should indicate that it is for a future phase imo

mariogiampieri (changes requested):
looks good minus the numbering issue in the base readme

README.md
raphaellaude: Will add a note that CMS is out of scope for now. As for the numbering, MD is smart and will number lists automatically so you don't need to worry about always keeping the numbers ordered.

image

mariogiampieri (approved):
looks good!

PR #63, opened 2024-08-19
Bug always render existing assignments

### LINKED ISSUE #41: Map sometimes fails to pull document assignments from backend on load
Still investigating when and why this happens. Seems to be slightly inconsistent.

raphaellaude:
## Description

- Fixes #41

## Reviewers

- Primary:
- Secondary:

## Checklist

### DIFF
--- a/app/fly.toml
+++ b/app/fly.toml
@@ -9,14 +9,14 @@ primary_region = 'ewr'
 [build]
 
 [http_service]
-  internal_port = 3000
-  force_https = true
-  auto_stop_machines = 'stop'
-  auto_start_machines = true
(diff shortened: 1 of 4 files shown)


nofurtherinformation:
@raphaellaude I'm missing some context here -- does this mean a problem with re-rendering already painted areas when painting an unrelated area/color?

If so, it is possibly worth integrating with Deck.gl that has more granular controls over [rendering update triggers](https://deck.gl/docs/api-reference/core/layer#updatetriggers)? Especially for the case of separate geometry rendering (eg. on tile load or zoom) from color rendering (eg. on zone paint)

PR #71, opened 2024-08-25
Update basemap and layers

raphaellaude:
## Description

Fixes #67 **see screenshots below**
- Add self hosted protomaps light basemap
    - Single 120GB basemap pmtiles layer loaded to R2 bucket using Cloudflare R2 migration service (from a personal S3 bucket).
    - Style generated with [protomaps/basemaps](https://github.com/protomaps/basemaps/tree/main) using the command `npm run generate-layers protomaps white`.
- Counties tileset with boundaries and labels
- Management CLI command for generating tiles

Follow-up PRs related to this one:
- Only show counties that are relevant to the project area?
- Allow turning counties layers on/off

## Reviewers

- Primary: @mariogiampieri 
- Secondary: @mduchin @bailliekova 

## Checklist

- [ ] Added/Updated related documentation (if applicable).
- [ ] Added/Updated related unit tests (if applicable).

## Screenshots (if applicable):
image
image
image

PR #72, opened 2024-08-27
County labels and painted layers toggle

raphaellaude:
## Description

- Splits sidebar into tabs following existing Districtr layout
- Adds legend (and utils) to toggle visibility of county layers and block layers
- Move basemap layers back to a ts file so that next will build without type errors. I think it loads faster anyway

> [!CAUTION]
> The PaintByCounty toggle doesn't actually make you paint by county for now. That's a slightly larger feature that should have it's own PR.

## Reviewers

- Primary: @mariogiampieri 
- Secondary: @bailliekova 

## Checklist

https://github.com/user-attachments/assets/a490bc12-05d8-4b10-8461-de81c496cd87


raphaellaude:
Will want to rebase branch once #71 is merged

app/src/app/store/mapStore.ts
```diff
@@ -69,6 +78,7 @@ export const useMapStore = create<MapStore>((set) => ({
     set((state) => {
       if (!state.mapRef) return { selectedLayer: null };
       addBlockLayers(state.mapRef, layer);
```
mariogiampieri, line 80: I'm not sure this is the issue exactly but I worry that there are collisions with the map state if changes are also coming from the zustand store. We should be using `subscribe`s for this: https://github.com/pmndrs/zustand?tab=readme-ov-file#using-subscribe-with-selector

app/src/app/components/sidebar/PaintByCounty.tsx
```diff
@@ -0,0 +1,42 @@
...
+  }));
+  const [checked, setChecked] = useState(false);
+
+  const paintByCounty = (prevIsChecked: boolean) => {
+    console.log("paintByCounty", prevIsChecked);
+    if (!mapRef || !mapRef?.current) return;
+    if (!prevIsChecked) {
+      console.log("BOX WAS NOT CHECKED, ADDING COUNTIES AS VISIBLE LAYERS");
+      COUNTY_LAYER_IDS.forEach((layerId) => {
+        mapRef.current?.setLayoutProperty(layerId, "visibility", "visible");
```
mariogiampieri, line 19: I left some other comments in `mapStore.ts`, but I get an error where there's some rendering confusion between here and the `Map` object: 
image
I put a full stack trace at 140 in `mapStore.ts`

mariogiampieri: It's worth investigating a bit more but I think we will also want to factor out some `useEffects` in `Map` regardless, but there may be a re-render happening there when I toggle this

app/src/app/constants/configuration.ts
```diff
@@ -1,14 +1,1939 @@
...
+      ],
+      "text-transform": "uppercase",
+    },
+    paint: {
+      "text-color": "#b8b8b8",
+    },
+  },
+];
+
+export const BASEMAP_STYLE: StyleSpecification = {
```
mariogiampieri, line 1917: I take it the above is the same as #71 but you were waiting for my review- apologies if that's so!

app/src/app/constants/configuration.ts
```diff
@@ -1,14 +1,1939 @@
...
+  sources: {
+    protomaps: {
+      type: "vector",
+      attribution:
+        '<a href="https://github.com/protomaps/basemaps">Protomaps</a> © <a href="https://openstreetmap.org">OpenStreetMap</a>',
+      url: `pmtiles://${process.env.NEXT_PUBLIC_S3_BUCKET_URL}/basemaps/20240325.pmtiles`,
+    },
+    counties: {
+      type: "vector",
+      url: `pmtiles://${process.env.NEXT_PUBLIC_S3_BUCKET_URL}/basemaps/tiger/tiger2023/tl_2023_us_county_full.pmtiles`,
```
mariogiampieri, line 1928: if this isn't passed in as a JSON, I think we'd want to pull out the file names as constants too instead of being hardcoded in

app/src/app/store/mapStore.ts
```diff
@@ -9,8 +9,13 @@ import {
 } from "../api/apiHandlers";
 import maplibregl from "maplibre-gl";
 import type { MutableRefObject } from "react";
-import { addBlockLayers } from "../constants/layers";
+import {
+  addBlockLayers,
```
mariogiampieri, line 13: I added some notes but I don't think we should be augmenting the map from the store- we can `subscribe` to changes and update that way but shouldn't call non-store functions inside

app/src/app/store/mapStore.ts
```diff
@@ -69,6 +78,7 @@ export const useMapStore = create<MapStore>((set) => ({
     set((state) => {
       if (!state.mapRef) return { selectedLayer: null };
       addBlockLayers(state.mapRef, layer);
+      state.addVisibleLayerIds([BLOCK_LAYER_ID, BLOCK_HOVER_LAYER_ID]);
```
mariogiampieri, line 81: I also don't love adding in these hard coded layer ids like this- shouldn't it follow the `layer` argument that's passed in somehow?

mariogiampieri (changes requested):
Per some of the notes, I was seeing a state collision issue in `PaintByCounty.tsx`.

app/src/app/store/mapStore.ts
```diff
@@ -125,4 +135,30 @@ export const useMapStore = create<MapStore>((set) => ({
   setFreshMap: (resetMap) => set({ freshMap: resetMap }),
   mapMetrics: null,
   setMapMetrics: (metrics) => set({ mapMetrics: metrics }),
+  visibleLayerIds: ["counties_boundary", "counties_labels"],
+  setVisibleLayerIds: (layerIds) => set({ visibleLayerIds: layerIds }),
+  addVisibleLayerIds: (layerIds: string[]) => {
```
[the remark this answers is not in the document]
raphaellaude: Yeah I was wondering about that. Only happens the first time

app/src/app/components/sidebar/Layers.tsx
```diff
@@ -0,0 +1,77 @@
...
+        My painted districts
+      </Heading>
+      <CheckboxGroup.Root
+        defaultValue={[]}
+        name="districts"
+        value={visibleLayerIds.includes(BLOCK_LAYER_ID) ? ["1"] : []}
+      >
+        <CheckboxGroup.Item
+          value="1"
+          onClick={() => toggleLayers([BLOCK_LAYER_ID, BLOCK_HOVER_LAYER_ID])}
```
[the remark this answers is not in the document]
raphaellaude: Do you mean the numbering of districts or the checkbox values?

app/src/app/store/mapStore.ts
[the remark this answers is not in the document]
raphaellaude: Needed to use a useEffect!

app/src/app/store/mapStore.ts
raphaellaude: Makes sense and this is much nicer.

app/src/app/components/sidebar/PaintByCounty.tsx
raphaellaude: Just was a bad implementation in PaintByCounty. I was attempting to modify state within a component and outside a useEffect 🙃

app/src/app/constants/configuration.ts
raphaellaude: Fair enough – for now if it's okay I'm going to leave it in since counties don't really change very often

app/src/app/store/mapStore.ts
raphaellaude: Makes total sense and the subscribe approach is much nicer

app/src/app/store/mapStore.ts
raphaellaude: I think it's okay for now since this is for the special block ids. Refactored into a subscribe

mariogiampieri (approved):
lgtm, nice work!

PR #82, opened 2024-09-05
Amend types in recharts PR

mapmeld:
Updates the previous recharts PR #78 with types for TypeScript build:

- label types for the custom tooltip component
- revert changes to `useMapStore` syntax - typing was too verbose
- sort `mapMetrics.data` by zone before rendering the chart
- ALSO reuse `numberFormat.format` in the custom tooltip component


raphaellaude (approved):
lgtm

PR #83, opened 2024-09-07
Eraser feature!

mapmeld:
Concepts:
- enable Eraser button
- when `activeTool === "eraser"`, we see the hover-highlight on the map
- clicking in eraser mode stores zoneAssignment = null in frontend parts of the app
- in eraser mode, we call `patchEraseAssignments` to delete rows; server confirms that zone assignment is null/None.
- `patchUpdateAssignments` has a filter to store only non-eraser areas on the server

Thoughts:
- now that I understand how `patchEraseAssignments` needs to work on the server, maybe I should rename it to delete
- old zone: null assignments should be cleared on the frontend after they've been successfully passed to the server

Screenshot 2024-09-07 at 3 37 24 PM

### DIFF
--- a/app/src/app/components/Map.tsx
+++ b/app/src/app/components/Map.tsx
@@ -20,6 +20,7 @@ import {
   FormatAssignments,
   getDocument,
   DocumentObject,
+  patchEraseAssignments,
   patchUpdateAssignments,
   AssignmentsCreate,
   getAssignments,
@@ -51,7 +52,24 @@ export const MapComponent: React.FC = () => {
     },
   });
 
+  const patchErase = useMutation({
+    mutationFn: patchEraseAssignments,
+    onMutate: () => {
+      console.log("Erasing assignments");
+    },
+    onError: (error) => {
+      console.log("Error erasing assignments: ", error);
+    },
+    onSuccess: (data: AssignmentsCreate) => {
+      console.log(
+        `Successfully erased ${data.assignments_upserted} assignments`,
+      );
+      mapMetrics.refetch();
+    },
+  });
+
   const {
+    activeTool,
     freshMap,
     zoneAssignments,
     mapDocument,
@@ -60,6 +78,7 @@ export const MapComponent: React.FC = () => {
     setMapRef,
     setMapMetrics,
   } = useMapStore((state) => ({
+    activeTool: state.activeTool,
     freshMap: state.freshMap,
     zoneAssignments: state.zoneAssignments,
     mapDocument: state.mapDocument,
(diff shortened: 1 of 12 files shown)


backend/app/main.py
```diff
@@ -131,6 +132,23 @@ async def update_document(
     return db_document
 
 
+@app.patch("/api/erase_assignments")
+async def erase_assignments(
+    data: AssignmentsCreate, session: Session = Depends(get_session)
+):
```
raphaellaude, line 138: H: Did you try using the existing `api/update_assignments` endpoint, passing nulls for the `zone`? We'll have to allow `zone` to be nullable (and possible run a migration).

```python
class AssignmentsBase(SQLModel):
    metadata = MetaData(schema=DOCUMENT_SCHEMA)
    document_id: str = Field(sa_column=Column(UUIDType, primary_key=True))
    geo_id: str = Field(primary_key=True)
    zone: int | None # <- right here!
```

If we can get away with a single endpoint I'd prefer that.

If we decide we prefer to have two separate endpoints then I would recommend we reimplement this w/o the slow python loop. Pydantic 2.0 was reimplemented in rust, which should give us really good perf when dumping the model (allowing for large document updates at a time). I think the python loop implementation will result in longer running updates, but that's just a guess.

app/src/app/utils/events/mapEvents.ts
```diff
@@ -145,7 +145,7 @@ export const handleMapMouseMove = (
 export const handleMapZoom = (
   e: MapLayerMouseEvent | MapLayerTouchEvent,
   map: MutableRefObject<Map | null>,
-  hoverFeatureIds: React.MutableRefObject<Set<string>>
+  hoverFeatureIds: React.MutableRefObject<Set<string>>,
```
raphaellaude, line 148: O: We should align on formatters! I think you're using the same one as @mariogiampieri , VS Code's prettier?

app/src/app/components/Map.tsx
```diff
@@ -173,7 +192,16 @@ export const MapComponent: React.FC = () => {
   useEffect(() => {
     if (mapLoaded && map.current && zoneAssignments.size) {
       const assignments = FormatAssignments();
-      patchUpdates.mutate(assignments);
+      if (activeTool === "brush") {
```
raphaellaude, line 195: H: Because we want the backend document to remain consistent with the map shown to the user, I'm hesitant to filter assignments sent to the backend here for the time being. 

This is another good reason to accept null zone assignments and use the same endpoint–`api/update_assignments`–for now if we can.

raphaellaude (changes requested):

raphaellaude:
Super pysched to see this feature come together @mapmeld ! Thanks for getting it started. Left a few comments above we should chat through.

backend/app/main.py
mapmeld: Sounds good, I'll create a migration around NOT NULL

app/src/app/components/Map.tsx
mapmeld: :+1:

mariogiampieri (approved):
looks good- nice work!

raphaellaude:
I'll apply the migration in prod shortly

raphaellaude:
> I'll apply the migration in prod shortly

Ran migration and confirmed working in prod

Issue #39, opened 2024-08-16
Calculate populations while painting from tiles


raphaellaude:
Currently the population stats are calculated server side. Though server-side metric calculation is intended to be the dominant paradigm for calculating metrics, population totals are a special case. They should be updated with less latency and updated as a user paints, rather than on mouse up (which we're currently doing for server-side metrics to minimize API calls).

nofurtherinformation:
Resolved in #217

PR #84, opened 2024-09-09
Shattering

### LINKED ISSUE #13: Shattering

raphaellaude:
## Todo
- [x] Add/update schema for shattering
  - [x] `DistrictrMap`
  - [x] `ParentChildEdges`
  - [x] `GerryDBTable`: remove tiles path, which should go in `DistrictrMap` now
- [x] Update management commands
  - [x] Compute intersection of two gerrydb tables and append to parent child table
  - [x] Delete parent child relationships for a gerrydb view
  - [x] Simplify command to generate tilesets
  - [x] Create separate command to join two gerrydb view tilesets into one
  - [x] Create materialized view of parent+child gerrydb tables
  - [x] Command to create districtr map
  - [ ] ~Command to patch districtr map. Optional but would be nice not to have to use `psql`~ Let's do this in a future PR
- [x] Update endpoints
  - [x] List districtr map views
  - [x] ~Update create document to take districtr map uuid instead of gerrydb table name.~ Existing endpoint will still work
  - [x] Update GET document endpoint to return tiles path again, joining from DistrictrMap instead of gerrydbtable, now
- [x] Shatter geometries endpoint
  - [x] Should get children from parent<>child table
  - [x] Update assignments
  - [x] Returning children
- [x] Frontend
  - [x] Context menu
  - [x] Button in menu for shattering polygons
  - [x] `setFilter` logic after hitting shatter endpoint
  - [x] Remember shattered parents on page reload (how to do this? Decided we should do a join on `parentchildedges` when sending assignments on page load)
  - [x] Get it to build!
- [x] Migrations
  - [x] Create migration script to create updated `districtrmap` records from existing `gerrydbtable` records 

## Description

- Fixes #13 

## Reviewers

- Primary:
- Secondary:

## Checklist


backend/app/models.py
```diff
@@ -40,15 +42,62 @@ class TimeStampMixin(SQLModel):
...
+class DistrictrMap(TimeStampMixin, SQLModel, table=True):
+    uuid: str = Field(sa_column=Column(UUIDType, unique=True, primary_key=True))
+    name: str | None = Field(nullable=False)
+    # This is intentionally not a foreign key on `GerryDBTable` because in some cases
+    # this may be the GerryDBTable but in others the pop table may be a materialized
+    # view of two GerryDBTables in the case of shatterable maps.
+    # We'll want to enforce the constraint tha the gerrydb_table_name is either in
+    # GerrydbTable.name or a materialized view of two GerryDBTables some other way.
+    gerrydb_table_name: str | None = Field(nullable=True)
+    # Null means default number of districts? Should we have a sensible default?
```
bailliekova, line 54: (O) Sensible default hard to set globally, IMO, maybe later we are more geo-aware and can default to the num of congressional districts in a state but for now I think null is fine.

backend/app/models.py
```diff
@@ -40,15 +42,62 @@ class TimeStampMixin(SQLModel):
...
+            UUIDType, ForeignKey("gerrydbtable.uuid"), default=None, nullable=True
+        )
+    )
+    # schema? will need to contrain the schema
+    # where does this go?
+    # when you create the view, pull the columns that you need
+    # we'll want discrete management steps
+
+
+class DistrictrMapCreate(BaseModel):
```
bailliekova, line 71: (Q) what does this one represent?

backend/app/models.py
```diff
@@ -40,15 +42,62 @@ class TimeStampMixin(SQLModel):
...
+
 class GerryDBTable(TimeStampMixin, SQLModel, table=True):
     uuid: str = Field(sa_column=Column(UUIDType, unique=True, primary_key=True))
+    # Must correspond to the layer name in the tileset if a
     name: str = Field(nullable=False, unique=True)
-    tiles_s3_path: str | None = Field(nullable=True)
+
+
+class ParentChildEdges(TimeStampMixin, SQLModel, table=True):
+    id: int = Field(sa_column=Column(Integer, primary_key=True, autoincrement=True))
```
bailliekova, line 87: (pp) IMO we don't need a primary key here- I'd drop the field.

raphaellaude: done

backend/app/models.py
raphaellaude: removed!

raphaellaude:
@bailliekova this is ready for an intermediate review! don't have a couple last things in place but would love to get your eyes on the mostly-done backend implementation, if possible!

So much to say about this but wanted to flag a couple things:
- I set districtrmap to have a unique `gerrydb_table_name`. This is probably _not_ what we want since we might have multiple districtr maps rely on the same underlying views (e.g. congressional vs. local maps?). I think to do this we'd need to send the districtrmap UUID to the client and remap `shatter_parents` to take the districtrmap UUID and parent geoids as params rather than the layer name and params.
- I need to test more edge cases. For now, the tests for the new features are basically just coverage tests. Not totally sure how to test the procedure since we'd need to hit the db again to pull the inserted records?

bailliekova:
What's our thought on an un-shatterable map? Is this if the "parent" is set to block groups? or a non-nesting geometry?

backend/app/sql/shatter_parent.sql
```diff
@@ -0,0 +1,50 @@
...
+        FROM (
+            SELECT a.document_id, edges.child_path, a.zone
+            FROM document.assignments a
+            INNER JOIN parentchildedges edges
+            ON edges.parent_path = a.geo_id
+            WHERE a.document_id = $1
+                AND a.geo_id = ANY(parent_geoids)
+                AND edges.districtr_map = districtr_map_uuid
+        ) child_geoids
+        ON CONFLICT (document_id, geo_id) DO UPDATE SET zone = EXCLUDED.zone
```
bailliekova, line 37: (q) is this just for safety? where do we expect a conflict?

backend/app/sql/shatter_parent.sql
```diff
@@ -0,0 +1,50 @@
...
+    WITH inserted_child_geoids AS (
+        INSERT INTO document.assignments (document_id, geo_id, zone)
+        SELECT $1, child_geoids.child_path, child_geoids.zone
+        FROM (
+            SELECT a.document_id, edges.child_path, a.zone
+            FROM document.assignments a
+            INNER JOIN parentchildedges edges
+            ON edges.parent_path = a.geo_id
+            WHERE a.document_id = $1
+                AND a.geo_id = ANY(parent_geoids)
```
bailliekova, line 34: (o) might try scooting this into the join condition- I don't have strong intuition if it'll be a a better execution plan, definitely might be.

backend/app/models.py
```diff
@@ -40,15 +42,58 @@ class TimeStampMixin(SQLModel):
...
+
+
+class GerryDBTable(TimeStampMixin, SQLModel, table=True):
+    uuid: str = Field(sa_column=Column(UUIDType, unique=True, primary_key=True))
+    # Must correspond to the layer name in the tileset
+    name: str = Field(nullable=False, unique=True)
+
+
+class ParentChildEdges(TimeStampMixin, SQLModel, table=True):
+    districtr_map: str = Field(
```
bailliekova, line 87: (i) how do we feel about partitioning this on districtr map as well?

bailliekova (commented):
This is looking really good

raphaellaude:
> What's our thought on an un-shatterable map? Is this if the "parent" is set to block groups? or a non-nesting geometry?

I set the `child_layer` in `DistrictrMap` to nullable – figured this would be a good flag to the frontend whether the map is shatterable or not. WDYT?

backend/app/models.py
raphaellaude: Probs a good idea!

backend/cli.py
```diff
@@ -73,6 +88,10 @@ def import_gerrydb_view(layer: str, gpkg: str, replace: bool, rm: bool):
             layer,  # must match layer name in gpkg
             "-lco",
             "OVERWRITE=yes",
+            "-lco",
+            "GEOMETRY_NAME=geometry",
+            "-nlt",
+            "MULTIPOLYGON",
```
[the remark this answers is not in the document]
raphaellaude: It's not changing functionality, at least in a breaking way! `ogr2ogr` is kind of annoying in that for e.g. GeoJSON's it'll load the geometry as `wkb_geometry` by defaylt, so this way we make sure the data type is correct / validated on load and the geometry column name won't come in as `geometry`, `geom`, `geography` (though technically geography is a different UDT in PostGIS, I learned recently)

backend/app/sql/shatter_parent.sql
raphaellaude: It's for safety but we _shouldn't_ see a conflict. Do you think it's actually more safe to fail loudly in this case? I was going back and forth

backend/app/sql/shatter_parent.sql
raphaellaude: I'll try running `EXPLAIN ANALYZE`!

backend/app/models.py
raphaellaude: done

app/src/app/components/ContextMenu.tsx
```diff
@@ -22,7 +28,48 @@ export const MapContextMenu: React.FC = () => {
       console.log("Error updating assignments: ", error);
     },
     onSuccess: (data) => {
```
raphaellaude, line 30: @nofurtherinformation implementing this `onSuccess` is where I left off, if you're looking for a place to pick up

raphaellaude: Note that setting the parent feature state and filter doesn't work properly. 

Additionally, we'll need to remove the shattered parents from the map store so they don't get added back in on the next paint event. That's honestly good enough reason for me not to track ALL of the assignments in the map store–which we presently do–instead just store a buffer of unsaved assignments (which should get cleared after successfully sending them to the db).

nofurtherinformation: Great. My approach here was to store the IDs of the shattered parents and children in the mapStore, and subscribe to that whenever a shatter event occurs. Then, use the `in` [filter on the parent and child layers](https://github.com/districtr/districtr-v2/pull/84/files#diff-4030c7e740e0aa657978a3a23f1e6eada92c0389829283ca4a50ffbf1f774b9bR215-R225). This could be swapped out for setFeatureState. 

There are still interactions to handle and reloading the shatters when the user returns, but, so far: 

Screenshot 2024-09-24 at 4 36 59 PM
Screenshot 2024-09-24 at 4 37 02 PM

raphaellaude:
> There are still interactions to handle and reloading the shatters when the user returns

You probably have these in mind but just making note of what I'm seeing:
- Unassigned blocks don't get shattered, just filtered. This is a backend thing. We should decide how we want to handle this. I can't see a reason why a user shouldn't be able to shatter a block then paint it and not paint, shatter, paint.
- Shattered blocks are unpaintable. This should be fairly easy, we just have to pass the parent and child IDs to the paint functions.

nofurtherinformation:
Today was a major refactor day for a couple key things. I started with enabling child layer interactivity, and tackled a couple things:
- Previous, hovered geometries were stored in a `Set` in a mutable `useRef`, but that would not support multiple map layers (parent and child). I moved the hovered IDs into state `hoverFeatures` as a list of the expected data shape for `setFeatureState` (`{source: string; sourceLayer?: string; id?: string | number}`). This way, the state changes can be more centralized, and we can perform map updates through `subscribe`
- However, these frequent state changes had performance issues. The way we were accessing parts of the zustand store caused the whole state to be pulled into the component (or a new object reference made each time?), and then destructured (`const { foo } = useMapStore(state => ({foo: state.foo}))`), instead, we need to access specific parts of the state individually (`const foo = useMapStore(state => state.foo)`). This is much less elegant, but does not result in excessive re-renders
- With that resolved, shatter, hover, and zone assignment map rendering are now set up as subscriptions to the zustand store as the source of truth, and map events are simplified to primarily interact with the state rather than change map rendering. There were some smaller refactors and convenience methods needed on the state to support this
- Last, filtering on the child layer needed to be applied to both the child `block` and `block_hover` layer, which is why interactions were not working, but are now good!

nofurtherinformation:
@raphaellaude @bailliekova I think you all have the right idea of storing the parent/child status of an assignment in the backend and passing that to the front, and keeping that in the frontend, too. I ran into this for coloring zone assignments and currently have [this is an awful hack](https://github.com/districtr/districtr-v2/blob/7bfca34c0736d8ba1449966d52c6ca5d8c18546a/app/src/app/store/mapStore.ts#L311-L314)

raphaellaude:
Awesome progress, thank you @nofurtherinformation! Overall, this is working very smoothly. Noticing one big thing and one small thing:

#### Big thing

Because we don't clear the zoneAssignments after a shatter, we are adding the shattered parent back to the assignments on the next paint event. This is easiest to see when you reload the map. After shattering:
image

After reload:
image

This is of course a big problem because our backend state is now out of sync with what the user is seeing. I think there are a couple of ways to resolve this:

- #114: Partial document updates. We could clear the zone assignments after each paint event and never get out of sync. This would, however, preclude us from calculating populations _while_ painting from the tile data (#39)
- Partial document updates but we treat the `zoneAssignments` and assignments sent to the backend as completely separate. We might separately maintain the `zoneAssignments` while painting and send `accumulatedGeoids` to the backend, clearing the buffer on success. We'd need to make sure we update the `zoneAssignments` after a shatter onSuccess to drop the parents.
- This is a yes and solution but we could also enforce a constraint on the backend that no parent can be in the assignments if their child is also assigned and vise-versa (@bailliekova thoughts on what implementing this constraint might look like?)

#### Small thing

I think the layer ids are slightly out of order. See how the grey lines show up more prominently on top of shattered blocks.

raphaellaude:
@nofurtherinformation and I chatted and decided to do wrap up the state mismatch issue by:
1. On shatter, lock map, remove the parent from `zoneAssignments` and `accumulatedGeoids` so it doesn't get sent back to the backend erroneously. Unlock the map after all state updates are done. 
1. Add a database constraint so that we don't have a parent and one or more of its children in the same assignments document. TBD how to implement this constraint

Other remaining to-dos:
1. Backend will send `parent_id` with the assignments on page load. FE will parse the returned data into the format it wants (`shatteredIds`)

nofurtherinformation:
The frontend now locks on shatter. Performance for this is logged in the console to help determine acceptability. 

There are still some of the map events that should be transitioned to subscriptions instead of useEffects and mouse event logic, which primarily needs the `@tanstack/react-query` stuff to be refactored to `@tanstack/query-core` which can be accessed in the state and through subscriptions.

nofurtherinformation:
@raphaellaude With the parents removed from the assignments, the population summaries are actually not quite working. 
- On the frontend, `accumulatedBlockPopulations` / `accumulatedGeoids`, doesn't quite work since the child geometries do not get returned from any event on the map, and may not be rendered depending on the zoom; after shatter, we should run the mapmetrics query, or send back populations from the shatter result with the list of child_ids
- On the backend, it looks like the `/total_pop` endpoint may not be working with shattered child IDs. I'm seeing paint -> shatter results in zero population reported for that geometry. I think the [total_pop_udf](https://github.com/districtr/districtr-v2/blob/53f781c61f55608fd11d736d05b29bc984f87036/backend/app/sql/total_pop_udf.sql#L8) needs to be updated to handle the child table, if present

Screenshot 2024-09-26 at 12 29 37 PM


```
// total pop result before shatter
[
    {
        "zone": 17,
        "total_pop": 58
    },
    {
        "zone": 25,
        "total_pop": 1662
    }
]
```
Screenshot 2024-09-26 at 12 29 45 PM

```
// total pop result after shatter
[
    {
        "zone": 17,
        "total_pop": 58
    },
    {
        "zone": 25,
        "total_pop": 0
    }
]
```

backend/app/sql/total_pop_udf.sql
```diff
@@ -5,8 +5,10 @@ DECLARE
     sql_query TEXT;
     total_pop_column_name TEXT;
 BEGIN
-    SELECT gerrydb_table INTO gerrydb_table_name
+    SELECT districtrmap.parent_layer INTO gerrydb_table_name
```
raphaellaude, line 8: @nofurtherinformation this is the issue. Should be `districtrmap.gerrydb_table_name INTO gerrydb_table_name` not the parent layer

raphaellaude:
> I think the [total_pop_udf](https://github.com/districtr/districtr-v2/blob/53f781c61f55608fd11d736d05b29bc984f87036/backend/app/sql/total_pop_udf.sql#L8) needs to be updated to handle the child table, if present

Good catch. We don't have to handle the child table actually, just use the materialized view rather than the parent_layer. Under the hood the materialized view will do the work for us of unioning the parent and child layers 🎉

backend/app/main.py
```diff
@@ -139,15 +137,58 @@ async def update_assignments(
...
+            Assignments.zone,
+            Assignments.document_id,
+            ParentChildEdges.parent_path,
+        )
+        .join(
+            ParentChildEdges,
+            Assignments.geo_id == ParentChildEdges.child_path,
+            isouter=True,
+        )
+        .where(Assignments.document_id == document_id)
```
raphaellaude, line 189: @nofurtherinformation this is close but once we have multiple sets of parentchildedges (one set per shatterable districtr map), we'll get a larger return set. We need to add a filter on the ParentChildEdges with the following logic:
```sql
SELECT
  doc.document_id,
  doc.geo_id,
  doc.zone,
  edges.parent_path
FROM "document.assignments_<uuid>" doc
LEFT JOIN parentchildedges edges
ON edges.geo_id = child_path
WHERE doc.document_id = :document_id
  AND edges.districtr_map = ( SELECT gerrydb_table FROM document.document WHERE document_id = :document_id )
```

nofurtherinformation: Ah yep, good catch. I forgot that these are not per-gerrydb tables / views

nofurtherinformation: This actually has a little more to it -- we need to get the DistrictrMap UUID not just the gerrydb table to filter parent child edges, so we need to snag districtrmap  (`districtrmap.gerrydb_table_name = document.gerrydb_table`) after getting the document (`document.document_id = :document_id`). 

Something like:

```sql
SELECT 
 assignments.geo_id, 
 assignments.zone, 
 assignments.document_id, 
 edges.parent_path 
FROM assignments 
LEFT OUTER JOIN edges 
  ON assignments.geo_id = edges.child_path 
JOIN districtrmap 
  ON edges.districtr_map = districtrmap.uuid 
JOIN document 
  ON assignments.document_id = document.document_id 
WHERE 
  assignments.document_id = :document_id
AND 
  districtrmap.gerrydb_table_name = document.gerrydb_table
```

Could also be subqueries, or a UDF if we wanted

raphaellaude: Looks about right – slightly updated the query. Right now I've enforced that the gerrydb_table_name is unique per districtrmap, which we probably don't want in the long term, but it simplifies the query. The query above (and implemented with the ORM) was failing a tests in BE when there was no parentchildedges

raphaellaude:
> I'm still seeing total population not correctly counting child geographies, but this is possibly user error on my end while configuring the shatterable view

This is now fixed!

raphaellaude:
@bailliekova requesting review because the BE should pretty much be done at this point, perhaps other than adding additional constraints.

backend/scripts/load_data.py
```diff
@@ -1,48 +1,17 @@
 import subprocess
-import sqlalchemy as sa
-from os import environ
-from glob import glob
+from os import environ, path
+import json  # Assuming this import is needed for the JSON operations
+import inspect
+
+currentdir = path.dirname(path.abspath(inspect.getfile(inspect.currentframe())))
```
raphaellaude, line 6: Nit: slightly prefer, pathlib's Path(__file__).parent

raphaellaude: I: And all caps for constants.

backend/scripts/load_data.py
```diff
@@ -55,12 +24,75 @@ def load_sample_data():
     Returns:
       None
     """
-    for gpkg in glob(f"{DATA_DIR}/*.gpkg"):
-        subprocess.run(["bash", "./scripts/load_gerrydb_view.sh", gpkg])
+
+    subprocess.run(["alembic", "upgrade", "head"], cwd=path.join(currentdir, ".."))
+    for view in config["gerrydb_views"]:
+        subprocess.run(
```
raphaellaude, line 30: PP: for each of these subprocess runs, you can use the util functions rather than CLI calls. The CLI is a simple wrapper around those util functions

raphaellaude:
Seeing one subtle bug here where after coming back to the page the children seem to have been lost in the state.

https://github.com/user-attachments/assets/08465832-0ac9-4974-b01a-a394493dc062

backend/scripts/load_data.py
raphaellaude: Yeah totally fine for now. That was a "PP" (personal preference) non-blocking comment. And yeah the click functions can't be called but they're simple wrappers around functions in `app/utils`, which _can_ be called directly

raphaellaude: Which is actually preferable because since you have to pass them a session, you can run them all in a single transaction, just `session.commit()` at the end. Anyway at some point (in diff PR) I can go in and make those changes

nofurtherinformation:
> Seeing one subtle bug here where after coming back to the page the children seem to have been lost in the state.
> 
>  subtle_shattering_bug.mp4

This is related to #119 -- I have a draft PR to address this issue (and enable redux devtools) in #120

mariogiampieri (changes requested):
Looks great save for the radix property, which does still allow me to click and remove parent without replacing with children. overall amazing work

mariogiampieri:
https://github.com/user-attachments/assets/3e527ace-ec31-40b2-a1e1-442161820430



I may have spoken too soon - it appears that if you shatter an unassigned block, it is not replaced by unassigned children. We should decide whether we require assignment before splitting or not and address accordingly. My sense is that we would want users to be able to shatter and then assign

backend/app/alembic/versions/5c4028ff26df_update_pop_udf.py
```diff
@@ -0,0 +1,89 @@
...
+
+
+def downgrade() -> None:
+    # ### commands auto generated by Alembic - please adjust! ###
+    op.drop_constraint(
+        "districtr_map_parent_child_edge_unique", "parentchildedges", type_="unique"
+    )
+    op.drop_constraint("districtr_map_uuid_unique", "districtrmap", type_="unique")
+    op.execute(
+        sa.text("""
```
bailliekova, line 44: oh, this is nice, I was worried about our migrations. I like this better than either having to create new files for new versions or do something very convoluted to get prior versions of functions.

backend/app/main.py
```diff
@@ -139,15 +137,61 @@ async def update_assignments(
...
-    stmt = select(Assignments).where(Assignments.document_id == document_id)
+    stmt = (
+        select(
+            Assignments.geo_id,
+            Assignments.zone,
+            Assignments.document_id,
+            ParentChildEdges.parent_path,
+        )
+        .join(Document, Assignments.document_id == Document.document_id)
+        .join(
```
bailliekova, line 185: (q) is this reasonably quick? a little nervous about having two joins just to retrieve assignments. I suppose this is only called on load, though, so it's not like we're doing it repeatedly.

backend/app/main.py
```diff
@@ -160,10 +204,17 @@ async def get_document(document_id: str, session: Session = Depends(get_session)
...
+            DistrictrMap.num_districts.label("num_districts"),  # pyright: ignore
+        )  # pyright: ignore
         .where(Document.document_id == document_id)
-        .join(GerryDBTable, Document.gerrydb_table == GerryDBTable.name, isouter=True)
+        .join(
+            DistrictrMap,
+            Document.gerrydb_table == DistrictrMap.gerrydb_table_name,
+            isouter=True,
+        )
         .limit(1)
```
bailliekova, line 218: (pp) not caused by this PR, but just noticing- limit(1) is going to prevent run time errors if we do for some reason violate uniqueness constraints, but, uh, we have uniqueness constraints on the database. I think we should be consistent-- pair result.one() with a .limit(1) each time.

bailliekova (approved):
looks great!

app/src/app/components/ContextMenu.tsx
```diff
@@ -0,0 +1,50 @@
...
+        {contextMenu.data.id && (
+          <ContextMenu.Label>
+            <Text size="1" color="gray">
+              {contextMenu.data.id}
+            </Text>
+          </ContextMenu.Label>
+        )}
+        <ContextMenu.Item
+          disabled={!mapDocument?.child_layer}
+          onClick={onClick}
```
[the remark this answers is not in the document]
nofurtherinformation: Updated, thank you!

backend/app/main.py
nofurtherinformation: This is a good question. I'm seeing ~250-300ms query execution time, ~500ms round trip locally. For UX, this happens concurrently with tile loading, which usually takes about the same time or slightly longer

nofurtherinformation:
> Screen.Recording.2024-10-14.at.3.14.47.PM.mov 
> I may have spoken too soon - it appears that if you shatter an unassigned block, it is not replaced by unassigned children. We should decide whether we require assignment before splitting or not and address accordingly. My sense is that we would want users to be able to shatter and then assign

The shattering UDF has been updated, and this should be resolved.

backend/app/main.py
raphaellaude: Agree. I barely noticed a change if any

backend/app/main.py
raphaellaude: Totally but I'd prefer to actually remove the limit and handle the runtime error, allowing us to send an informative error and catch bugs earlier. Bugs ofc not caused by erroneous db state-because we have our constraints in place-but in our queries

raphaellaude:
> The shattering UDF has been updated, and this should be resolved.

To reviewers, cc @mariogiampieri, if you're testing locally you'll need to re-run migrations or use `psql` to redefine the `backend/app/sql/shatter_parent.sql` UDF.

Here's the fix in action:

https://github.com/user-attachments/assets/e8fc33ad-92f3-4962-a9a2-512c7fd7b6d8

PR #87, opened 2024-09-11
Fixes #30 Summary stats metadata endpoint

### LINKED ISSUE #30: return metadata about available summary stats
```[tasklist]
### Tasks
- [ ] add list of summary stats when importing gerrydb view
```

bailliekova:
## Description

- Item 1

## Reviewers

- Primary:
- Secondary:

## Checklist


raphaellaude:
FYI I'm not seeing that many maps with P2-4 summary stats available. Will investigate shortly.
```
fly postgres connect -a districtr-v2-87-db
Connecting to fdaa:9:907c:a7b:10:420b:1527:2... complete
psql (15.8 (Debian 15.8-1.pgdg120+1))
Type "help" for help.

postgres=# select available_summary_stats from districtrmapl
postgres-# ^C
postgres=# ;
postgres=# select 1;
 ?column?
----------
        1
(1 row)

postgres=# select available_summary_stats from districtrmap;
ERROR:  relation "districtrmap" does not exist
LINE 1: select available_summary_stats from districtrmap;
                                            ^
postgres=# \l
postgres=# \c districtr_v2_api
You are now connected to database "districtr_v2_api" as user "postgres".
districtr_v2_api=# select available_summary_stats from districtrmap;
 available_summary_stats
-------------------------





 {P1}
 {P1}
 {P1}
 {P1}
 {P1}

 {P1}
 {P1}
 {P1}



 {P1}
 {P1}
 {P1}


 {P1}
 {P1}
(24 rows)

districtr_v2_api=# select gerrydb_table_name, available_summary_stats from districtrmap;
     gerrydb_table_name     | available_summary_stats
----------------------------+-------------------------
 ks_demo_view_census_vtd    |
 ks_demo_view_census_blocks |
 de_demo_view_census_vtd    |
 pa_demo_view_census_vtd    |
 de_demo_view_census_blocks |
 ga_block_all_basic_pops    | {P1}
 ri_vtd_p1_view             | {P1}
 co_block_p1_view           | {P1}
 ga_block_p1_view           | {P1}
 ga_vtd_p1_view             | {P1}
 ga_block_p4_view           |
 co_vtd_p1_view             | {P1}
 co_vtd_all_basic_pops      | {P1}
 ri_block_p1_view           | {P1}
 ri_vtd_p4_view             |
 co_block_p4_view           |
 ri_block_p4_view           |
 ri_block_all_basic_pops    | {P1}
 co_block_all_basic_pops    | {P1}
 ga_vtd_all_basic_pops      | {P1}
 co_vtd_p4_view             |
 ga_vtd_p4_view             |
 ri_vtd_all_basic_pops      | {P1}
 co_p1_view                 | {P1}
(24 rows)

districtr_v2_api=# \q
```

nofurtherinformation:
@raphaellaude  If I understand correctly from a Slack discussion, the share of total pops are the most important stat. I implemented a maybe dumb way of getting this from the parent_layer table, since shatterable views will double count population (blocks and vtds). What do you think of these changes?

Also, are there cases where blocks won't quite add up to the same value as VTDs (census margin of error?) or for our purposes will these always be perfect stacked up (VTD pop is the sum of all blocks, always) ?

raphaellaude:
> @raphaellaude If I understand correctly from a Slack discussion, the share of total pops are the most important stat. I implemented a maybe dumb way of getting this from the parent_layer table, since shatterable views will double count population (blocks and vtds). What do you think of these changes?

My read–while we wait on Moon's reply–is that each zone should sum to 100%.

image

^ See

As such I'd recommend we either
- Add a [`@computed_filed`](https://docs.pydantic.dev/2.9/concepts/fields/#the-computed_field-decorator) to the `P1SummaryStats` return model for the total and do the division in FE
- Compute the total by district in the FE and do the division in the FE too
- Compute the total by district in the SQL UDF and do the division in the FE

I wouldn't compute the shares (do the division) in the DB.

> Also, are there cases where blocks won't quite add up to the same value as VTDs (census margin of error?) or for our purposes will these always be perfect stacked up (VTD pop is the sum of all blocks, always) ?

Theoretically, yes – currently we assume the data from gerrydb does this but we should probably enforce sum checks when loading data.

backend/app/sql/summary_stats_p1.sql
```diff
@@ -36,7 +36,19 @@ BEGIN
...
+        SELECT 
+            ''Total'' AS zone,
+            SUM(COALESCE(other_pop, 0))::BIGINT AS other_pop,
+            SUM(COALESCE(asian_pop, 0))::BIGINT AS asian_pop,
+            SUM(COALESCE(amin_pop, 0))::BIGINT AS amin_pop,
+            SUM(COALESCE(nhpi_pop, 0))::BIGINT AS nhpi_pop,
+            SUM(COALESCE(black_pop, 0))::BIGINT AS black_pop,
+            SUM(COALESCE(white_pop, 0))::BIGINT AS white_pop
+        FROM gerrydb.%I
+    ', doc_districtrmap.gerrydb_table_name, doc_districtrmap.parent_layer);
```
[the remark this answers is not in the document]
nofurtherinformation: Doing a summary stat endpoint and keeping that in state, updating on `mapDocument.gerrydb_table`, makes sense to me!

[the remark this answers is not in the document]
nofurtherinformation: We will need these anyways for calculating the unassigned population

nofurtherinformation:
There is now a `P1TOTPOP` endpoint for summary stats, that the frontend fetches on `mapDocument.gerry_db` change. This returns the total population by P1 race categories

nofurtherinformation:
Additionally, the frontend now calculates the remaining unassigned population (count and percent), zone averages, and standard deviations. A nascent table visualizations that show +/-2 standard deviations by coloring the background of the cells is now an available option in the table:

Screenshot 2024-11-13 at 11 57 37 AM

backend/app/main.py
```diff
@@ -253,6 +258,53 @@ async def get_total_population(
...
+            "P1": (
+                text(
+                    "SELECT * from get_summary_stats_p1(:document_id) WHERE zone is not null"
+                ),
+                SummaryStatsP1,
+            ),
+            "P1TOTPOP": {
+                text("SELECT * from get_summary_stats_pop_totals(:document_id)"),
+                PopulationStatsP1,
+            },
```
raphaellaude, line 284: H: I don't think it makes sense to couple the document summaries and the pop summaries in an endpoint. Less code perhaps but I'd recommend we create a new endpoint for summaries which takes a gerrydb table name. This will have the added advantages of:
- having one less query in the UDF
- allowing us to share the "P1" id for both summary and document
- being able to get pop totals for the child layer if we ever need to do sum checks
- creating less coupling / being clearer what each endpoint does

raphaellaude: Just an example of where this coupling is confusing is `"results": [SummaryStatsModel.from_orm(row) for row in results],`; this doesn't really make sense – result totals shouldn't be iterable

nofurtherinformation: Ah, I see what you meant earlier now. Agreed and makes sense -- I can integrate this today

nofurtherinformation:
Main is merged in as of now. If you hit issues with migrations, try:

```sh
alembic downgrade dc391733e10a
alembic upgrade head
```

To take your db back to the branchpoint, then sync up with main

app/src/app/components/sidebar/Evaluation.tsx
```diff
@@ -0,0 +1,325 @@
...
+              <th className="py-2 px-4 text-left font-semibold">Zone</th>
+              {columnConfig.map((f, i) => (
+                <th className="py-2 px-4 text-right font-semibold" key={i}>
+                  {f.label}
+                </th>
+              ))}
+            </tr>
+          </thead>
+          <tbody>
+            {/* {!!(averages && showAverages) && (
```
nofurtherinformation, line 253: Q: these were some options we ended up not using -- should we clean up the file or keep these in case we'd want something like them in the future?

nofurtherinformation (approved):
Lgtm!

Issue #79, opened 2024-09-03
Upload flag for CLI


raphaellaude:
For dev set-ups, it's not necessary to upload generated tiles to S3, but it is necessary to populate the `tiles_s3_path` column in `gerrydbtable`. The following command will do that (which assumes the tilesets already exist in the S3 bucket):
```sql
UPDATE gerrydbtable SET tiles_s3_path = CONCAT('tilesets/', name, '.pmtiles');
```

This is a bit awkward as a dev experience and should at a minimum be documented. Adding a flag to make uploads to S3 conditional would at least allow devs to run all CLI commands locally.

raphaellaude:
Did this for batch tileset loads but not yet supported for single tileset uploads

fangge518:
The current CLI already does not auto-upload to S3, except for create-county-tiles when the S3 file is nonexistent, which is arguably desirable. No problem with dev experience now, so closing this issue...

PR #88, opened 2024-09-18
Docker compose config and data load script

nofurtherinformation:
This PR provides a docker compose configuration that runs the backend, frontend, and a postgis database with a one-liner `docker-compose up`. This makes environment management a lot simpler, and will help keep things consistent and easier for folks to get started.

This is also important to have set up in case we add additional postgres extensions, or a specific version, etc. 

## Description
- Provides docker-compose config that mounts relevant host folders for hot-reload
- Updates env with `LOAD_GERRY_DB_DATA` variable true/false to load in gerrydb data on `docker-compose up`
- Updates env with `GPKG_DATA_DIR` variable to point to where you have some gpkg data to load
- Adds a single gpkg to the `sample_data` folder, which we can add to if needed
- Does not interfere with production `Dockerfile` configurations or sizes 
- Related to #79, this handles the tileset column issue when running locally

## Reviewers
- Primary: @raphaellaude 
- Secondary: @bailliekova 

## Checklist
- [ ] TODO: Update docs 
- [ ] TODO: Incorporate tests into Docker-compose config (Probably a future Issue/PR)

### DIFF
--- a/backend/.env.dev
+++ b/backend/.env.dev
@@ -4,6 +4,8 @@ ENVIRONMENT=local
 PROJECT_NAME="Districtr v2 backend"
 BACKEND_CORS_ORIGINS="http://localhost,http://localhost:5173,http://localhost:3000,http://127.0.0.1:3000"
 SECRET_KEY="ASDFASDFASDF"
+LOAD_GERRY_DB_DATA=true
+GPKG_DATA_DIR=sample_data
 
 # Postgres
 DATABASE_URL=postgresql+psycopg://postgres:postgres@localhost:5432/districtr
--- a/backend/.env.dev
+++ b/backend/.env.dev
@@ -4,8 +4,6 @@ ENVIRONMENT=local
 PROJECT_NAME="Districtr v2 backend"
 BACKEND_CORS_ORIGINS="http://localhost,http://localhost:5173,http://localhost:3000,http://127.0.0.1:3000"
 SECRET_KEY="ASDFASDFASDF"
-LOAD_GERRY_DB_DATA=true
-GPKG_DATA_DIR=sample_data
 
 # Postgres
 DATABASE_URL=postgresql+psycopg://postgres:postgres@localhost:5432/districtr
--- a/backend/.env.dev
+++ b/backend/.env.dev
@@ -2,16 +2,16 @@
 DOMAIN=localhost
 ENVIRONMENT=local
 PROJECT_NAME="Districtr v2 backend"
-BACKEND_CORS_ORIGINS="http://localhost,http://localhost:5173"
-SECRET_KEY={fill-me}
+BACKEND_CORS_ORIGINS="http://localhost,http://localhost:5173,http://localhost:3000,http://127.0.0.1:3000"
+SECRET_KEY="ASDFASDFASDF"
 
 # Postgres
-DATABASE_URL=postgresql+psycopg://postgres:{fill-me}@localhost:5432/districtr
+DATABASE_URL=postgresql+psycopg://postgres:postgres@localhost:5432/districtr
 POSTGRES_SCHEME=postgresql+psycopg
 POSTGRES_USER=postgres
-POSTGRES_PASSWORD={fill-me}
+POSTGRES_PASSWORD=postgres
 POSTGRES_DB=districtr
-POSTGRES_SERVER=localhost
+POSTGRES_SERVER=db
 POSTGRES_PORT=5432
 
 # Volumes
--- a/backend/.env.dev
+++ b/backend/.env.dev
@@ -2,16 +2,16 @@
 DOMAIN=localhost
 ENVIRONMENT=local
 PROJECT_NAME="Districtr v2 backend"
-BACKEND_CORS_ORIGINS="http://localhost,http://localhost:5173,http://localhost:3000,http://127.0.0.1:3000"
-SECRET_KEY="ASDFASDFASDF"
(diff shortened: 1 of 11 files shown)


sample_data/basic_c22cea3c83e14fd79be9a69c87dd089c.gpkg
raphaellaude, line 1: I don't love checking a binary into git. What do you think about pulling a file from S3? The load gerrydb command supports s3 compatible URIs

backend/.env.production
```diff
@@ -6,6 +6,8 @@ ENVIRONMENT=production
 PROJECT_NAME="Districtr v2 backend"
 BACKEND_CORS_ORIGINS="https://districtr-v2-frontend.fly.dev,https://districtr-v2-api.fly.dev" # todo
 SECRET_KEY={fill-me}
+LOAD_GERRY_DB_DATA=false
+GPKG_DATA_DIR=sample_data
```
raphaellaude, line 10: H: Don't think we will need these in the prod env file, right?

backend/.env.dev
```diff
@@ -2,16 +2,18 @@
...
 # Postgres
-DATABASE_URL=postgresql+psycopg://postgres:{fill-me}@localhost:5432/districtr
+DATABASE_URL=postgresql+psycopg://postgres:postgres@localhost:5432/districtr
 POSTGRES_SCHEME=postgresql+psycopg
 POSTGRES_USER=postgres
-POSTGRES_PASSWORD={fill-me}
+POSTGRES_PASSWORD=postgres
 POSTGRES_DB=districtr
-POSTGRES_SERVER=localhost
+POSTGRES_SERVER=db
```
raphaellaude, line 16: H: These will vary by user. Can we set these env vars in the docker compose rather than changing them here?

docker-compose.yml
```diff
@@ -0,0 +1,59 @@
...
+    build: 
+      context: ./backend
+      dockerfile: Dockerfile.dev
+    container_name: backend
+    volumes:
+      - ./backend/app:/districtr-backend/app   # Adjust volumes to point to the backend folder
+      - ./backend/scripts:/districtr-backend/scripts  # Ensure scripts are mounted from backend directory
+      - ./sample_data:/districtr-backend/sample_data
+    env_file:
+      - ./backend/.env  # Ensure the env file is also from the backend folder
```
raphaellaude, line 12: I: Rather than using `./backend/.env` what if we used `environment:` or created a `.env.docker` or something?

backend/scripts/load_data.py
```diff
@@ -0,0 +1,62 @@
+import subprocess
+import sqlalchemy as sa
+from os import environ
+from glob import glob
+
+# Optionally, set a data directory to load in 
+data_dir = environ.get("GPKG_DATA_DIR", "sample_data")
+# flag to load data, by default, will load data
+load_data = environ.get("LOAD_GERRY_DB_DATA", "true")
```
raphaellaude, line 9: PP: all caps for constants

backend/scripts/load_data.py
```diff
@@ -0,0 +1,62 @@
...
+                sa.text("UPDATE public.gerrydbtable SET tiles_s3_path = CONCAT('tilesets/', name, '.pmtiles')")
+            )
+            updated_rows = result.rowcount
+            print(f"Successfully updated {updated_rows} rows in gerrydbtable.")
+            connection.commit()
+        except sa.exc.SQLAlchemyError as e:
+            print(f"Error updating gerrydbtable: {str(e)}")
+            connection.rollback()
+
+def load_sample_data():
```
raphaellaude, line 41: H: Looks like you're using two-space indentation from here on? Did you set up the precommits?

raphaellaude (changes requested):
Nice dev-ex addition! I think we shouldn't be prescriptive about how devs want to set up, though, so would prefer we not set any env vars in `.env.*` coupled to the `docker-compose` setup, instead setting those env vars in the `docker-compose`, as you've done for the db.

backend/.env.production
nofurtherinformation: Ah, correct. This is a mistake

backend/scripts/load_data.py
nofurtherinformation: Good catch, will do

sample_data/basic_c22cea3c83e14fd79be9a69c87dd089c.gpkg
nofurtherinformation: Gpkg removed and git history cleaned. Let's choose some sample data to document / publish at some point, but for now I think its ok to leave this open to the dev to choose data

backend/.env.dev
nofurtherinformation: Done!

docker-compose.yml
nofurtherinformation: Done!

backend/scripts/load_data.py
nofurtherinformation: Pre-commits are now running on my end. Getting these to play nicely with the monorepo structure and docker is a bit weird, but the docker-compose configuration can manually run the pre-commit

backend/scripts/load_data.py
nofurtherinformation: Done!

nofurtherinformation:
@raphaellaude This is ready for review. The `.env`s are cleaned up as well as the commit history -- thanks for your feedback on this!

raphaellaude (approved):
lgtm

Issue #92, opened 2024-09-24
Set # palette colors based on target # of zones


mariogiampieri:
i.e., show eight colors if the target is eight congressional districts, 35 if the target is 35 state senate districts, etc.

nofurtherinformation:
Resolved at some point...

Issue #81, opened 2024-09-03
Use legacy Districtr painting color palette, which has good accessibility properties


raphaellaude:
From @mapmeld 
> color palette: we did some research into color blind palette and main idea was smaller 3-5 district maps would be visually distinct. Admittedly with 10+ districts this will be a problem no matter what

mariogiampieri:
closed via #93

PR #93, opened 2024-09-24
Revert to classic districtr colors

mariogiampieri:
## Description

- Add the base set of 39 Districtr colors back from v1, closing #81 . I did not re-implement the color blending, as I agree with the note [here](https://github.com/districtr/districtr/blob/latest/src/colors.js) that too many colors does not aid comprehension. 
- Right now, the user has access to all colors, however we will want to limit color options based on target number of zones #92 

## Reviewers

- Primary: @raphaellaude 
- Secondary:

## Checklist

image


app/src/app/components/sidebar/ColorPicker.jsx
```diff
@@ -22,7 +22,7 @@ export function ColorPicker() {
     accumulatedGeoids: state.accumulatedGeoids,
     resetAccumulatedBlockPopulations: state.resetAccumulatedBlockPopulations,
   }));
-  const colorArray = color10;
+  const colorArray = _colorScheme;
```
raphaellaude, line 25: PP: Why the leading underscore?

app/src/app/components/sidebar/charts/HorizontalBarChart.tsx
```diff
@@ -58,7 +58,8 @@ export const HorizontalBar = () => {
       </Heading>
       <ResponsiveContainer
         width="100%"
-        height={color10.length * 18}
+        // should this instead be set based on the target number of zones? see https://github.com/districtr/districtr-v2/issues/92
+        height={_colorScheme.length * 18}
```
raphaellaude, line 62: This works for now IMO

app/src/app/constants/colors.ts
```diff
@@ -27,7 +27,7 @@ import {
   sky,
 } from "@radix-ui/colors";
 
-export const palette = {
+const palette = {
```
raphaellaude, line 30: PP: If this palette is no longer being used I would remove it. We can always grab it from the git history if we want it later.

raphaellaude (dismissed):
Lgtm. I think we should handle the target number of districts after shattering goes through, which has the updated data model. The updated data model adds `num_districts` to the MapDocument object.

raphaellaude (approved):
lgtm

Issue #69, opened 2024-08-24
Undo and redo


raphaellaude:

bailliekova:
Right now, we send the entire document to the backend, which is pretty slow on large documents. 

We designed for sending deltas. This would be really helpful for undo/redo. However, when we start sending partial updates, we need to deal with enforcing consistency between DB and frontend. 

Need to be really intentional about how we modify the assignments table. Here lie dragons.

bailliekova:
Slowness: anecdotally, times of >10 secs, less than a couple seconds on full block maps.

nofurtherinformation:
Resolved in #221

PR #118, opened 2024-10-07
Undo/redo actions support

mapmeld:
## Description
- Adds undo and redo buttons with radix-icons for #69 

Screenshot 2024-10-06 at 11 39 57 PM

- There is an undo stack (`recentZoneAssignments`, to reverse the last mousedown -> mouseup draw action. 
- Once you undo an action, the "Redo" button is enabled and it reapplies draw actions
- Because undo/redo use `setFeatureState` and `setZoneAssignments`, they update the map, chart, and serverside doc
- If you undo and then draw new districts, the future Redo-able actions are erased
- A new flag prevents Undo/Redo from adding its actions to the undo stack

## Remaining issues
- Better integration into drawing UI 
- `Object.keys(..).forEach(...)` was odd syntax to re-apply zone draws, could be refactored & consider issues with debouncing if we update several zones at once
- ~Limit the number of actions in the undo stack~ (**did 7**) before erasing the first actions
- ~Doesn't work with changes to zones from the original map load, because those are not in `state.zoneAssignments`~

### DIFF
--- a/app/src/app/components/sidebar/Sidebar.tsx
+++ b/app/src/app/components/sidebar/Sidebar.tsx
@@ -2,7 +2,7 @@ import { Box, Flex, Heading } from "@radix-ui/themes";
 import { MapModeSelector } from "./MapModeSelector";
 import { ColorPicker } from "./ColorPicker";
 import { ResetMapButton } from "./ResetMapButton";
-import { UndoButton } from "./UndoButton";
+import { UndoRedoButton } from "./UndoRedoButton";
 import { GerryDBViewSelector } from "./GerryDBViewSelector";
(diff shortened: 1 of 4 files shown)


mapmeld:
I set a limit of 7 undo-able actions, and removed a call to `resetZoneAssignments()` which cleared the assignments which had been loaded from a saved map.

nofurtherinformation:
Hey @mapmeld! What would you think about using this [undo/redo middleware](https://github.com/charkour/zundo) for zustand? I think this will be ultimately less code to maintain, and with the changes coming down in from `Shattering` the map rendering happens as subscriptions/side-effects of state changes which should make the app and map state easier to observe

PR #120, opened 2024-10-14
State enhancements: serializable and dev tools

nofurtherinformation:
Zustand is a generally good state management library but we hit a couple edge cases:

1. React MutableRefs of an HTML element break redux dev tools, so the current `MapStore['mapRef']` makes observing the state harder
2. On de-focus from the tab/window, the state serializes and goes to sleep, but that does not work for JS classes like `Set` or `Map` that do not serialize

To resolve this, we need to do one of the following:
1. Swap to a more robust state library
2. Refactor slightly to use vanilla JS objects and access the map ref slightly differently
3. Hack Zustand in some slightly non-standard ways

## Description
This PR takes the second approach and swaps out `Set` and `Map` for `Array`s and `Record`s (objects) repsectively. Additionally, this swaps `mapRef` to `getMapRef`, which can be set as a function that returns the map object, rather than a direct reference. 

## Reviewers
- Primary:
- Secondary:

## Checklist
- [ ] Refactor the state to no longer use set/maps
- [ ] Refactor mapRef to getMapRef
- [ ] Update all instances referring to old methods like `.size` or `.get`
- [ ] Ensure items are not being duplicated (eg. set automatically de-duplicates list)
- [ ] Sanity check performance implications
- [ ] Test interactions

### DIFF
--- a/app/src/app/components/Map.tsx
+++ b/app/src/app/components/Map.tsx
@@ -54,7 +54,7 @@ export const MapComponent: React.FC = () => {
             action.action as keyof MapLayerEventType,
             layer, // to be updated with the scale-agnostic layer id
             (e: MapLayerMouseEvent | MapLayerTouchEvent) => {
-              action.handler(e, map);
+              action.handler(e, map.current);
             }
           );
         }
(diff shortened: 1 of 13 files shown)


nofurtherinformation:
Closing - will re-open with rebased branch.

Issue #40, opened 2024-08-16
Persist list of user maps


raphaellaude:
Allow users to return to maps they were working on in a more ergonomic way than copy-pasting their `document_id` from the URL to a sticky note.

This feature could be achieved either by
- Setting an anonymous `user_id` in cookies / local storage and creating a table in the db with `(user_id, document_id)` entries, along with an endpoint to fetch those
- Storing a list of `document_id`s in local storage

nofurtherinformation:
Initial implementation of this in #127

PR #122, opened 2024-10-15
Persist user maps

nofurtherinformation:
Apropos #40 

## Description
This PR is for the feature enhancement descried in #40. The approach of this PR is to use local storage, with the intention that more in-depth future work might handle map sharing, etc.. 

## Reviewers

- Primary:
- Secondary:

## Checklist
- [x] User maps save to local storage on create
- [x] User maps can be reloaded
- [x] User maps can have custom names
- [x] User maps can be removed from local storage


nofurtherinformation:
Closing this PR to rebase

Issue #108, opened 2024-09-25
Reset function


mariogiampieri:
depends on #38

nofurtherinformation:
Closed in #126

Issue #38, opened 2024-08-16
Endpoint to drop all existing assignments from document


raphaellaude:
Once we have this we can support "reset" map functionality in the FE

PR #126, opened 2024-10-16
Reset Map

nofurtherinformation:
#38  and #108 

## Description
This PR creates an endpoint to reset all assignments when the user clicks 'reset map'

## Reviewers

- Primary:
- Secondary:

## Checklist
- [x] Reset endpoint on the API
- [x] Handler and mutation for reset
- [x] State management and handler
- [x] Map rendering: colors
- [x] Map rendering: shatter


backend/app/main.py
```diff
@@ -171,6 +171,18 @@ async def shatter_parent(
     return result
 
 
+@app.patch(
```
raphaellaude, line 174: I: Use [`app.delete`](https://sqlmodel.tiangolo.com/tutorial/fastapi/delete/)

backend/app/main.py
```diff
@@ -171,6 +171,18 @@ async def shatter_parent(
...
+    "/api/update_assignments/{document_id}/reset", status_code=status.HTTP_200_OK
+)
+async def reset_map(document_id: str, session: Session = Depends(get_session)):
+    # drop all rows in assignments
+    stmt = delete(Assignments).where(Assignments.document_id == document_id)
```
raphaellaude, line 179: I: Return number of deleted records? `.returning(Assignments.geo_id)` then `n_deleted = len(stmt.fetchall())`. Not strictly necessary, but might be useful for the FE to be able to validate that the operation was successful.

backend/app/main.py
raphaellaude, line 174: H: We were imagining implementing this by dropping the partition on assignments rather than deleting the individual records. Should be more performant and clean up not having a bunch of empty assignment partitions lying around.

raphaellaude (changes requested):
Couple small comments and one big one. Mostly requesting changes because I'd like to discuss whether we should fully delete user maps? One option is simply to soft delete and create a fresh map.

backend/app/main.py
bailliekova: +1 this is the preferred method; note that we'll want to recreate the partition also-- we can't insert records into a partition that doesn't exist. 

So I'd do a DROP TABLE <partition_name> and then another CREATE TABLE ... PARTITION OF

backend/app/main.py
bailliekova: if we do the drop partition I don't think we get the number of records back.

bailliekova (commented):
resetting the map implemented this way seems like it also unshatters any shattered geoids; just confirming-- is that the desired behavior?

raphaellaude:
> resetting the map implemented this way seems like it also unshatters any shattered geoids; just confirming-- is that the desired behavior?

True. That seems ok for now and we can see if that's desired

backend/app/main.py
raphaellaude: Maybe it should be patch though if we're recreating the partition...

backend/app/main.py
nofurtherinformation: I believe that is correct

backend/app/main.py
nofurtherinformation: Makes sense -- thanks!

nofurtherinformation:
I'm picking back up on this

backend/app/main.py
nofurtherinformation: I think patch is appropriate here, but no strong opinions. Lmk your preference!

nofurtherinformation: This has been updated to implement this approach.

nofurtherinformation:
This has been updated to reflect feedback on the method for reseting the backend, and I've also cleaned up some of the frontend logic. There will need to be some updates with other in-progress changes (eg. reseting new store properties on map reset) -- but I think this is ready for review @raphaellaude  @bailliekova

app/src/app/components/sidebar/ResetMapButton.tsx
```diff
@@ -2,16 +2,11 @@ import {useMapStore} from '@/app/store/mapStore';
...
 export function ResetMapButton() {
-  const mapStore = useMapStore.getState();
-
-  const handleClickResetMap = () => {
-    mapStore.setFreshMap(true);
-    // clear map metrics
-    mapStore.setMapMetrics(null);
-  };
+  const handleClickResetMap = useMapStore(state => state.handleReset)
+  const noZonesAreAssigned = useMapStore(state => !state.zoneAssignments.size)
```
raphaellaude, line 6: OOOooohh love how you're just doing the bang in the callback

backend/app/main.py
```diff
@@ -174,6 +174,26 @@ async def shatter_parent(
     return result
 
 
+@app.patch(
+    "/api/update_assignments/{document_id}/reset", status_code=status.HTTP_200_OK
+)
+async def reset_map(document_id: str, session: Session = Depends(get_session)):
```
raphaellaude, line 180: H: Please add a quick test for this function in `test_main.py`!

nofurtherinformation: Added in a0c76299654f8e8f28ecbafe0359440b21620beb !

app/src/app/components/sidebar/ResetMapButton.tsx
nofurtherinformation: ty !

raphaellaude (approved):
lgtm

PR #127, opened 2024-10-16
Persist user maps 

nofurtherinformation:
Apropos #40 

## Description
This PR is for the feature enhancement descried in #40. The approach of this PR is to use local storage, with the intention that more in-depth future work might handle map sharing, etc.. 

## Reviewers

- Primary:
- Secondary:

## Checklist
- [x] User maps save to local storage on create
- [x] User maps can be reloaded
- [x] User maps can have custom names
- [x] User maps can be removed from local storage


raphaellaude (commented):
Nice work. I'll go through this in more detail later. @nofurtherinformation I'm curious if you looked at using [zustand's persistent store](https://github.com/pmndrs/zustand/blob/main/docs/integrations/persisting-store-data.md) rather than the custom hook for this? Creating a second separate store seems like a fine thing to do but curious if you weighed the trade-offs.

nofurtherinformation:
> Nice work. I'll go through this in more detail later. @nofurtherinformation I'm curious if you looked at using [zustand's persistent store](https://github.com/pmndrs/zustand/blob/main/docs/integrations/persisting-store-data.md) rather than the custom hook for this? Creating a second separate store seems like a fine thing to do but curious if you weighed the trade-offs.

Thanks! Thinking about this more, using the persist middleware might make more sense. It looks pretty robust, even including versioning and migrations. I'll swap this over today

app/src/app/store/mapStore.ts
```diff
@@ -95,231 +101,287 @@ export interface MapStore {
...
-      children: new Set(),
-    },
-    handleShatter: async (document_id, geoids) => {
-      set({ mapLock: true });
-      const shatterResult = await patchShatter.mutate({
-        document_id,
-        geoids,
-      });
+export const useMapStore = create(
+  persist(
```
raphaellaude, line 121: H: I really don't think we should persist the _entire_ store, no? I mean we could but I was imagining we just persist data that is unique to the user, e.g. the list of user maps.

app/src/app/components/sidebar/RecentMapsModal.tsx
```diff
@@ -0,0 +1,157 @@
...
+import { usePathname, useSearchParams, useRouter } from "next/navigation";
+import { DocumentObject } from "../../utils/api/apiHandlers";
+type NamedDocumentObject = DocumentObject & { name?: string };
+export const RecentMapsModal = () => {
+  const router = useRouter();
+  const pathname = usePathname();
+  const searchParams = useSearchParams();
+  const mapDocument = useMapStore((store) => store.mapDocument);
+  const userMaps = useMapStore((store) => store.userMaps);
+  const upcertUserMap = useMapStore((store) => store.upcertUserMap);
```
mariogiampieri, line 23: PP: should be "upsert" instead of "upcert" throughout to avoid confusion later on

app/src/app/store/mapStore.ts
mariogiampieri: That's super cool- it may be worth us adding a note to the docs to remind where / how that's being set

app/src/app/store/persistConfig.ts
```diff
@@ -0,0 +1,12 @@
+import { PersistOptions } from "zustand/middleware"
+import { MapStore } from "./mapStore"
+
+
+export const persistOptions: PersistOptions<MapStore, Partial<MapStore>> = {
+  name: 'districtr-persistrictr',
```
mariogiampieri, line 6: love this

app/src/app/components/sidebar/RecentMapsModal.tsx
```diff
@@ -0,0 +1,157 @@
+import { useMapStore } from "@/app/store/mapStore";
```
mariogiampieri, line 1: there's a silent warning passed in the console: `[zustand persist middleware] Unable to update item 'districtr-persistrictr', the given storage is currently unavailable.`. Per the discussion [here](https://github.com/pmndrs/zustand/discussions/1961) it seems it may be a hydration issue, but as far as I can tell it does not affect the saving or recalling of user maps

mariogiampieri (approved):
I would prefer the global renaming from "upcert" to "upsert" but non blocking. It would also be great to document that `persistConfig.ts` is where the items to save are being indicated

PR #129, opened 2024-10-17
Enhanced shattering

nofurtherinformation:
This PR works to enhance the shattering UX to support the use cases and detailed work on shattered geographies. 

## Description

The workflow this aims to accomplish is:
- A user selects a parent geography to shatter
- They draw on the blocks that are children of the shattered geography. No other geographies are interacted with
- The user exists the shatter view
- The user can re-enter the shatter view
- When outside of the shatter view, drawing over the shattered children will assign them as a parent geography

WIP screencap:

https://github.com/user-attachments/assets/9e2bd185-fb12-4931-9d0c-ac70f6c0d41b

## Reviewers

- Primary:
- Secondary:

## Checklist
- [x] Focus child geographies on shatter
- [x] Lock non-child geographies on shatter
- [x] Re-shatter
- [x] Save parent-child edges to browser on shatter 
- [x] Save parent-child edges to browser on load
- [x] Handle hover, paint, and other interactions when drawing over a shattered parent
- [x] Locking function, probably
- [x] ~Refactor to better abstraction for going from `ReturnType<queryRenderFeatures>` -> various map actions~ There are some refactors that could be good for a future PR, but I think we should mostly leave it for now.
- [x] Test user interactions
- [x] Merge main and resolve conflicts
- [x] Translucency of unpaintable areas is too subtle
- [x] We should add a big outline for the paintable area / VTD being shattered to make it obvious where you can paint
- [x] If we paint the entire set of blocks one color OR don't paint at all, we should unshatter the VTD.
- [x] Rename to "shatter mode" rather than "shatter view"


nofurtherinformation:
Stronger emphasis on paint area when shattered, and less opaque non-draw area:
Screenshot 2024-10-29 at 2 28 19 PM

nofurtherinformation:
VTDs now unshatter on paint when:
https://github.com/user-attachments/assets/f4c97523-8bd6-4f60-b873-0918c6dd06bf

nofurtherinformation:
Updates on this PR: It's pretty close! Also, this PR is not as big as the lines modified suggest, just hit a bunch of files with Prettier.

## Breaking (aka Shattering)
This PR adds a new focused block mode. On breaking a parent into blocks, either through the context menu or with the Break tool, the user is focused in on the selected parent and its children. Only those blocks can be interacted with until the user exits the block view.
Screenshot 2024-11-05 at 2 40 14 PM


There is a new option in the sidebar to highlight broken parent geographies.
Screenshot 2024-11-05 at 2 39 47 PM


## Healing (aka Unshattering)
After exiting the block view, the state checks to see if the broken parent should be "healed" -- removing the child block assignments and added back the parent. When exiting the block view, if all of the children are painted with the same zone or none are painted, the broken parent is healed. 

Additionally, when the user paints, if they paint over a broken parent, the state will do a similar check to see if all of the children are painted the same zone -- healing the parent if all the children and the same zone. This takes place after an assignment patch is complete and there are no other pending mutations to avoid race conditions.

## Locking
There were some similar uses to locking in this PR (like the block view described above) so it was a good time to bring in locking behavior. `lockedFeatures` is a `Set` in the application state that keeps track of IDs that the user wants to remain the same, and those IDs get
Screenshot 2024-11-05 at 2 40 52 PM
 filtered out on paint. 


On exiting block view, users can chose to lock the child blocks they just painted. We additionally introduce a new 'lock' tool that can lock specific geometries. For bulk locking (and unlocking) there is an option to lock all painted areas, meaning that everywhere with a zone assignment will be locked. 
Screenshot 2024-11-05 at 2 41 47 PM
Screenshot 2024-11-05 at 2 41 02 PM


## Painting, events, and state
As we've refactored more business logic and state to exist in the Zustand store, we've been able to move away from passing around state as a function's argument and instead simplify logic within the store. 

The ideal workflow moving forward is:
- Keep track of some sort of feature state in the store (eg. focused features like in block view, highlighted features to find broken parents, locked features, etc.)
- Subscribe via Zustand to that part of the state
- Apply feature state / rendering to maplibre based on the current and previous state (eg. semi-transparent to indicate locked) 

Keeping more in the store simplifies debugging and also orchestrating multi-part server interactions. Tanstack query takes some work to play nicely with Zustand, relying on `queryClient.isMutating()` to determine if there are other server interactions happening. In the future, really going all in on zustand (or including a more fully featured state management library like Redux) might make sense.

Additionally, future PRs might refactor some or all of the following:
- Split `mapStore` into slices
- Refactor style layer management
- Cleaner handling of events and optimizing `queryRenderedFeatures`

nofurtherinformation:
You can now lock by color zone, using a similar interface to the brush zone selector when the lock tool is selected. You can still lock/unlock individual geographies.

(This is actually the final feature enhancement of this PR)

Screenshot 2024-11-07 at 4 23 34 PM

backend/app/main.py
```diff
@@ -174,6 +176,34 @@ async def shatter_parent(
...
+    results = session.execute(
+        statement=stmt,
+        params={
+            "input_document_id": document_id,
+            "parent_geoids": data.geoids,
+            "input_zone": data.zone,
+        },
+    )
+    session.commit()
+    result = UnShatterResult(parents=GEOIDS(geoids=data.geoids))
```
raphaellaude, line 203: PP to use `RETURNING` from the SQL query rather than just pass back the input data.

backend/app/models.py
```diff
@@ -163,11 +163,19 @@ class GEOIDS(BaseModel):
...
+class ZoneAndGEOIDS(GEOIDS):
+    zone: int | None
+
+
 class ShatterResult(BaseModel):
     parents: GEOIDS
     children: list[Assignments]
 
 
+class UnShatterResult(BaseModel):
```
raphaellaude, line 175: PP: Why double nest the result when we could just return `GEOIDS`? I guess it's not a bad idea to be explicit.

backend/app/models.py
```diff
@@ -163,11 +163,19 @@ class GEOIDS(BaseModel):
     geoids: list[str]
 
 
+class ZoneAndGEOIDS(GEOIDS):
```
raphaellaude, line 166: O: Naming? I guess this is explicit...! I would have maybe called this AssignedGEOIDS or something 🤷🏻

backend/app/alembic/versions/65a4fc0a727d_add_unshatter_udf.py
```diff
@@ -0,0 +1,38 @@
+"""Add unshatter UDF
+
+Revision ID: 65a4fc0a727d
+Revises: 5ab466c5650a
```
raphaellaude, line 4: H: Need to adjust the revised migration. See recent example from the bounds work:
```python
"""add extent to gerrydb

Revision ID: dc391733e10a
Revises: 5ab466c5650a
Create Date: 2024-10-27 19:38:13.798056

"""

from typing import Sequence, Union

from alembic import op
import sqlalchemy as sa

# revision identifiers, used by Alembic.
revision: str = "dc391733e10a"
down_revision: Union[str, None] = "5ab466c5650a"
branch_labels: Union[str, Sequence[str], None] = None
depends_on = ("167892041d95",)  # geometry col renaming
down_revision = "167892041d95"
```

You'll want to add the `depends_on` and change the `down_revision` to `dc391733e10a`

app/src/app/utils/helpers.ts
```diff
@@ -71,26 +77,65 @@ export const boxAroundPoint = (
...
+  }
+  // if filtering locked feature, remove those
+  if (filterLocked && lockedFeatures.size) {
+    features = features.filter(f => !lockedFeatures.has(f.id?.toString() || ''));
+  }
 
-  return map?.queryRenderedFeatures(bbox, {layers});
+  // if there is a child layer and parents have been shattered
+  // check if any of the selected IDs are parents
+  if (mapDocument?.child_layer && shatterIds.parents.size) {
```
raphaellaude, line 108: H: Not sure where to put this comment but I managed to get to a state on the test app with a call stack exceeded error and overlapping geoms.

Screenshot 2024-11-07 at 10 39 17 PM

raphaellaude (changes requested):
lgtm except for a revision update and an FE error we might want to investigate.

backend/app/models.py
nofurtherinformation: be3a89f6ed0d48fb1b042faa9ca4601eab2d074e

backend/app/models.py
nofurtherinformation: Resolved in defeb6ab38d32bcaa8a1863526e3e5a32334ee7e

app/src/app/utils/helpers.ts
nofurtherinformation: Investigating now -- thank you for flagging this!

backend/app/alembic/versions/65a4fc0a727d_add_unshatter_udf.py
nofurtherinformation: I updated the comment at the top of the alembic revision, but I think the rest is correct. Here is the current revision config:

```python
revision: str = "65a4fc0a727d"
down_revision: Union[str, None] = "dc391733e10a"
branch_labels: Union[str, Sequence[str], None] = None
depends_on: Union[str, Sequence[str], None] = None
```

And with main merged in here is the history:
```bash
# alembic history
dc391733e10a -> 65a4fc0a727d (head), Add unshatter UDF
167892041d95 (167892041d95) -> dc391733e10a, add extent to gerrydb
5ab466c5650a -> 167892041d95, rename geometry cols to geometry
...
```

Are there additional revisions this needs to depend on?

app/src/app/utils/helpers.ts
nofurtherinformation: I wasn't able to replicate this exactly but was getting some strange rendering behavior on blur/refocus of the window/tab. I added some safety around that (a new app loading state 'blurred') and cleaned up some rendering subscriptions

backend/app/main.py
```diff
@@ -174,6 +175,33 @@ async def shatter_parent(
     return result
 
 
+@app.patch(
```
raphaellaude, line 178: H: Sorry forgot to mention this before but could you add a test for this function, too?

app/src/app/constants/layers.ts
```diff
@@ -67,68 +134,73 @@ export function getBlocksLayerSpecification(
...
+        '#e5ff00', // yellow color when highlighted
+        '#000000', // Default color
+      ],
+      'line-width': [
+        'case',
+        ['boolean', ['feature-state', 'focused'], false],
+        10, // Width of 10 when focused
+        ['boolean', ['feature-state', 'highlighted'], false],
+        10, // Width of 5 when highlighted
+        0, // Default width
```
raphaellaude, line 153: PP: What do you think of using a round cap? Looks a little messy with the super thick line. Also have a personal pref to reduce the thickness a bit cause it's _really_ thick but won't die on that hill!

image

raphaellaude (changes requested):
Looking really good – two _v_ small comments, only one blocking.

backend/app/main.py
nofurtherinformation: Ah, yes 100%. Had this on my list todo but totally forgot!

app/src/app/constants/layers.ts
nofurtherinformation: Round cap is a good call!

raphaellaude (approved):
lgtm!

PR #132, opened 2024-10-18
Pull tiles from s3

### LINKED ISSUE #130: Move data to S3 buckets
We were having an issue on Cornell's networks with r2 buckets being blocked. Maybe has something to do with the firewall blocking suspicious looking endpoints e.g. `https://pub-fa71193941a74e14a38eee99f30f53d9.r2.dev`

S3 should be slightly faster anyway, though more expensive

raphaellaude:
## Description

- Fixes #130

## Reviewers

- Primary: @mariogiampieri 
- Secondary: @nofurtherinformation @mapmeld 

## Checklist


mariogiampieri (approved):
This is good- we will also need to change the repo's `NEXT_PUBLIC_S3_BUCKET_URL` secret so that the pr app action runs properly as well

Issue #133, opened 2024-10-21
Pointer cursor when painting


raphaellaude:
The cursor should change from the pan open hand to the `pointer` hand when painting. See existing Districtr for reference.

PR #142, opened 2024-10-22
Shatter tool

### LINKED ISSUE #134: Add shatter tool
Shattering is currently a bit of an easter-egg. Users need to know it's there. We should add a shatter tool / button to make this easier for the user to discover.

raphaellaude:
## Description

- Fixes #134 and #133

## Reviewers

- Primary: @mariogiampieri 
- Secondary: @mapmeld @nofurtherinformation 

## Checklist

### DIFF
--- a/app/src/app/store/mapRenderSubs.ts
+++ b/app/src/app/store/mapRenderSubs.ts
@@ -20,7 +20,7 @@ export const getRenderSubscriptions = (useMapStore: typeof _useMapStore) => {
     (state) => [state.mapDocument, state.getMapRef],
     ([mapDocument, getMapRef]) => {
       const mapStore = useMapStore.getState();
-      const mapRef = getMapRef()
+      const mapRef = getMapRef();
       if (mapRef && mapDocument) {
         addBlockLayers(mapRef, mapDocument);
         mapStore.addVisibleLayerIds([BLOCK_LAYER_ID, BLOCK_HOVER_LAYER_ID]);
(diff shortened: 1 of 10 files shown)


app/src/app/store/mapRenderSubs.ts
```diff
@@ -91,10 +95,38 @@ export const getRenderSubscriptions = (useMapStore: typeof _useMapStore) => {
       { equalityFn: shallowCompareArray },
     );
 
+  const _updateMapCursor = useMapStore.subscribe<MapStore["activeTool"]>(
```
nofurtherinformation, line 98: PP: This could also be in the map component and use tailwind `cursor-x` [properties](https://tailwindcss.com/docs/cursor)

app/src/app/utils/events/mapEvents.ts
```diff
@@ -57,6 +58,12 @@ export const handleMapClick = (
         SelectZoneAssignmentFeatures(mapStore);
       });
     }
+  } else if (activeTool === "shatter") {
+    const documentId = mapStore.mapDocument?.document_id;
+    const featureId = e.features?.[0].id?.toString();
```
nofurtherinformation, line 63: PP: Highlight on mouse move? Eg. Select features, `mapStore.setHoveredFeatures(features[0])`

nofurtherinformation (commented):
Two small suggestions here but otherwise looks good! 

I think having the cursor properties be in the component makes more sense than in the subscriptions, since modifying CSS is a bit less tricky than the maplibre rendering operations. 

Also, if we can add a hover state while using the shatter tool on mouse move that would be ideal!

app/src/app/utils/events/mapEvents.ts
raphaellaude: Good call!

app/src/app/store/mapRenderSubs.ts
raphaellaude: I like this suggestions but I'm sort of in a hurry so created new issue #145 so we can do this later

app/src/app/utils/events/mapEvents.ts
raphaellaude: Added

nofurtherinformation (commented):
Looks good to me. @mariogiampieri ?

mariogiampieri (dismissed):
looks great

PR #144, opened 2024-10-23
autostop apps

raphaellaude:
## Description

- Fly is so fast at spinning up machines again that we should autostop all apps except dbs, esp. for PR apps. We can always set min machines running to 1 for prod down the line, having different configs, but for now I think this is appropriate.

## Reviewers

- Primary: @mariogiampieri 
- Secondary:

## Checklist


mariogiampieri (approved):
lgtm!

PR #147, opened 2024-10-27
Feat/zoom to view

mariogiampieri:
use `fitBounds()` to center map properly on AOI

## Description

- start map at continental us scale
- make migration to backfill extents based on gerrydb tables
- `cli.py` additions to make bounds when importing gerrydb views
- setting map extent based on document creation, loading

## Reviewers

- Primary: @raphaellaude 
- Secondary:

## Checklist


mariogiampieri:
See https://districtr-v2-147-app.fly.dev/map for demo. I haven't fully tested yet the setting of the extent when loading in via `cli.py` but this should suffice to test for existing views

backend/app/alembic/versions/dc391733e10a_add_extent_to_gerrydb.py
```diff
@@ -0,0 +1,117 @@
...
+
+def upgrade() -> None:
+    # ### commands auto generated by Alembic - please adjust! ###
+    # Make column nullable to avoid errors during initial insert
+    op.add_column(
+        "gerrydbtable", sa.Column("extent", sa.ARRAY(sa.Float()), nullable=True)
+    )
+
+    # Use a session to interact with the database
+    bind = op.get_bind()
```
raphaellaude, line 30: I: I think it's a good idea to create these bboxes but was wondering if we should [align all geometry columns to "geometry"](https://github.com/districtr/districtr-v2/pull/148) to simplify this logic.

backend/app/utils.py
```diff
@@ -104,3 +107,45 @@ def create_parent_child_edges(
             "districtr_map_uuid": districtr_map_uuid,
         },
     )
+
+
+def transform_bounding_box(bbox: list, source_epsg: str, target_epsg: str) -> list:
```
raphaellaude, line 112: H: I would much rather doing this in postgis. This logic looks quite complex and depends on an additional package, `osgeo`, which [seems largely to be unmaintained](https://pypi.org/project/osgeo/) seeing as it's on v0.0.1 since Jun '21. If we do do this in python I'd prefer to depend on packages like `fiona` and `shapely` since they're actively maintained and are likely to be added to our environment when we bring in MGGG packages to do things like contiguity.

backend/cli.py
```diff
@@ -111,19 +131,22 @@ def import_gerrydb_view(layer: str, gpkg: str, replace: bool, rm: bool):
...
 
-    upsert_query = text("""
+    upsert_query = text(
+        """
         INSERT INTO gerrydbtable (uuid, name, updated_at)
-        VALUES (gen_random_uuid(), :name, now())
+        VALUES (gen_random_uuid(), :name, :extent, now())
         ON CONFLICT (name)
         DO UPDATE SET
             updated_at = now()
```
raphaellaude, line 140: H: You're not actually setting the extent on conflict. Not sure if that's intentional or not but if the extent was null in an existing table it would remain null. Also you have a mismatch in the number of inserted values and columns specified so this would fail anyway.

backend/cli.py
```diff
@@ -96,6 +97,25 @@ def import_gerrydb_view(layer: str, gpkg: str, replace: bool, rm: bool):
         ],
     )
 
+    # calculate layer extent to store in gerrydbtable
+    result = subprocess.run(
```
raphaellaude, line 101: H: Related to previous comments, I think we should add this as part of the `districtrmap` and not the `gerrydbtable`.

raphaellaude (changes requested):
Frontend looks good and have a few backend comments: I think we can simplify significantly by moving the extent to `districtrmap`, doing more with postgis, and renaming all geometry columns as [suggested here](https://github.com/districtr/districtr-v2/pull/148).

app/src/app/utils/api/mutations.ts
```diff
@@ -8,6 +8,7 @@ import {
 } from "@/app/utils/api/apiHandlers";
 import { useMapStore } from "@/app/store/mapStore";
 import { mapMetrics } from "./queries";
+import { use } from "react";
```
raphaellaude, line 11: Q: where does this get used?

backend/app/alembic/versions/dc391733e10a_add_extent_to_gerrydb.py
```diff
@@ -0,0 +1,85 @@
...
+        cases.append(case)
+    # Combine all cases into a single SQL statement
+    case_statement = " ".join(cases)
+    # Execute a single UPDATE statement
+    update_query = sa.text(
+        f"""
+        UPDATE districtrmap
+        SET extent = CASE
+            {case_statement}
+        -- ELSE ARRAY[0, 0, 0, 0] if this fails, there is no
```
raphaellaude, line 66: PP: `ELSE NULL` – and we default to Kansas view? Non-blocking esp. since this is in a migration.

backend/app/main.py
```diff
@@ -210,8 +212,12 @@ async def get_document(document_id: str, session: Session = Depends(get_session)
             DistrictrMap.child_layer.label("child_layer"),  # pyright: ignore
             DistrictrMap.tiles_s3_path.label("tiles_s3_path"),  # pyright: ignore
             DistrictrMap.num_districts.label("num_districts"),  # pyright: ignore
+            DistrictrMap.extent.label("extent"),  # pyright: ignore
         )  # pyright: ignore
-        .where(Document.document_id == document_id)
+        .where(
+            Document.document_id == document_id
+            and DistrictrMap.parent_layer == GerryDBTable.name
```
raphaellaude, line 219: H: We can not remove this expression after the `and`, right?

backend/app/main.py
```diff
@@ -19,6 +19,7 @@
     Document,
     DocumentCreate,
     DocumentPublic,
+    GerryDBTable,
```
raphaellaude, line 22: H: And remove this import?

backend/app/main.py
```diff
@@ -220,6 +226,7 @@ async def get_document(document_id: str, session: Session = Depends(get_session)
         .limit(1)
     )
     result = session.exec(stmt)
+    print(result)
```
raphaellaude, line 229: PP: clean up prints

backend/app/utils.py
```diff
@@ -1,6 +1,7 @@
 from sqlalchemy import text
 from sqlalchemy import bindparam, Integer, String, Text
 from sqlmodel import Session
+from osgeo import ogr, osr
```
raphaellaude, line 4: H: Remove unused import

raphaellaude (changes requested):
This looks really great. Pretty much ready to merge. If my questions/comments are dumb please excuse – working with v tired brain while reviewing...

raphaellaude:
Just want to jump in to say that playing with the app this is really smooth and nice. Super psyched to paint some maps outside of Colorado

app/src/app/utils/api/mutations.ts
mariogiampieri: old / unused!

backend/app/main.py
mariogiampieri: yep good call

backend/app/utils.py
mariogiampieri: good catch

backend/app/alembic/versions/dc391733e10a_add_extent_to_gerrydb.py
mariogiampieri: agreed and i don't expect it'll happen on prod. the column is set to not allow for nulls and I'd rather keep it that way so will leave as-is for now, but am open to better handling issues down the line

mariogiampieri:
> Just want to jump in to say that playing with the app this is really smooth and nice. Super psyched to paint some maps outside of Colorado

ty! and your latest review is super helpful, thanks for catching those things. agreed that it'll be fun to move around the map finally!

backend/app/alembic/versions/dc391733e10a_add_extent_to_gerrydb.py
raphaellaude: Oh I thought I saw `nullable=True` in the model, though now that I'm thinking about it not in the revision.

backend/app/alembic/versions/dc391733e10a_add_extent_to_gerrydb.py
```diff
@@ -0,0 +1,87 @@
...
+        -- if this fails, there is no
+        -- matching table from parent_layer
+        -- and that's a problem
+        END;
+    """
+    )
+
+    bind.execute(update_query)
+
+    # Make the `extent` column non-nullable
```
mariogiampieri, line 76: the last step of the upgrade is to flip it back to not nullable

raphaellaude: Ah got it

app/src/app/constants/layers.ts
```diff
@@ -176,39 +167,45 @@ const addBlockLayers = (
...
-      LABELS_BREAK_LAYER_ID,
+      LABELS_BREAK_LAYER_ID
     );
   }
-  useMapStore.getState().setMapRenderingState("loaded")
+  useMapStore.getState().setMapRenderingState("loaded");
+  useMapStore
+    .getState()
+    .getMapRef()
+    ?.fitBounds(mapDocument.extent as [number, number, number, number], {
```
nofurtherinformation, line 199: PP: Map bounds could stored in state and have a `useEffect` in the map component to `fitBounds` to decouple from layers add/remove

mariogiampieri:
> This is looking and feeling really smooth. I added one small comment but otherwise this is great

very good call- now streamliined / ready for re-review

raphaellaude (approved):

raphaellaude:
Looks great! Let's 🚢

PR #157, opened 2024-10-28
Implement undo/redo with zundo

mapmeld:
Testing if we can replace #118 with the zundo plugin , and whether GitHub Actions will produce a test build for this
- Adds undo/redo buttons to UI
- Wraps mapState in temporal(...)

Issues:
`onSave` still needs to update MapBox and the server

### DIFF
--- a/.github/workflows/fly-deploy-pr.yml
+++ b/.github/workflows/fly-deploy-pr.yml
@@ -1,4 +1,4 @@
-name: Fly Deploy Disctictr V2 Pull Request app
+name: Fly Deploy Districtr V2 Pull Request app
 on:
     pull_request:
         types: [opened, reopened, synchronize, closed]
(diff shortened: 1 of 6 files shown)


mapmeld:
This is working better now (possibly to undo/redo paint actions and those get reflected on the map). I think that I need to change the equality function to make sure it doesn't save intermediate / invisible changes

mapmeld:
ok, interestingly by the undo/redo works pretty well by default after a block has a paint or eraser (null) value attached to it. Then actions after that can be undone. Will revisit

nofurtherinformation:
There is an optimization that checks the previous state before changing the feature state, but this is producing some unexpected behavior. For now, I'd recommend removing this and just have all the zone assignments assigned each time.

https://github.com/districtr/districtr-v2/blob/6cc5e6e47461750dd7cfd4b2b291c93977dd94f9/app/src/app/utils/helpers.ts#L345

Also as a heads up there will be some substantial merge conflicts coming in from main!

mapmeld:
I agree this should be paused especially since I don't know enough yet about shattering to say anything informed on implementing that

PR #161, opened 2024-11-02
Fix backend tests

raphaellaude:
## Description

- We broke the tests by merging a broken migration and requiring parent layers to exist when creating a districrmap. That's probably a reasonable assumption but requires a larger refactor of the test suite.
- In the meantime let's just make the extent nullable and create a separate command to add the extent. Added function and command `add_extent_to_districtrmap`. This function will be called by default when creating a districtr map. Coupling the extent creation logic in the CLI ensures we still calculate the extent on creation but don't have to couple logic deep in UDFs (specifically `create_districtr_map`).
- Added options to the `create_districtr_map` and `add_extent_to_districtrmap` CLI commands to manually override bounds.

> [!CAUTION]
> After merging, we will need to rollback the broken `dc391733e10a` and reapply the now fixed migration.

## Reviewers

- Primary: @mariogiampieri 
- Secondary: @nofurtherinformation @bailliekova @mapmeld 

## Checklist


mariogiampieri (changes requested):
PP: Lgtm although we may as well remove that function altogether if it's not (not likely will be) used anywhere

mariogiampieri (approved):

raphaellaude:
Looks like the tests are messed up. I'll use this PR to fix them

raphaellaude:
@mariogiampieri this is now ready for review!

mariogiampieri (approved):
This looks solid, nicely done!

PR #163, opened 2024-11-05
Move PMTiles To API

nofurtherinformation:
This PR is mostly to test speed vs PMTiles directly from the browser.

## Description
- Make a little NextJS Api route to handle tiles

## Reviewers

- Primary:
- Secondary:


raphaellaude:
I think it's worth testing this with a separate `go-pmtiles` server rather than running it in node, if we go this route

PR #165, opened 2024-11-10
Refactor tests to be idempotent

### LINKED ISSUE #91: Backend tests should rollback changes
Right now, we set up a postgres db for all backend tests and don't rollback any fixtures / data committed to the db. That makes it very difficult to reason about adding tests because adding a test can change the global state in which tests are run.

We should instead probably have each test be run atomically, rolling back all other data committed when running a test.

raphaellaude:
## Description

- Currently the backend tests are not idempotent, so you have to keep the state of the test db in your head as you write tests. This is of course difficult and annoying, not to mention less clear / explicit since adding fixtures for the state needed by a test can actually break it, due to unique constraints.
- This PR runs all tests in an outer transaction, which is always rolled back (fixes #91).
- Finally, renames the persistent session to `persistent_session` so this is still an option (for e.g. larger gerrydb tables needed for testing).

## Reviewers

- Primary: @bailliekova @nofurtherinformation 
- Secondary: @mapmeld @mariogiampieri 

## Checklist


backend/tests/test_main.py
```diff
@@ -406,29 +393,6 @@ def test_get_document_population_totals(
     assert data == [{"zone": 1, "total_pop": 67}, {"zone": 2, "total_pop": 130}]
 
 
-def test_get_document_population_totals_null_assignments(
```
raphaellaude, line 409: This was a duplicate test–just cleaning up.

nofurtherinformation:
This is looking nice and definite QOL improvement for testing. 

I think just needs main merged (new unshatter test) and then any docs you see needed but otherwise lgtm!

nofurtherinformation (dismissed):
`test_utils` lgtm! (+ assuming rest will squashed in the merge)

nofurtherinformation:
lgtm!

nofurtherinformation (approved):

PR #166, opened 2024-11-10
Implement county paint using FIPS prefix

mapmeld:
Blocks, block groups, and VTDs get their ID / GEOID from their Census FIPS code, where the first two digits represent their state and the next three digits represent their county.

While in county paint mode, we can select all blocks/VTDs in the same county by this prefix.
Because this uses `queryRenderedFeatures`, it will not catch blocks outside the current view.

This makes assumptions, and if this is wrong I should change it, or enable/disable county paint on specific maps:
- the county layer is consistent between maps and always has COUNTYFP
- the paintable features always have an 'id' property with **update: first 5 digits in the string being the FIPS code** ~the format 'vtd:42123....'~
- the paintable features are always blocks, block groups, or US Census VTDs (which we've confirmed do not cross counties)

Update: removed turf from dependencies


app/src/app/utils/helpers.ts
```diff
@@ -119,66 +116,19 @@ export const getFeaturesIntersectingCounties = (
...
-    // @ts-ignore: Property 'coordinates' does not exist on type 'Geometry'.
-    countyPoly = multiPolygon(countyFeatures[0].geometry.coordinates);
-  }
-
   return features.filter(p => {
-    const point = pointOnFeature(p);
-    return booleanWithin(point, countyPoly);
+    if (!(p?.id)) return false;
+    const myFips = p.id.toString().match(/\d{5}/);
+    return myFips && fips.has(myFips[0]);
```
raphaellaude, line 128: Q: This logic looks good and works in the app but feels a little laggy. `Set.has` should be a constant time operation which is why it's sort of weird that it would be the cause of the lag. We could try using [`Set.prototype.intersection()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Set/intersection) like:
```javascript
return fips.intersection(
    new Set(features.filter(p => !(p?.id) &&  p.id.toString().match(/\d{5}/)))
)
```
... but I'm not sure if that's the issue or...

Q: What does querying `undefined` do? We only need to query the paintable layers so perhaps passing those IDs in will improve perf.

app/src/app/utils/helpers.ts
```diff
@@ -119,66 +116,19 @@ export const getFeaturesIntersectingCounties = (
   });
 
   if (!countyFeatures) return;
+  const fips = new Set<string>(countyFeatures.map(c => c.properties.STATEFP + c.properties.COUNTYFP));
```
raphaellaude, line 119: O: Eventually we should create a dedicated column for this so we're not doing this computation so much.

raphaellaude (commented):
Looks good – just have a couple questions around perf

app/src/app/utils/helpers.ts
mapmeld: `undefined` is the screen bounding box, undefined meaning the full viewport
I thought that { layers } in that call is limiting it to one layer, as we also have `layers: string[] = [BLOCK_HOVER_LAYER_ID]` in the function args

app/src/app/utils/helpers.ts
```diff
@@ -119,64 +116,13 @@ export const getFeaturesIntersectingCounties = (
...
-      let y = coords[i + 1];
-      sw.lng = Math.min(sw.lng, x);
-      sw.lat = Math.min(sw.lat, y);
-      ne.lng = Math.max(ne.lng, x);
-      ne.lat = Math.max(ne.lat, y);
-    }
-  });
-
-  return [sw, ne];
+  return features.filter(p => (p?.id) &&
```
nofurtherinformation, line 124: Q: This approach won't capture geometries that are not rendered. We've discussed that the issue of non-rendered geometries is not too bad for painting blocks, but in this case a county could extend outside of the window bounds.

Should we consider an approach with a backend query to get all the expected IDs?

Eg.pseudo
```javascript
const fips = brushedFeature.properties.fips
const countyChildren = await fetch(`api/county_children/{document_id}/{fips}`)
setZoneAssignments(
  countyChildren,
  zone
)
```

nofurtherinformation (changes requested):
This looks good but I did want to ask about concerns for non-rendered geometries. I know we want to avoid too much back and forth between server and client vis-a-vis assignments.

app/src/app/utils/helpers.ts
raphaellaude: > `undefined` is the screen bounding box

Ah, got it ty

app/src/app/utils/helpers.ts
raphaellaude: PP: it's reasonable that a user might only paint what they see. Since the current functionality is actively buggy I'd recommend we go with this and can create an issue to experiment w/ a backend approach. Also–and I'm not wedded to this opinion–there's something kind of nice about FE being responsible for painting and API/DB only being responsible for documents and their assignments

nofurtherinformation: Sounds good!

app/src/app/utils/helpers.ts
```diff
@@ -119,64 +116,13 @@ export const getFeaturesIntersectingCounties = (
   });
 
   if (!countyFeatures) return;
```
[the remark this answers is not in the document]
mapmeld: ✅ fixed in latest commit

raphaellaude (approved):
lgtm and is working nicely in test app. @nofurtherinformation ?

nofurtherinformation (dismissed):
agreed lgmt!

nofurtherinformation:
(P.s. Just needs main merged in but conflict looks minimal)

mapmeld:
Rebased and updated dependencies so I need a renewed ✅ then this is ready to merge

raphaellaude (approved):
Lgtm

nofurtherinformation (commented):
lgmt!

PR #167, opened 2024-11-11
Chart enhancements + stub of `places` and `districtingProblems`

mariogiampieri:
## Description

- Chart behavior based on `MapDocument` properties, including a target number of districts (manually set for now) and a total population. This chart behavior includes:
    - an ideal population per zone indicated as a vertical labelled line
    - the chart x-max set based on ideal pop * 2 or dataMax, whichever is larger (i.e. if someone is going overboard the chart will follow
    - separately, setting the # of colors in the palette to equal the target # of districts. 

Re: places and problems, to circle back on:
- Create draft `DistrictrPlace` and `DistrictrProblem` models, following v1 places / problems and syntax that a user selects to start a session. This was in service of getting the proper target # districts for each map. I suspect that we'll want to discuss and revise how these objects fit in with our data model. For now at least, they provide a target for each test map, based on districtr v1 statewide places and problems
- Per the above, utility scripts (still WIP) to load data from districtr v1 into these new tables. Again, we should discuss where these things should live long term (holding for a later branch)

## Reviewers

- Primary:
- Secondary:

## Checklist

image


backend/app/main.py
```diff
@@ -212,6 +274,8 @@ async def get_document(document_id: str, session: Session = Depends(get_session)
             DistrictrMap.tiles_s3_path.label("tiles_s3_path"),  # pyright: ignore
             DistrictrMap.num_districts.label("num_districts"),  # pyright: ignore
             DistrictrMap.extent.label("extent"),  # pyright: ignore
+            # get total population based on parent layer
+            population_subquery.as_scalar().label("total_population"),
```
raphaellaude, line 278: H: It feels really heavy to calculate the total population here in a subquery. Especially since this is a frequently hit endpoint. I'd suggest adding `total_population` to `DistrictrMap` and getting the column in the join w/ `DistrictrMap.total_population.label("total_population")`

backend/load_dv1_places_problems.py
```diff
@@ -0,0 +1,273 @@
+from uuid import uuid4
```
raphaellaude, line 1: I: I've been trying to keep the extract/transform logic separate from the app in [pipelines](https://github.com/districtr/districtr-v2/tree/main/pipelines). So you can prepare a dataset, dump it to bucket storage, and then separately load it to the app. I'd recommend following that pattern so that
- The app isn't responsible for brittle pipeline logic
- The pipeline logic is more easily adjustable / testable if it's not coupled to postgres
- We have a single artifact in S3 that's easy to debug outside of the postgres database
- it's easier to understand what's going on since we have 1 artifact -> loaded into 1 table rather than hundreds/thousands of queries. This will also greatly improve load perf and not burden the prod DB when we run this

backend/load_dv1_places_problems.py
```diff
@@ -0,0 +1,273 @@
...
+        print(
+            f"Updated map entry {map_id} with place_id {place_id} with {problem_num_parts} target districts"
+        )
+        session.commit()
+
+
+if __name__ == "__main__":
+    places = load_dv1_places()
+    places = assign_problems_to_places(places)
+    upsert_places_and_problems(places)
```
raphaellaude, line 273: H: Use `cli.py` for these types of jobs, which provides `--help`, is our current pattern, and makes it easier to split up the work into smaller pieces.

raphaellaude (commented):
I know you didn't want a review yet but I had a few big comments I thought would be better to discuss sooner than later

raphaellaude:
> I know you didn't want a review yet but I had a few big comments I thought would be better to discuss sooner than later

PS. Nice work. The chart looks super awesome, too.

mariogiampieri:
> > I know you didn't want a review yet but I had a few big comments I thought would be better to discuss sooner than later
> 
> PS. Nice work. The chart looks super awesome, too.

TY and agree with all your points! I will respond to each in kind, but for now, the PR app does have a draft bar chart that's more sensible: https://districtr-v2-167-app.fly.dev/

backend/app/main.py
mariogiampieri: confirmed not needed now and removed in favor of changes in #87

mariogiampieri:
Ok, the chart is working as intended now, with a pin put in the `places` and `problems`. The `num_districts` can be manually set for now for the chart to behave properly.

nofurtherinformation (commented):
I added a couple small comments but FE stuff looks good to me for user testing. I didn't look too deeply into the backend work here for now

app/src/app/components/sidebar/charts/HorizontalBarChart.tsx
```diff
@@ -24,6 +36,45 @@ const CustomTooltip = ({active, payload: items}: TooltipInput) => {
...
+  >([]);
+
+  useEffect(() => {
+    if (summaryStats && numDistricts) {
+      const totalPop = summaryStats.totpop ? getEntryTotal(summaryStats.totpop.data) : 0;
+      setIdealPopulation(totalPop / numDistricts);
+    }
+  }, [summaryStats, numDistricts]);
+
+  const calculateChartObject = () => {
```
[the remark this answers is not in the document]
mariogiampieri: good catch

app/src/app/components/sidebar/charts/HorizontalBarChart.tsx
```diff
@@ -24,6 +36,45 @@ const CustomTooltip = ({active, payload: items}: TooltipInput) => {
...
+  const [idealPopulation, setIdealPopulation] = useState<number | null>(null);
+  const maxNumberOrderedBars = 40; // max number of zones to consider while keeping blank spaces for missing zones
+  const [totalExpectedBars, setTotalExpectedBars] = useState<
+    Array<{zone: number; total_pop: number}>
+  >([]);
+
+  useEffect(() => {
+    if (summaryStats && numDistricts) {
+      const totalPop = summaryStats.totpop ? getEntryTotal(summaryStats.totpop.data) : 0;
+      setIdealPopulation(totalPop / numDistricts);
```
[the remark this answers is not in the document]
mariogiampieri: nice!  good call

app/src/app/components/sidebar/charts/HorizontalBarChart.tsx
```diff
@@ -44,30 +88,49 @@ export const HorizontalBar = () => {
   return (
     <Flex gap="3" direction="column">
       <Heading as="h3" size="3">
-        Population by Zone
+        Population by District
```
raphaellaude, line 91: PP: looks like Moon has been using sentence case which I think is better style -> `Population by district`

app/src/app/utils/api/apiHandlers.ts
```diff
@@ -172,16 +172,16 @@ export interface P1ZoneSummaryStats {
...
 
 export const P1ZoneSummaryStatsKeys = [
   'other_pop',
   'asian_pop',
   'amin_pop',
   'nhpi_pop',
   'black_pop',
-  'white_pop'
-] as const
+  'white_pop',
```
raphaellaude, line 183: I: Will want to add `two_or_more_races_pop` here

app/src/app/utils/api/apiHandlers.ts
```diff
@@ -192,7 +192,7 @@ export const CleanedP1ZoneSummaryStatsKeys = [
   'nhpi_pop_pct',
   'black_pop_pct',
   'white_pop_pct',
```
raphaellaude, line 194: I: and add `two_or_more_races_pop` here

backend/app/main.py
```diff
@@ -5,7 +5,7 @@
 from starlette.middleware.cors import CORSMiddleware
 from sqlalchemy.dialects.postgresql import insert
 import logging
-from sqlalchemy import bindparam
+from sqlalchemy import bindparam, func
```
raphaellaude, line 8: Q: is func used here? I don't see it in the diff

backend/app/sql/summary_stats_pop_totals.sql
```diff
@@ -0,0 +1,40 @@
+DROP FUNCTION IF EXISTS get_summary_stats_pop_totals;
```
raphaellaude, line 1: H: This function isn't being used now, no? We can delete in favor of using the either `get_summary_p1_totals` or `get_summary_p4_totals` (and eventually others...)?

backend/app/sql/summary_stats_pop_totals.sql
```diff
@@ -0,0 +1,40 @@
...
+CREATE OR REPLACE FUNCTION get_summary_stats_pop_totals(gerrydb_table TEXT)
+RETURNS TABLE (
+    other_pop BIGINT,
+    asian_pop BIGINT,
+    amin_pop BIGINT,
+    nhpi_pop BIGINT,
+    black_pop BIGINT,
+    white_pop BIGINT
```
raphaellaude, line 9: H: Even though we should remove this function, we'll need to add `two_or_more_races_pop` to `get_summary_p1_totals`.

pipelines/simple_elt/main.py
```diff
@@ -311,5 +314,96 @@ def merge_gerrydb_tilesets(
...
+        except Exception as e:
+            LOGGER.error(f"Failed to download problems for {url}: {e}")
+            continue
+
+
 if __name__ == "__main__":
     cli()
+
+
+def upsert_places_and_problems():
```
raphaellaude, line 374: PP: Move above `if __name__ == "__main__"` but add `raise NotImplementedError` in the function body. Or just `warnings.warn("Not implemented")` or something

app/src/app/utils/api/apiHandlers.ts
raphaellaude: Will handle in separate PR

app/src/app/utils/api/apiHandlers.ts
raphaellaude: Will handle in separate PR

raphaellaude (changes requested):
Only one blocking comment to delete unused UDF. Let's handle the `two_or_more_races_pop` issue in a separate PR to get this in.

raphaellaude (approved):
lgtm

PR #171, opened 2024-11-12
Paint by county follow up - heal/unshatter bug fix

nofurtherinformation:
When painting by county, there is a bug that broken parents will not be checked unless they are included in the paint event. 

## Description
- Refactors paint feature filters to a more reusable function
- Slight optimizations

## Reviewers
- Primary: @mapmeld 
- Secondary: @raphaellaude 

## Checklist
- [ ] Refactor filters
- [ ] Apply to all paint functions
- [ ] Optimization: combine loops of filters

### DIFF
--- a/app/src/app/utils/helpers.ts
+++ b/app/src/app/utils/helpers.ts
@@ -81,7 +81,7 @@ export const getFeaturesInBbox = (
   filterLocked: boolean = true
 ): MapGeoJSONFeature[] | undefined => {
   const bbox = boxAroundPoint(e, brushSize);
-  const {captiveIds, lockedFeatures, mapDocument, checkParentsToHeal, shatterIds, isPainting} =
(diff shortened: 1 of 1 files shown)


mapmeld:
The refactor makes sense that the features go through this same function. How do you decide which ones are additional filters and which are filters saved in the mapStore?

raphaellaude (approved):
Lgtm

nofurtherinformation:
> The refactor makes sense that the features go through this same function. How do you decide which ones are additional filters and which are filters saved in the mapStore?

I chose the three filters within the function should basically always be checked (locked features/captive features, and parents to heal) regardless of application state, and others (like county FIPS checks) are added as needed. These could be structured a bit differently. wdyt?

mapmeld (approved):
:shipit:

Issue #169, opened 2024-11-12
Add option to highlight un-assigned areas


nofurtherinformation:
Highlight areas that have not yet been assigned to a zone. Style should be something like:

```javascript
...
'line-color': [
        'case',
        ['boolean', ['feature-state', 'focused'], false],
        '#000000', // Black color when focused
        ['boolean', ['feature-state', 'highlighted'], false],
        '#e5ff00', // yellow color when highlighted
        ['==', ['feature-state', 'zone'], null],
        '#e5ff00', // yellow color when zone is not assigned
        '#000000', // Default color
      ],
      'line-width': [
        'case',
        ['boolean', ['feature-state', 'focused'], false],
        5, // Width of 5 when focused
        ['boolean', ['feature-state', 'highlighted'], false],
        5, // Width of 5 when highlighted
        ['==', ['feature-state', 'zone'], null],
        5,
        0, // Default width
      ],
...
```

nofurtherinformation:
Resolved in #173

PR #173, opened 2024-11-13
Highlight unassigned geographies

nofurtherinformation:
#169 

## Description
- Adds option to highlight un assigned geometries to make them easier to find and paint

## Reviewers
- Primary: @mapmeld 
- Secondary: @raphaellaude 

## Checklist
- [x] Update `mapOptions` type
- [x] Update highlight layer specification to optionally show these ares
- [x] Add subscription to mapOptions to update layer rendering based on options
- [x] Add checkbox in sidebar with this option
- [x] Added child layer highlight, when present

Screenshot 2024-11-13 at 1 27 48 PM

### DIFF
--- a/app/src/app/constants/layers.ts
+++ b/app/src/app/constants/layers.ts
@@ -132,7 +132,6 @@ export function getHighlightLayerSpecification(
   layerId: string,
   highlightUnassgned?: boolean
 ): LineLayerSpecification {
-  const [unassignedHighlightColor, unassignedHighlightWidth] = highlightUnassgned ? ['#FF0000', 5] : ['#000000', 0]
   return {
     id: layerId,
     source: BLOCK_SOURCE_ID,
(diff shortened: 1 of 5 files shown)


app/src/app/constants/layers.ts
```diff
@@ -256,24 +266,17 @@ export function removeBlockLayers(map: Map | null) {
...
+  [
+    BLOCK_LAYER_ID,
+    BLOCK_LAYER_ID_HIGHLIGHT,
+    BLOCK_HOVER_LAYER_ID,
+    BLOCK_LAYER_ID_CHILD,
+    BLOCK_HOVER_LAYER_ID_CHILD,
+    BLOCK_LAYER_ID_HIGHLIGHT_CHILD,
+    BLOCK_SOURCE_ID,
+  ].forEach(layer => {
+    map.getLayer(layer) && map.removeLayer(layer);
```
raphaellaude, line 278: nice – much better!

raphaellaude (approved):
Looks good and works well – only thing I'd app perhaps is making the stroke width a little thinner and using round caps.

nofurtherinformation:
> Looks good and works well – only thing I'd app perhaps is making the stroke width a little thinner and using round caps.

Thanks! I just reduced the line width and cleaned up the style expression a little. The cap is already rounded, although it does look a little rough. The challenge he is that we cannot combine map style specification `case`  and `interpolate` so there is no way to get a smaller line when zoomed out and also check some additional logic

raphaellaude (approved):
> The challenge here is that we cannot combine map style specification case and interpolate so there is no way to get a smaller line when zoomed out and also check some additional logic

Ah makes sense. Lgtm!

Issue #176, opened 2024-11-13
Individual locked features unlock erroneously 


nofurtherinformation:
Currently, individual locked features become unlocked on various state actions. Additional logic is needed to preserve individually locked features

nofurtherinformation:
Resolved in #179 #177

PR #177, opened 2024-11-13
Fix logic for checking individual locked areas

nofurtherinformation:
#176 

## Description
- Bug fix
- Adds the following logic

When setting locked features after a zone assignment change, app will check if the zone assignment is:
- Included in the locked zones
- Not null
- When locked areas did not change, was the feature already in the locked features
- When locked areas did change, was the zone previously locked and is now unlocked and was the feature already in the locked features

This allows individually locked features to remain locked while doing other business, AND allows individual features in a locked zone to be unlocked and painted 

## Reviewers
- Primary: @raphaellaude 
- Secondary: @mapmeld 

## Checklist
- [x] Additional logic in lock subscription

### DIFF
--- a/app/src/app/store/mapRenderSubs.ts
+++ b/app/src/app/store/mapRenderSubs.ts
@@ -102,8 +102,14 @@ export const getRenderSubscriptions = (useMapStore: typeof _useMapStore) => {
     ],
     (curr, prev) => {
       colorZoneAssignments(curr, prev);
-      const {captiveIds, shatterIds, getMapRef, setLockedFeatures, mapRenderingState} =
-        useMapStore.getState();
+      const {
+        captiveIds,
+        shatterIds,
+        getMapRef,
+        setLockedFeatures,
+        lockedFeatures,
+        mapRenderingState,
+      } = useMapStore.getState();
       const mapRef = getMapRef();
       if (!mapRef || mapRenderingState !== 'loaded') return;
       [...PARENT_LAYERS, ...CHILD_LAYERS].forEach(layerId => {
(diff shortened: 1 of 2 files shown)


app/src/app/store/mapRenderSubs.ts
```diff
@@ -128,15 +136,29 @@ export const getRenderSubscriptions = (useMapStore: typeof _useMapStore) => {
...
+                lockPaintedAreas.includes(value) ||
+                // locked zones are the same, and this individual feature was previously locked
+                (sameLockedAreas && lockedFeatures.has(key)) ||
+                // locked zones are changed, BUT this individual feature is not in a zone
+                // that was previously locked
+                (!sameLockedAreas &&
+                  previousWasArray &&
+                  !lockPaintedAreas.includes(value) &&
+                  !prevLockPaintedAreas.includes(value) &&
+                  lockedFeatures.has(key))
```
raphaellaude, line 157: PP: Are all of these sets? I wonder if you could use boolean logic–i.e. `intersection`, `difference`, `union`–instead of destructuring `zoneAssignments` and iterating through each element. Anyway probably negligible perf benefits if any

raphaellaude (approved):
lgtm just one non-blocking comment

app/src/app/store/mapRenderSubs.ts
raphaellaude: PS. Remembering that `zoneAssignments` is a Map not a Set (duh!) so you can ignore the part of my comment about destructuring `zoneAssignments` as that would have to be done either way

raphaellaude: (And since this is an active bug we should merge and can always come back to this if it's hurting perf, which I doubt it is)

PR #180, opened 2024-11-14
Bugfix - remove source

nofurtherinformation:
Bug introduced in refactor that missed removing the maplibre source on remove layers. Current refactor is more readable for future changes.

## Description
- In `removeBlockLayers` adds a second loop for sources that calls `removeSource`
- A future refactor should centralize layer and source manipulations in a single class or store to make it more consistent and ergonomic 

## Reviewers
- Primary: @raphaellaude

### DIFF
--- a/app/src/app/constants/layers.ts
+++ b/app/src/app/constants/layers.ts
@@ -279,6 +279,12 @@ export function removeBlockLayers(map: Map | null) {
   ].forEach(layer => {
     map.getLayer(layer) && map.removeLayer(layer);
   });
+  
+  [
+    BLOCK_SOURCE_ID,
+  ].forEach(source => {
+    map.getSource(source) && map.removeSource(source);
+  });
 }
 
 export {addBlockLayers};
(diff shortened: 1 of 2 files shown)


app/src/app/constants/layers.ts
```diff
@@ -275,10 +280,13 @@ export function removeBlockLayers(map: Map | null) {
     BLOCK_LAYER_ID_CHILD,
     BLOCK_HOVER_LAYER_ID_CHILD,
     BLOCK_LAYER_ID_HIGHLIGHT_CHILD,
-    BLOCK_SOURCE_ID,
   ].forEach(layer => {
     map.getLayer(layer) && map.removeLayer(layer);
   });
+
+  [BLOCK_SOURCE_ID].forEach(source => {
```
nofurtherinformation, line 287: This is the actual bugfix, rest is just prettier

raphaellaude (dismissed):

nofurtherinformation:
P.s.  @raphaellaude  added one small refactor, the fit to bounds was acting slightly buggy so I pulled that into the `setMapDocument` state method

raphaellaude (approved):

PR #183, opened 2024-11-18
Soft delete DistrictrMap

raphaellaude:
## Description

- Ahead of testing it would be nice to not have to delete `DistrictMaps` from the DB but rather soft delete them
- This PR also cleans up the create_districtr_map_udf, which wasn't actually being applied! The extent logic from what I can tell is valid but adding it in causes all sorts of trouble with the tests

## Reviewers

- Primary: @nofurtherinformation @mariogiampieri 
- Secondary:

## Checklist


nofurtherinformation (approved):
Lgtm!

Issue #178, opened 2024-11-14
No handling of missing document


raphaellaude:
Can't believe I never thought of this but if there's a document id that's missing we don't do... anything? We just leave it and drop errors in the console.

We should probably show an error message ("Your map could not be found") then clear the document id.

PR #185, opened 2024-11-18
Various bug fixes

nofurtherinformation:
This PR has a few bug fixes for previous flagged issues.

## Description
- #178 
- #187 
- #188 
- #194 
- #197 
- #201 

## Reviewers
- Primary: @raphaellaude 
- Secondary: @mariogiampieri 

## Checklist
- [x] Add error notification state 
- [x] Add error notification component


app/src/app/components/ErrorNotification.tsx
```diff
@@ -0,0 +1,78 @@
...
+    <>
+      <Text as="p" mb={'4'}>
+        {errorNotification.message}
+      </Text>
+      <Text as="p" mb={'4'}>
+        {!!errorNotification?.id && <i>Error ID: {errorNotification.id}</i>}
+      </Text>
+      <Text as="p" mb={'4'}>
+        {!!errorNotification?.severity && <i>Severity: {errorNotification.severity}</i>}
+      </Text>
```
raphaellaude, line 28: PP: Not sure we need the severity here.

Q: Do you know if Sentry will capture these errors? It would be good to still have them tracked there but not pressing to figure that out

app/src/app/map/page.tsx
```diff
@@ -16,6 +17,7 @@ export default function Map() {
           <MapComponent />
           <MobileTopNav />
           <MapContextMenu />
+          <ErrorNotification />
```
raphaellaude, line 20: Q: Should this be further up in the DOM in case other pages eventually need it?

app/src/app/components/ErrorNotification.tsx
```diff
@@ -0,0 +1,78 @@
+"use client"
+import {useEffect, useState} from 'react';
+import {useMapStore} from '../store/mapStore';
+import {AlertDialog, Button, Flex, Text} from '@radix-ui/themes';
+import * as Toast from '@radix-ui/react-toast';
+
+export const ErrorNotification = () => {
```
raphaellaude, line 7: PP: Good for now but there's probably an argument to be made either way that this should be an error dialog.

app/src/app/store/mapStore.ts
```diff
@@ -455,7 +473,17 @@ export const useMapStore = create(
                 document_id,
                 geoids,
               });
-
+          if (!shatterResult.children.length){
+            const mapDocument = get().mapDocument
+            set({
+              errorNotification: {
+                severity: 2,
+                message: `Breaking this geography failed. Please refresh this page and try again. If this error persists, please share the error code below the Districtr team.`,
```
raphaellaude, line 481: nice

raphaellaude (approved):
LGTM and everything was working nicely in https://districtr-v2-185-app.fly.dev/

app/src/app/map/page.tsx
nofurtherinformation: Possibly but I'm not sure if that will mess up SSR stuff

app/src/app/components/ErrorNotification.tsx
nofurtherinformation: Severity 1 is a dialog!

app/src/app/components/ErrorNotification.tsx
nofurtherinformation: I'm using severity as a way to sort alert type - dialog, toast, silent log ?

nofurtherinformation: I'm not super familiar with sentry but I'm sure this can integrate with it

app/src/app/store/mapStore.ts
nofurtherinformation: Thank you!

app/src/app/components/ErrorNotification.tsx
raphaellaude: Makes sense – just not sure it needs to be exposed to the user!

PR #199, opened 2024-11-20
Paint by county with ID instead of features

nofurtherinformation:
Instead of painting by county using features, this PRs drafts a method of caching all IDs and then filtering based on those

## Description
- Cache IDs of features on `data` in the map
- When painting by county, search those IDs

## Reviewers
- Primary: @raphaellaude 
- Secondary: @mariogiampieri

### DIFF
--- a/app/src/app/store/mapRenderSubs.ts
+++ b/app/src/app/store/mapRenderSubs.ts
@@ -11,6 +11,7 @@ import {
   BLOCK_LAYER_ID_HIGHLIGHT,
   getHighlightLayerSpecification,
   BLOCK_LAYER_ID_HIGHLIGHT_CHILD,
+  COUNTY_LAYERS,
 } from '../constants/layers';
 import {
   ColorZoneAssignmentsState,
@@ -310,6 +311,18 @@ export const getRenderSubscriptions = (useMapStore: typeof _useMapStore) => {
       } 
     }
   );
+  
+  const filterCountiesSub = useMapStore.subscribe<[string|undefined, MapStore['getMapRef']]>(state => [state.mapOptions.currentStateFp, state.getMapRef],
+    ([stateFp, getMapRef]) => {
+      const mapRef = getMapRef()
+      if (!mapRef) return
+      const filterExpression = (stateFp ? ["==", "STATEFP", stateFp] : true) as any
+      COUNTY_LAYERS.forEach(layer => {
+        mapRef.setFilter(layer,  ["any", filterExpression])
+      })
+    }
+  )
+  
   return [
     addLayerSubMapDocument,
     _shatterMapSideEffectRender,
@@ -318,5 +331,6 @@ export const getRenderSubscriptions = (useMapStore: typeof _useMapStore) => {
     _updateMapCursor,
     _applyFocusFeatureState,
     highlightUnassignedSub,
+    filterCountiesSub
   ];
 };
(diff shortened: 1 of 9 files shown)


nofurtherinformation:
Also filters counties to only active state

app/src/app/store/mapRenderSubs.ts
```diff
@@ -303,6 +311,18 @@ export const getRenderSubscriptions = (useMapStore: typeof _useMapStore) => {
...
   );
+  
+  const filterCountiesSub = useMapStore.subscribe<[string|undefined, MapStore['getMapRef']]>(state => [state.mapOptions.currentStateFp, state.getMapRef],
+    ([stateFp, getMapRef]) => {
+      const mapRef = getMapRef()
+      if (!mapRef) return
+      const filterExpression = (stateFp ? ["==", "STATEFP", stateFp] : true) as any
+      COUNTY_LAYERS.forEach(layer => {
+        mapRef.setFilter(layer,  ["any", filterExpression])
+      })
```
raphaellaude, line 322: H: Need to check that the county layer is added first

image

app/src/app/utils/helpers.ts
```diff
@@ -133,16 +134,25 @@ export const getFeaturesIntersectingCounties = (
...
-  });
-
-  return filterFeatures(
-    features,
-    true,
-    [(feature) => Boolean(feature?.id && feature.id.toString().match(/\d{5}/)?.[0] === fips)]
-  );
+  return filterFeatures([...cachedParentFeatures, ...childFeatures], true, [
+    feature => Boolean(feature?.id && feature.id.toString().match(/\d{5}/)?.[0] === fips),
+  ]);
```
raphaellaude, line 155: Random comment but we should probably start testing the FE, esp. would be easy to add unit tests for little utilities like these even if we don't go full FE testing w/ a framework

raphaellaude (changes requested):
Sort of rushed through this review so apologies it's not more in depth. Only one comment re: unavailable county layers breaking things

app/src/app/store/mapRenderSubs.ts
nofurtherinformation: Ah yep

app/src/app/utils/helpers.ts
raphaellaude: Here's an issue for that #208

app/src/app/store/mapRenderSubs.ts
```diff
@@ -308,6 +316,18 @@ export const getRenderSubscriptions = (useMapStore: typeof _useMapStore) => {
...
   );
+  
+  const filterCountiesSub = useMapStore.subscribe<[string|undefined, MapStore['getMapRef']]>(state => [state.mapOptions.currentStateFp, state.getMapRef],
+    ([stateFp, getMapRef]) => {
+      const mapRef = getMapRef()
+      if (!mapRef) return
+      const filterExpression = (stateFp ? ["==", "STATEFP", stateFp] : true) as any
+      COUNTY_LAYERS.forEach(layer => {
+        mapRef.getLayer(layer) && mapRef.setFilter(layer,  ["any", filterExpression])
+      })
```
raphaellaude, line 327: I: This doesn't seem to be capturing the point label id

image

raphaellaude (changes requested):
almost there!

app/src/app/store/mapRenderSubs.ts
nofurtherinformation: Had a typo in the layer name but should be good now

raphaellaude (approved):
Lgtm

Issue #98, opened 2024-09-25
Zoom to unassigned


mariogiampieri:
Goal would be to help user focus on unassigned areas.

nofurtherinformation:
In progress in #223

Issue #96, opened 2024-09-25
District numbers appear on map


mariogiampieri:
We have a placeholder checkbox for this but need to implement

nofurtherinformation:
Added in #210

PR #210, opened 2024-11-21
District numbers and outlines

nofurtherinformation:
Adds zone numbers to map. Hacky approach, but mostly works

## Description
- If selected, this option grabs the rendered map features, dissolves the geometries based on the zone assignment, and adds labels to the map with the zone numbers

## Reviewers
- Primary: @raphaellaude 
- Secondary: @mariogiampieri


app/src/app/constants/layers.ts
```diff
@@ -205,8 +222,28 @@ export function getBlocksLayerSpecification(
...
+        'interpolate',
+        ['exponential', 1.6],
+        ['zoom'],
+        6,
+        lineWidth * 0.125,
+        9,
+        lineWidth * 0.35,
+        14,
+        lineWidth,
+      ],
```
nofurtherinformation, line 246: This was just code styling

nofurtherinformation:
I've taken another pass to optimize and clean up the way this works a bit. Previously, this functionality added a bunch of `map.queryRenderedFeatures` calls, which are generally not very good.

Now, whenever a tile loads in, the event listener (`map.on("data")`) caches the IDs for painting by county and sends the geometry to the geometry worker (WebWorker, off-main-thread) and that caches the geometries if not already present. We keep track of the x-y-z index of the tile, so there isn't a ton of work duplicated. The only additional work here is some stringification for data sent from the main thread to the worker, but it is relatively minimal.

Whenever the assignments updates, the results get sent to the worker which mutates its copy of the data with the updated zones. When zone districts are toggled on or the viewport changes, the worker uses `bboxClip` followed by `dissolve(by="zone"`) to get zone geometries and center of mass. The center of mass is checked to make sure its actually in the geometry, and if not a random point in the geometry is chosen.

I think there are some additional use cases that will be good to have this amount of geometric data available on the client, and having it on a separate thread means that UX performance impact is minimal. I'm fairly convinced this is a reasonable direction, especially in the interest of keeping the backend lightweight -- eager to hear feedback from you all @raphaellaude @mariogiampieri !

app/src/app/utils/GeometryWorker/geometryWorker.ts
```diff
@@ -0,0 +1,110 @@
...
+        propertyName: 'zone',
+      }
+    );
+    let largestDissolvedFeatures: Record<number, {feature: GeoJSON.Feature; area: number}> = {};
+
+    dissolved.features.forEach(feature => {
+      const zone = feature.properties?.zone;
+      if (!zone) return;
+      const featureArea = area(feature);
+      if (!largestDissolvedFeatures[zone] || featureArea > largestDissolvedFeatures[zone].area) {
```
mariogiampieri, line 62: O: this makes sense for now given that we are not enforcing contiguity on zones, but could likely be refactored later when that rule is enforced

mariogiampieri (changes requested):
This is super slick- it makes a lot of sense to create that separate thread and the user experience was great in my testing. I agree there could be other use cases for a separate worker like- nice work!

Having the dissolved zone areas is nice- I could see a world where we want to outline the zones. 

Only minor issue is removing the unused imports but otherwise g2g!

app/src/app/constants/layers.ts
```diff
@@ -10,6 +10,10 @@ import {getBlocksSource} from './sources';
 import {DocumentObject} from '../utils/api/apiHandlers';
 import {MapStore, useMapStore} from '../store/mapStore';
 import {colorScheme} from './colors';
+import {map, throttle} from 'lodash';
+import {wrap} from 'comlink';
```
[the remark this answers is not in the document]
nofurtherinformation: Good catch, thank you!

app/src/app/utils/GeometryWorker/geometryWorker.ts
nofurtherinformation: Agreed! I added this note in a `TODO` along with a couple other future directions we could add

nofurtherinformation:
Thank you @mariogiampieri! I just updated the unused imports and added a couple notes to our future selves.

mariogiampieri (dismissed):
lgtm, nice work!

mariogiampieri (approved):
lgtm! great work

raphaellaude (approved):
lgtm!

raphaellaude:
PS This closes a few issues:
- #96 
- #98

PR #211, opened 2024-11-22
Text Updates

nofurtherinformation:
Small final text updates ahead of user testing


mariogiampieri (approved):
lgtm! looks like there may be a code formatting issue again, something to return to and sync up on after testing

PR #217, opened 2024-12-03
Population chart enhancements

nofurtherinformation:
This PR enhances the population chart and makes ancillary updates related to this. The main goals, related to feedback from our users are to:

1. Make the population chart respond faster and give more immediate feedback
2. Label bars with population values district numbers
3. Scale the x-axis to focus in on detailed population balancing
4. Add a deviation target indicator
5. Add on-hover population indicators

## Description
To handle these requirements, the following changes in this PR:
- Swap from Re-charts to Visx for charting (I tried with Re-charts but the performance and control is not good enough)
- Change map metrics on paint based on features 
- Add map option to show population of hovered geographies
- Clean up events in the mapstore when brushing -- this has some performance gains in reverting to using `accumulatedGeoids: Set<string>` and only periodic state updates
- Periodically updating the state has the problem that we don't know the previous assignment of a geography (unless we track it in state) but after some reading we are able to dive into the feature state of the map a bit more easily than previously thought
- [code org / devex] Refactor state to make debugging and observation easier
- [code org] Refactor population sidebar files a bit more discretely

## Reviewers
- Primary: @mariogiampieri 
- Secondary: @raphaellaude 

## Checklist
- [x] Refactor from recharts to Visx
- [x] Handle population change logic on feature selection
- [x] Clean up map store events
- [x] Add map option to show population on tooltip
- [x] Refactor stores for overall cleaner devex and debugging (state and edge cases, performance, etc.)


app/src/app/components/sidebar/PopulationPanel/PopulationChart/PopulationLabels.tsx
```diff
@@ -0,0 +1,48 @@
...
+      left={xScale(entry.total_pop)}
+      top={yScale(index) + barHeight}
+      style={{pointerEvents: 'none'}}
+    >
+      {!!(isHovered || showPopNumbers) && (
+        <text x={5} y={-2} fontSize={14} fontWeight={'bold'} textAnchor="start">
+          {popLabel}
+        </text>
+      )}
+      {!!(isHovered || showTopBottomDeviation) && (
```
mariogiampieri, line 39: these look great in general- I did notice that for bars with small pops, the labels often fall off the side of the bar. What about checking the label width and not showing if it's below a certain threshold? 
```
const barX = xScale(entry.total_pop);
const labelWidth = 50; // or whatever; this worked in a quick test
const isPopDiffLabelVisible = barX - labelWidth > 0;
```
and then adding `isPopDiffLabelVisible` as a condition to render?

app/src/app/store/hoverFeaturesStore.ts
mariogiampieri, line 1: Is this being here / empty intentional?

app/src/app/store/mapStore.ts
```diff
@@ -336,48 +324,58 @@ export const useMapStore = create(
...
             selectedZone: _selectedZone,
-            zoneAssignments,
           } = get();
 
           const map = getMapRef();
           const selectedZone = activeTool === 'eraser' ? null : _selectedZone;
           if (!map || !mapDocument?.document_id) {
             return;
           }
+          // We can access the inner state of the map in a more ergonomic way than the convenience method `getFeatureState`
```
mariogiampieri, line 338: O: this is super cool! nice work

mariogiampieri (changes requested):
This is great overall- the only thing blocking IMO is implementing something to prevent bar labels from falling off the bars.

app/src/app/store/mapStore.ts
nofurtherinformation: Thank you!

app/src/app/components/sidebar/PopulationPanel/PopulationChart/PopulationLabels.tsx
nofurtherinformation: This is a great catch, thank you! I added a check as you suggested in the [labels components](https://github.com/districtr/districtr-v2/blob/e72508803d0bf5e68b1eab799e9e559c61766557/app/src/app/components/sidebar/PopulationPanel/PopulationChart/PopulationLabels.tsx#L33-L38) that nudges the label group right if the bar is too short, and left in the probably unusual case that it is too long. 

Screenshot 2024-12-17 at 9 34 26 AM

We _may_ need additional descriptive text to explain what these numbers mean, but my instinct is to keep it clean and get some user feedback

mariogiampieri (approved):
lgtm!

PR #218, opened 2024-12-03
Fix pinch speed

nofurtherinformation:
The pinch gesture on trackpad feels to slow. However, the mouse wheel or two finger scroll feels fine, and the events are hard to disambiguate. After some reading, there is a de facto standard to indicate a pinch event, which is the event logs `ctrlKey` as held down. 

## Description
- Adds a map listener that changes the speed of zoom on pinch/wheel
- Fix county label filter issue


raphaellaude (dismissed):
lgtm

nofurtherinformation:
@raphaellaude Merged with main, pls re-approve at your convenience!

raphaellaude (approved):
Lgtm

PR #220, opened 2024-12-09
Remove deprecated setZoneAssignments

mapmeld:
Mentioned this on Slack - now that we moved to an event listener, `setZoneAssignments` is only called by the zone picker. We don't think this is necessary anymore now that the mouse exiting the map should apply the changes.

The calls were added to the picker in this commit and PR:
https://github.com/districtr/districtr-v2/commit/8dc9328f2d7d926b497b8b0adf52d84d74f4fcc2
https://github.com/districtr/districtr-v2/pull/129 (likely copying the Picker)

### DIFF
--- a/app/src/app/components/sidebar/Picker.jsx
+++ b/app/src/app/components/sidebar/Picker.jsx
@@ -5,12 +5,9 @@ import {useMapStore} from '../../store/mapStore';
 export function ZoneTypeSelector() {
   const selectedZone = useMapStore(state => state.selectedZone);
   const setSelectedZone = useMapStore(state => state.setSelectedZone);
-  const setZoneAssignments = useMapStore(state => state.setZoneAssignments);
   const accumulatedGeoids = useMapStore(state => state.accumulatedGeoids);
 
   const handlePickerValueChange = value => {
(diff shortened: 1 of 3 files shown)


nofurtherinformation:
Hey @mapmeld ! Thanks for this! 

I reviewed and realized that the addition of zone updates within `mapStore.selectMapfeatures` was an error I introduced. It works and keeps the state highly synced with the map, but causes some performance and excess request issues. 

I think the right way to handle zone updates (for now at least) is to only [mutate the `accumulatedGeoids` set and handle any map rendering and tile logic (eg. clientside population)](https://github.com/districtr/districtr-v2/blob/716fc18c70f060d6a56e1ec23948d1ff735389d9/app/src/app/store/mapStore.ts#L356-L376) and then at the [end of a paint event update the zone assignments.](https://github.com/districtr/districtr-v2/blob/716fc18c70f060d6a56e1ec23948d1ff735389d9/app/src/app/store/mapStore.ts#L865-L871). This direction made more sense in the population chart enhancements and I think will simplify undo/redo in terms of number of events firing off.

PR #221, opened 2024-12-19
Undo redo

nofurtherinformation:
This PR contains collaborative work from @mapmeld and myself. The goal of this PR is to add undo-redo capabilities without adding a large number of lines of code. To do this, the PR uses Zundo, an undo/redo middleware for zustand. The complexity of our use case is handling heal and break (nee shatter) operations elegantly without adding intermediary states in between and make sure the changes sent to the backend remain correct. 

## Description
- Adds Zundo middleware
- Adds undo/redo buttons
- Refactors middleware to single wrapper
- Adds handlers to silently heal and silently break geographies

## Reviewers
- Primary: @raphaellaude 
- Secondary: @nofurtherinformation 

## Checklist
- [x] Undo/redo on normal paint operations
- [x] Undo/redo on break
- [x] Undo/redo on heal
- [x] Undo/redo buttons and UI
- [ ] Test UX and speed on slower machines


app/src/app/store/mapEditSubs.ts
```diff
@@ -6,6 +6,8 @@ import {shallowCompareArray} from '../utils/helpers';
 import {updateAssignments} from '../utils/api/queries';
 import {queryClient} from '../utils/api/queryClient';
 
+export let allowSendZoneUpdates = true
```
mapmeld, line 9: This seems very un-React-y but as I understand it keeps additional updates from occurring during shatter/heal events. If it's just for shattering, it might be helpful to put a comment here

mapmeld:
Compiling some edge cases:
- Reset Map should clear undo/redo history
- If I open a fresh map or load a saved map, and draw something, the undo button might be disabled too long and prevent undoing the first draw (this was an issue on older undo/redo button logic, too)
- if I slowly draw a squiggle over a few seconds, it shows up as multiple events in undo/redo (likely debounce / timing)
- if I erase or draw some things, switch tabs, and return, I see the map's old state flash and then update to current state; it's not just rendering because they can be undone in one undo event

app/src/app/store/mapEditSubs.ts
nofurtherinformation: PP: Agreed, we should add a comment here. Something like:

```javascript
// We use basic plain ol javascript variables here to ensure that any
// asynchronous operations / subscriptions on the state.
// This is set to false when operations are occurring that heal or break vtds
// during an undo/redo operation. This can create intermediary state changes to the zone updates
// so it's important that we avoid sending those to the backend.
export let allowSendZoneUpdates = true

**Review by nofurtherinformation — COMMENTED** [review]
Overall looks good! I added one blocking code org comment but otherwise I think looks good to go. 

@raphaellaude Do you have time to look at this with fresh eyes?

File: app/src/app/components/sidebar/ResetMapButton.tsx [hunk 2]
```diff
@@ -1,9 +1,21 @@
 import {useMapStore} from '@/app/store/mapStore';
 import {AlertDialog, Button, Flex} from '@radix-ui/themes';
+import {useStore} from 'zustand';
+
+/* convert zundo to a React hook */
+const useTemporalStore = <T,>(
```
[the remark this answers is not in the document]
mapmeld: ✅  I put into `store/temporalStore`

app/src/app/utils/api/queries.ts
```diff
@@ -127,7 +127,12 @@ export const updateAssignments = (mapDocument: DocumentObject) => {
 
 fetchAssignments.subscribe(assignments => {
   if (assignments.data) {
-    useMapStore.getState().loadZoneAssignments(assignments.data);
+    const {loadZoneAssignments, appLoadingState} = useMapStore.getState();
+    if (appLoadingState === 'initializing') {
+      loadZoneAssignments(assignments.data);
+    } else {
+      console.log('did not load prior map during rendering state: ' + appLoadingState);
```
nofurtherinformation, line 134: Q: @mapmeld with the previous condition based on map rendering state, any time the layers rendered before the assignments came back it would prevent loading the assignments. I swapped this to the app loading state instead, which seems to have the desired behavior.

Was this intended for tab blur/refocus preventing re-loading assignments? I was getting some confusing behaviors with this, but let me know if I missed something!

mapmeld: Your update to use appLoadingState does address the tab blur/refocus bug where it flashed an earlier map. 
I am seeing a weird bug where changes _after_ tab blur/refocus are merged into one undoable action, even if the changes are minutes apart, likely this is an equality function issue?

nofurtherinformation: Hmm, strange. I'm not sure how to address this for now. The previous update to this stopped any assignments loading on refresh

mapmeld: removed a line from my previous work on equality/diff function and that seems to fix it, maybe no longer needed after `handleVisibilityChange` does pause and resume

nofurtherinformation: Nice!

mapmeld:
ok I had a chance to review drawing, erasing, county paint, blur/refocus, and a scan through the code for any new stuff; as best I can tell this is working / demo ready

nofurtherinformation:
> ok I had a chance to review drawing, erasing, county paint, blur/refocus, and a scan through the code for any new stuff; as best I can tell this is working / demo ready

Same, I did a run through of everything and seems to be stable!

app/src/app/utils/api/queries.ts
mariogiampieri: I did have the previous state appear after refreshing once or twice while testing but am not able to reproduce on purpose. I had been testing various combinations of paint -> undo, erase -> undo, and refreshing / changing tabs interchangeably. I only notice the erroneous states on refresh (I think) and not while switching tabs

app/src/app/store/mapStore.ts
```diff
@@ -38,6 +38,8 @@ import {parentIdCache} from './idCache';
 import {getMapMetricsSubs} from './metricsSubs';
 import {queryClient} from '../utils/api/queryClient';
 import { useChartStore } from './chartStore';
+import { createWithMiddlewares } from './middlewares';
+import { map } from 'lodash';
```
mariogiampieri, line 42: PP: remove unused imports

mariogiampieri:
> Compiling some edge cases:
> 
> * Reset Map should clear undo/redo history
> * If I open a fresh map or load a saved map, and draw something, the undo button might be disabled too long and prevent undoing the first draw (this was an issue on older undo/redo button logic, too)
> * if I slowly draw a squiggle over a few seconds, it shows up as multiple events in undo/redo (likely debounce / timing)
> * if I erase or draw some things, switch tabs, and return, I see the map's old state flash and then update to current state; it's not just rendering because they can be undone in one undo event

H: I cannot undo the first paint event or any events that precede the first paint (for example breaking). Confirming that I can paint -> break -> paint and then undo back to the first paint event (at which point the undo button becomes disabled)

app/src/app/store/temporalStore.ts
```diff
@@ -0,0 +1,11 @@
+import type {TemporalState} from "zundo";
+import {useStore} from 'zustand';
+import {useMapStore, type MapStore} from '@/app/store/mapStore';
+
+// Convert zundo to a React hook
+// from https://github.com/charkour/zundo?tab=readme-ov-file#for-reactive-changes-to-member-properties-of-the-temporal-object-optionally-convert-to-a-react-store-hook
+
+export const useTemporalStore = <T,>(
```
mariogiampieri, line 8: very cool!

app/src/app/store/middlewareConfig.ts
```diff
@@ -28,3 +29,34 @@ export const devToolsConfig: DevtoolsOptions = {
...
+
+export const temporalConfig: ZundoOptions<any, MapStore> =  {
+  // If diff returns null, not state is stored
+  diff: (past: Partial<MapStore>, curr: Partial<MapStore>) => {
+    // if not yet loaded, or is a temporal action (eg. silent heal) don't store
+    if (past.mapRenderingState !== 'loaded' || curr.isTemporalAction) return null;
+    // if current state has no zoneAssignments, don't store
+    if (curr.zoneAssignments?.size === 0) return null;
+    const pastAssignments = past.zoneAssignments || new Map();
+    const currAssignments = curr.zoneAssignments || new Map();
```
mariogiampieri, line 41: Q: the right side of this can never happen right?

app/src/app/components/sidebar/ResetMapButton.tsx
```diff
@@ -1,9 +1,16 @@
-import {useMapStore} from '@/app/store/mapStore';
 import {AlertDialog, Button, Flex} from '@radix-ui/themes';
+import {useStore} from 'zustand';
```
mariogiampieri, line 2: PP: remove unnecessary imports

app/src/app/store/mapStore.ts
```diff
@@ -521,6 +573,7 @@ export const useMapStore = create(
               parents: existingParents,
               children: existingChildren,
             },
+            isTemporalAction: false,
```
mariogiampieri, line 576: Q: does this need to be true, or is the shatter action otherwise protected during undo/redo actions?

app/src/app/utils/api/queries.ts
mariogiampieri: Some instances I observe where there is odd behavior:
- was painting, reloaded the page to a blank plan, started painting again and the previous plan state returned

mariogiampieri:
> > Compiling some edge cases:
> > 
> > * Reset Map should clear undo/redo history
> > * If I open a fresh map or load a saved map, and draw something, the undo button might be disabled too long and prevent undoing the first draw (this was an issue on older undo/redo button logic, too)
> > * if I slowly draw a squiggle over a few seconds, it shows up as multiple events in undo/redo (likely debounce / timing)
> > * if I erase or draw some things, switch tabs, and return, I see the map's old state flash and then update to current state; it's not just rendering because they can be undone in one undo event
> 
> H: I cannot undo the first paint event or any events that precede the first paint (for example breaking). Confirming that I can paint -> break -> paint and then undo back to the first paint event (at which point the undo button becomes disabled)

I see you fixed already- very nice!

app/src/app/utils/api/queries.ts
```diff
@@ -127,7 +141,12 @@ export const updateAssignments = (mapDocument: DocumentObject) => {
 
 fetchAssignments.subscribe(assignments => {
   if (assignments.data) {
-    useMapStore.getState().loadZoneAssignments(assignments.data);
+    const {loadZoneAssignments, appLoadingState} = useMapStore.getState();
+    if (appLoadingState === 'initializing') {
+      loadZoneAssignments(assignments.data);
```
raphaellaude, line 146: H: Will look through the code in more detail but the assignments don't seem to be updating properly with undos:

https://github.com/user-attachments/assets/3966179e-36c8-4a88-a692-da4569d7a71b

raphaellaude (changes requested):

app/src/app/utils/api/queries.ts
nofurtherinformation: This is a good catch, thank you for this Raph! Like we chatted about, but recording here for posterity... -- the issue was that the backend would overwrite old assignments on conflict, but if an assignment were no longer in the frontend update of assignments, the previously assigned zone would remain

The solution, until a better refactor, is to track any painted GEOID in a set, then while formatting the zone assignments to send to the back end, fill in any no longer present zones with nulls, as if they have been erased

app/src/app/store/mapStore.ts
nofurtherinformation: I _think_ this is ok as is, but it is a subtle difference. As is, it saves a little work comparing the past and previous sets of shatter ids, but does make an intermediate state where the VTD is broken/shattered and painted the parent color but none of the children are painted -- might be a way to refine this a bit

nofurtherinformation: I added a todo for future work on this!

app/src/app/store/middlewareConfig.ts
nofurtherinformation: I think that's right, since if the previous loading state was not fully loaded, it will already return. I cleaned up the conditions a bit though for clarity and the keep typescript happy

app/src/app/store/mapStore.ts
nofurtherinformation: Ty!

nofurtherinformation:
- [x]  Reset Map should clear undo/redo history
- [x]  If I open a fresh map or load a saved map, and draw something, the undo button might be disabled too long and prevent undoing the first draw (this was an issue on older undo/redo button logic, too)
- [ ] if I slowly draw a squiggle over a few seconds, it shows up as multiple events in undo/redo (likely debounce / timing)
_Not quite able to reproduce this one, although I saw some similar behavior here and there_
- [ ] if I erase or draw some things, switch tabs, and return, I see the map's old state flash and then update to current state; it's not just rendering because they can be undone in one undo event
_Trying to replicate and chase down this one..._

raphaellaude (approved):
Looks to be working much better! Let's merge for user testing and address any follow-on improvements / bug fixes in subsequent PRs

PR #228, opened 2024-12-29
Hovered features should reset after mouse leaves the map

mapmeld:
Bug was repeatable: having the cursor leave the map while painting leaves that area shaded until you resume painting, this can be a longer time if you change to the move cursor

Screenshot 2024-12-28 at 11 34 13 PM

This might have caused a colors issue here:

Proposed solution: `useHoverStore` gives us a function to clear the hover features; this requires a short timeout to avoid the map re-selecting these features

### DIFF
--- a/app/src/app/components/Map.tsx
+++ b/app/src/app/components/Map.tsx
@@ -7,7 +7,7 @@ import React, {useEffect, useRef} from 'react';
 import {MAP_OPTIONS} from '../constants/configuration';
 import {handleWheelOrPinch, mapContainerEvents, mapEvents} from '../utils/events/mapEvents';
 import {INTERACTIVE_LAYERS} from '../constants/layers';
-import {useMapStore} from '../store/mapStore';
+import {useHoverStore, useMapStore} from '../store/mapStore';
 import {MapTooltip} from './MapTooltip';
 
 export const MapComponent: React.FC = () => {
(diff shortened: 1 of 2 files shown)


app/src/app/components/Map.tsx
```diff
@@ -36,6 +37,12 @@ export const MapComponent: React.FC = () => {
   };
   useEffect(fitMapToBounds, [mapOptions.bounds]);
 
+  useEffect(() => {
+    map.current?.on('mouseout', () => {
+      setTimeout(() => setHoverFeatures([]), 250);
```
nofurtherinformation, line 42: H: Timeout approach seems to work here, but this should be in `mapEvents.ts` > `handleMapMouseOut. Unless this needs to rely on the component lifecycle?

nofurtherinformation (changes requested):
Just a code org suggestion but this approach works!

app/src/app/components/Map.tsx
mapmeld: :+1: looks like I can add a timeout to the two places where hover features are reset to the empty array, moved the changes into there

nofurtherinformation: Thank you!

nofurtherinformation (approved):
Splendid!

PR #229, opened 2024-12-30
Lock should prevent zone from expanding

mapmeld:
Modifies the function to filter out locked areas from painting, to also prevent adding new areas to a locked zone.

This was requested, but could be argued either way (is lock feature to freeze a zone completely or to prevent its existing parts from being overwritten?)


app/src/app/utils/helpers.ts
```diff
@@ -472,15 +472,17 @@ const filterFeatures = (
...
   const parentIdsToHeal: MapStore['parentsToHeal'] = [];
   const filterFunctions: Array<(f: MapGeoJSONFeature) => boolean> = [...additionalFilters];
   if (captiveIds.size) {
     filterFunctions.push(f => captiveIds.has(f.id?.toString() || ''));
   }
-  if (filterLocked && lockedFeatures.size) {
-    filterFunctions.push(f => !lockedFeatures.has(f.id?.toString() || ''));
+  if (filterLocked) {
+    filterFunctions.push(f => !lockedFeatures.has(f.id?.toString() || '') &&
+      mapOptions.lockPaintedAreas !== true && (mapOptions.lockPaintedAreas === false || !mapOptions.lockPaintedAreas.includes(selectedZone))
```
nofurtherinformation, line 484: Q: If this condition is met, why not do an early return of an empty array?

nofurtherinformation (commented):
I think this can wait until after user testing, but I added a suggestion to simplify things. Also, should we make a UI change to disable locked zone colors from being painted?

nofurtherinformation (dismissed):
@mapmeld I made some updates here and merged in main. Approving if this looks good to you

nofurtherinformation:
@mapmeld Cleaned this up a bit but looks good on my end. What do you think?

app/src/app/utils/helpers.ts
```diff
@@ -498,15 +498,35 @@ const filterFeatures = (
...
+    if (activeTool === 'brush' && mapOptions.lockPaintedAreas.includes(selectedZone)) {
+      return [];
+    } else if (mapOptions.lockPaintedAreas.length) {
+      const lockedAreas = mapOptions.lockPaintedAreas;
+      filterFunctions.push(
+        f =>
+          !lockedFeatures.has(f.id?.toString() || '') &&
+          !lockedAreas.includes(zoneAssignments.get(f.id?.toString() || '') || null)
+      );
+    } else if (lockedFeatures.size) {
```
mapmeld, line 527: maybe it's too long since I looked at this - I thought that lockedFeatures was a set of IDs that have a locked color (from mapOptions.lockPaintedAreas), is it connected to shattering or county paint?

mariogiampieri (dismissed):
this looks solid to me- i see a consistent locking experience across the fly app.

nofurtherinformation (approved):
lgtm

PR #231, opened 2024-12-31
Update PR DB machine memory

raphaellaude:
## Description

- Adjust default memory of postgres machine 4GB -> 1GB, [saving $15.01](https://fly.io/docs/about/pricing/) per PR app per month.
- TLDR it is (annoyingly) not possible to create a fly postgres cluster specifying the memory size on startup. Instead it is recommended to start with the closest preset (4GB for `shared-cpu-2x`) and then [adjust subsequently](https://fly.io/docs/postgres/managing/scaling/). It _is_ possible to [set the default machine memory in the `fly.toml` config](https://community.fly.io/t/unexpected-ram-increase-in-shared-cpu-1x-after-deployment/17134) but since we are forking from prod (incl. its configs) this would mean prod has to be set up with fewer resources.

## Reviewers

- Primary: @mariogiampieri 
- Secondary: @nofurtherinformation 

## Checklist

### DIFF
--- a/.github/workflows/fly-deploy-pr.yml
+++ b/.github/workflows/fly-deploy-pr.yml
@@ -53,7 +53,7 @@ jobs:
               id: fork-db
               if: github.event.action != 'closed'
               run: |              
-                DB_NAME="${{ github.event.repository.name }}-${{ github.event.number }}-api"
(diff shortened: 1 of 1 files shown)


nofurtherinformation (approved):
Lgtm!

PR #232, opened 2024-12-31
Better FE request management 

nofurtherinformation:
We get in some weird states where the async nature of the population and updates can get out of sync with the frontend. This shows as population bars jumping around when the server is under load. 

## Description
- Adds abort controllers to update and zone population queries
- Increase debounce on zone updates
- Track last zone updates with a hash to eliminate duplicate zone updates (eg. on mouse out, painting stops and a zone update is triggered even if no features have been updated)

## Reviewers
- Primary: @raphaellaude 
- Secondary:

## Checklist
- [ ] Abort controllers work under normal circumstances
- [ ] Abort controllers work under heavy load
- [ ] Population bars do not jump around substantially
- [ ] No work is lost under normal and heavy loads

### DIFF
--- a/app/src/app/store/mapStore.ts
+++ b/app/src/app/store/mapStore.ts
@@ -816,7 +816,7 @@ export const useMapStore = create(
         setSelectedZone: zone => set({selectedZone: zone}),
         zoneAssignments: new Map(),
         assignmentsHash: '',
-        lastUpdatedHash: '',
+        lastUpdatedHash: Date.now().toString(),
         setAssignmentsHash: hash => set({assignmentsHash: hash}),
         accumulatedGeoids: new Set<string>(),
         setAccumulatedGeoids: accumulatedGeoids => set({accumulatedGeoids}),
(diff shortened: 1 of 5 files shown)


app/src/app/utils/api/apiHandlers.ts
```diff
@@ -132,6 +133,11 @@ export interface ZonePopulation {
   total_pop: number;
 }
 
+
+export let populationAbortController: AbortController | null = null;
+export let updateAbortController: AbortController | null = null;
+export let currentHash: string = '';
```
raphaellaude, line 139: I: Approving to get this bug fix in but wanted to flag that tanstack query automatically can [provide a signal to each query function](https://tanstack.com/query/v4/docs/framework/react/guides/query-cancellation)! This might be a nicer pattern than maintaining a growing list of globals in `apiHandlers`

If using the tanstack built-in functionality seems like a good pattern, I'd recommend leaving a `// TODO: ...` for the next person here!

raphaellaude (dismissed):
lgtm, just one non-blocking comment about using tanstack built-in abort signals

app/src/app/store/mapStore.ts
```diff
@@ -812,6 +816,7 @@ export const useMapStore = create(
         setSelectedZone: zone => set({selectedZone: zone}),
         zoneAssignments: new Map(),
         assignmentsHash: '',
+        lastUpdatedHash: '',
```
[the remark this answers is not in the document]
nofurtherinformation: Done!

app/src/app/utils/api/apiHandlers.ts
nofurtherinformation: Added!

raphaellaude (approved):

PR #233, opened 2024-12-31
Partial document updates [Fixes #114]

### LINKED ISSUE #114: Assignment updates with partial document
Right now, we send the entire document to the backend after each paint event. This has a number of pros/cons

**Pros:**
- We're less likely to get out of sync with the backend (maybe, see below). This prevents us from having issues like the scenario where we update the FE map state but hit an error saving the assignments on the backend (this isn't the worst case though because a simple reload will set things straight).

**Cons:**
- Requires maintaining a replica of the assignments in three places: backend, mapStore's zoneAssignments, and in the map's feature state. This can get out of sync really easily–as we're seeing with the shattering work–and lead to erroneous state.
- Performance: so far this hasn't been a perf issue but you can see the latency get longer for large maps and could become unmanageable, especially as we support more concurrent users.

Note: the backend currently already does partial document updates so this requires only a FE implementation.

raphaellaude:
## Description

- Leverages existing API upsert behavior to only push incremental updates
- Seems to be a bit buggy still but the _huge_ drop in latency is a promising sign for this approach–just need to make sure we're not dropping any assignments!

TODO
- Assignment updates not properly occurring on mouse up so spamming the server way too much

## Reviewers

- Primary:
- Secondary:

## Checklist

### DIFF
--- a/.github/workflows/fly-deploy-pr.yml
+++ b/.github/workflows/fly-deploy-pr.yml
@@ -53,7 +53,7 @@ jobs:
               id: fork-db
               if: github.event.action != 'closed'
               run: |              
-                DB_NAME="${{ github.event.repository.name }}-${{ github.event.number }}-api"
+                DB_NAME="${{ github.event.repository.name }}-${{ github.event.number }}-db"
                 if flyctl postgres list | grep -q $DB_NAME; then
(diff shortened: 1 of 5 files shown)


raphaellaude:
Going to start fresh with some new ideas

PR #234, opened 2025-01-02
Save + share plans

mariogiampieri:
You should be able to start a new plan and then click the "Save/Share" (open to other language there) button to add a name or tags to the plan. This data is saved to the document.document_metadata table and is accessible if you return to the map later (via the `districtr-persistrictr` store) or if you share the map via link. 

## Description

- Save plan (metadata) to the database
  - [x] create flexible `DocumentMetadata` table that holds `key:value` pairs of metadata to allow for future extensibility
  - [x] start with plan name and "tag or event code" in keeping with districtr v1 
  - [x] return plan name and any tags when document is loaded via URL
  - [x] return plan name and any tags when document is loaded via recents table
  - [x] align this schema / behavior with what's shown in the recents table (should be same name; right now setting name in Save does not propagate to name in Recents)
 - Share plan via url
   - [x] create shareable link for sharing plan. _Partially complete:_ in theory the link could be shared and the plan could be updated now. I want to add a URL param that flags that the plan was shared, so that a copy of the plan with a fresh documet ID is created (and the user is informed of the copy being made) so as to not mess with the original plan. 
   - [x] Protect plan with a password or token. I did some research into this, and think we would be able to create a password, hash it, return a temporary token to share via url (for example). that way we aren't sending a password directly in the URL string. Have not implemented yet but seems viable. 

## Reviewers

- Secondary: @raphaellaude 

## Checklist

image


app/src/app/components/sidebar/ShareMapsModal.tsx
```diff
@@ -0,0 +1,187 @@
...
+
+  // if no gerrydb table selected or no edits made, return null
+  if (!gerryDBTable || !userMaps?.length) {
+    return null;
+  }
+
+  const handleMapSave = () => {
+    const metadataObjects = [
+      {key: 'name', value: name},
+      {key: 'tags', value: tagsTeam},
```
nofurtherinformation, line 100: H: Maybe add an explicit 'saved' property to help differentiate all recents vs saved ones `{key: 'saved', value: 'true'}` or similar

app/src/app/components/sidebar/RecentMapsModal.tsx
```diff
@@ -32,11 +32,12 @@ export const RecentMapsModal = () => {
   const setMapDocument = useMapStore(store => store.setMapDocument);
   const [dialogOpen, setDialogOpen] = React.useState(false);
```
nofurtherinformation, line 33: H: Add an option in recent maps modal to filter for saved maps, to help cull down the long list?

```typescript
const [showAllMaps, setShowAllMaps] = React.useState(false);
const filteredMaps = showAllMaps
  ? userMaps
  : userMaps.filter(f => f.metadata?.find(f => f.key === 'saved'));
    
...
<Text as="label" size="2">
  <Flex as="span" gap="2">
    <Checkbox
      checked={showAllMaps}
      onClick={() => setShowAllMaps(prev => !prev)}
    />
    "Show all recent maps"
  </Flex>
</Text>
...
{filteredMaps.map((userMap, i) => (
              <RecentMapsRow
```

nofurtherinformation (changes requested):
This looks really good. I think we should add a little bit to make it easier to users to see explicitly saved maps and cut down on the long list of recents, but otherwise looks good to me for an MVP on this feature!

backend/app/main.py
```diff
@@ -406,6 +442,30 @@ async def get_gerrydb_summary_stat(
             )
 
 
+@app.post("/api/document/metadata/{document_id}", status_code=status.HTTP_200_OK)
```
raphaellaude, line 445: I: I'd suggest we use a slightly more RESTful endpoint format, `/api/document/{document_id}/metadata`

backend/app/models.py
```diff
@@ -158,6 +160,30 @@ class DocumentPublic(BaseModel):
...
+
+    document_id: UUID4 = Field(
+        sa_column=Column(
+            UUIDType,
+            ForeignKey("Document.document_id"),
+            nullable=False,
+        )
+    )
+    key: str | None = Field(nullable=False)
+    value: str | None = Field(nullable=False)
```
raphaellaude, line 186: Q: Do we think these will always be key-value formatted? I could imagine that eventually some metadata comes along that would want to be represented differently (e.g document tags which would make sense as `{ tags: list[str] }`.

raphaellaude: Additionally, this is quite opaque as we have no guardrails for what is a valid key. I could imagine the FE abusing this KV store, leading to really challenging backwards compatibility issues.

backend/app/alembic/versions/80339c4d6442_create_update_metadata_udf.py
```diff
@@ -0,0 +1,68 @@
...
+branch_labels: Union[str, Sequence[str], None] = None
+depends_on: Union[str, Sequence[str], None] = None
+
+
+def upgrade() -> None:
+    # ### commands auto generated by Alembic - please adjust! ###
+    op.execute(
+        sa.text(
+            """
+            CREATE OR REPLACE FUNCTION document.update_metadata(doc_id TEXT, metadata JSONB)
```
raphaellaude, line 27: H: If we go this route, create a new `.sql` file in `backend/app/sql/` with this function.

backend/app/alembic/versions/0b0d9e1036de_add_document_metadata.py
```diff
@@ -0,0 +1,55 @@
...
+            nullable=False,
+        ),
+        sa.Column(
+            "updated_at",
+            sa.TIMESTAMP(timezone=True),
+            server_default=sa.text("CURRENT_TIMESTAMP"),
+            nullable=False,
+        ),
+        sa.Column("document_id", sa.String(), nullable=False),
+        sa.Column("key", sa.String(), nullable=False),
```
raphaellaude, line 40: H: String column for key has no constraints.

backend/app/alembic/versions/0b0d9e1036de_add_document_metadata.py
```diff
@@ -0,0 +1,55 @@
...
+        sa.Column("value", sa.String(), nullable=False),
```
raphaellaude, line 41: I: String column for value is not very flexible, requiring FE to always transform the value into a format it prefers.

app/src/app/store/mapStore.ts
```diff
@@ -378,11 +386,28 @@ export const useMapStore = create(
           useChartStore.getState().updateMetrics(popChanges);
           set({
             assignmentsHash: Date.now().toString(),
-          })
+          });
         },
         mapViews: {isPending: true},
         setMapViews: mapViews => set({mapViews}),
         mapDocument: null,
+        mapName: () => get().mapDocument?.metadata?.find(k => k.key === 'name')?.value || undefined,
```
raphaellaude, line 394: H: A challenge of having an untyped KV store is that for each possible config we need to search for the key.

backend/app/models.py
```diff
@@ -158,6 +160,30 @@ class DocumentPublic(BaseModel):
     updated_at: datetime
     extent: list[float] | None = None
     available_summary_stats: list[str] | None = None
+    metadata: List[Dict[str, Any]] = []
+
+
+class DocumentMetadata(TimeStampMixin, SQLModel, table=True):
```
raphaellaude, line 166: H: I think there are a few challenges having this as a separate entity-attribute / KV modeled table. I've noted these in some of the other comments.

I'd recommend we turn this into a pydantic model which enumerates all metadata settings. This will allow us to type the metadata and make certain fields required. The actual column would be a JSON field on the `DistrictrMap` table, which we can validate with the output model.

backend/app/main.py
```diff
@@ -270,14 +275,45 @@ async def get_document(document_id: str, session: Session = Depends(get_session)
...
             Document.gerrydb_table == DistrictrMap.gerrydb_table_name,
             isouter=True,
         )
+        .join(
+            DocumentMetadata,
+            cast(Document.document_id, UUIDType)
+            == cast(DocumentMetadata.document_id, UUIDType),
+            isouter=True,
+        )
+        .group_by(
```
raphaellaude, line 305: I: If we go the route of using the metadata KV model, I would prefer we compute a view with document metadata so we can do a 1-1 join rather than have the group by here.

backend/app/models.py
```diff
@@ -158,6 +160,30 @@ class DocumentPublic(BaseModel):
     updated_at: datetime
     extent: list[float] | None = None
     available_summary_stats: list[str] | None = None
+    metadata: List[Dict[str, Any]] = []
```
raphaellaude, line 163: See comment above but would type this as a pydantic model `DistrictrMapMetadata` – the actual table would have the type JSON

raphaellaude (dismissed):
I'd love to discuss the metadat data model. I think there are a number of significant disadvantages to using an entity-attribute / KV model for the metadata.

Also, we should add unit tests!

raphaellaude:
Also would love to discuss

> Protect plan with a password or token. I did some research into this, and think we would be able to create a password, hash it, return a temporary token to share via url (for example). that way we aren't sending a password directly in the URL string. Have not implemented yet but seems viable.

Is the idea that the map is in a view-only mode if it isn't unlocked? Or is it not shown at all? Might want to discuss w/ Moon.

app/src/app/components/sidebar/ShareMapsModal.tsx
```diff
@@ -0,0 +1,187 @@
...
+    // close dialog
+    setDialogOpen(false);
+  };
+
+  // get map name from metadata if it exists
+  const mapName = mapDocument?.metadata?.find(k => k.key === 'name')?.value || undefined;
+  const mapTags = mapDocument?.metadata?.find(k => k.key === 'tags')?.value || undefined;
+
+  const [name, setName] = React.useState(mapName);
+  const [tagsTeam, setTagsTeam] = React.useState(mapTags);
```
raphaellaude, line 66: Because these are persisted, the tags seem to show up across maps.

mariogiampieri:
> Also would love to discuss
> 
> > Protect plan with a password or token. I did some research into this, and think we would be able to create a password, hash it, return a temporary token to share via url (for example). that way we aren't sending a password directly in the URL string. Have not implemented yet but seems viable.
> 
> Is the idea that the map is in a view-only mode if it isn't unlocked? Or is it not shown at all? Might want to discuss w/ Moon.

you'd need a pw for the map to be visible; if it is visible, you would be able / required to make a copy to edit

app/src/app/components/Toolbar/PasswordPromptModal.tsx
```diff
@@ -0,0 +1,61 @@
...
+        <Box p="3">
+          <Text size="2" weight="bold">
+            This plan is password protected.{' '}
+          </Text>
+          <Text size="3">Please enter a valid password to access this plan</Text>
+
+          <TextField.Root
+            placeholder="Password"
+            size="3"
+            value={password}
```
nofurtherinformation, line 50:
```suggestion
            value={password ?? undefined}
```

Should fix the build!

raphaellaude (commented):
This is really shaping up! Nicely done. I've left a few comments / questions for now but haven't reviewed everything quite yet. Will return to the review tonight.

Only a couple blocking "H" comments. Happy to budge on the I's, PP's, O's, and Q's to get this checked in ASAP.

nofurtherinformation:
Two other semi-tangential things:

1. Some bugs may be arising from keeping the cache assignments in sync with the backend. Because the parent child edge index makes the get_assignments endpoint pretty trivially fast, I'm tempted to cut out the local caching for stability and simplicity. Eg, always hit the server for assignments on load
2. In `bakcend/Dockerfile.dev` pls swap the image to `FROM python:3.12` which has the gcc pre req for the new packages

cdonnay:
Hi @mariogiampieri ! I just spent some time playing with this PR and have a few bugs I found:

- when I share a map with a password, and then go to use the link, it never asks for the password and never loads the map.
- I shared a map as view only but it edited my original map
- when I load a shared map, I have to reload the page to get paint to render

I have a screen recording I will share on slack!

Questions:
- is the intended behavior when I save a map under a new name that it does not make a copy and rather just renames the original map?

backend/app/models.py
```diff
@@ -145,8 +154,107 @@ class Document(TimeStampMixin, SQLModel, table=True):
     gerrydb_table: str | None = Field(nullable=True)
 
 
+class DistrictrMapMetadata(BaseModel):
```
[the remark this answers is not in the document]
mariogiampieri: renamed

backend/app/main.py
```diff
@@ -460,6 +659,39 @@ async def get_gerrydb_summary_stat(
...
+        session.execute(stmt)
+        session.commit()
+
+    except Exception as e:
+        logger.error(e)
+        session.rollback()
+        raise HTTPException(
+            status_code=status.HTTP_500_INTERNAL_SERVER_ERROR,
+            detail="Metadata update failed",
+        )
```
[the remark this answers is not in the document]
mariogiampieri: updated

backend/app/main.py
```diff
@@ -460,6 +659,39 @@ async def get_gerrydb_summary_stat(
             )
 
 
+@app.post("/api/document/{document_id}/metadata", status_code=status.HTTP_200_OK)
```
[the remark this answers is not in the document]
mariogiampieri: updated

backend/app/models.py
```diff
@@ -145,8 +154,107 @@ class Document(TimeStampMixin, SQLModel, table=True):
...
+        sa_column=Column(Integer, primary_key=True, autoincrement=True)
+    )
+
+    document_id: str = Field(
+        sa_column=Column(
+            UUIDType,
+            ForeignKey("document.document_id"),
+        )
+    )
+    map_metadata: DistrictrMapMetadata = Field(
```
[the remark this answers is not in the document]
mariogiampieri: i'm open to this but think it'd be good to think through all of the potential items we could see being included here first - from our last checkin I'd wonder how things like custom bins for choropleth maps, locks/not locked districts, etc. are handled

backend/app/main.py
```diff
@@ -303,21 +400,123 @@ async def get_document(document_id: str, session: Session = Depends(get_session)
...
+@repeat_every(seconds=60)  # Run every minute
+def cleanup_expired_locks():
+    session = sessionmaker(bind=engine)()
+    try:
+        N_HOURS = 1  # arbitrary for now
+        expiry = datetime.now() - timedelta(hours=N_HOURS)
+
+        stmt = text(
+            "DELETE FROM document.map_document_user_session WHERE updated_at < :expiry"
+        )
```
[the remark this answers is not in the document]
mariogiampieri: good catch, updated

app/package.json
```diff
@@ -54,6 +55,7 @@
     "react-map-gl": "^7.1.3",
     "react-resizable": "^3.0.5",
     "superjson": "^2.2.1",
+    "uuid": "^11.0.5",
```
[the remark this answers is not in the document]
mariogiampieri: very cool, had never heard of nanoid before, swapped!

app/src/app/components/Topbar.tsx
```diff
@@ -165,9 +173,22 @@ export const Topbar: React.FC = () => {
...
+            >
               Share
             </Button>
+            <Button
+              variant="outline"
+              className="mr-2"
+              disabled={!mapDocument?.document_id}
+              onClick={() => setSaveMapsModal(true)}
+            >
+              {mapDocument?.genesis === 'shared' ? 'Make a Copy' : 'Save'}
```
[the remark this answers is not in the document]
mariogiampieri: I had been wondering the same- open to thoughts but have updated for now to "Save Metadata"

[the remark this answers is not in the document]
mariogiampieri: huh interesting- I'm definitely able to reproduce the error you mention below in the `mutations` comment, where upon creating a copy the bars show but not the map, but when i e.g. create a map -> create a view only link w/o password -> paste in different browser window i do see the greyed out edit tools and "make a copy" instead of "save". do you remember any other factors that i could use to reproduce?

app/src/app/components/Toolbar/PasswordPromptModal.tsx
```diff
@@ -0,0 +1,81 @@
...
+          <TextField.Root
+            placeholder="Password"
+            size="3"
+            type="password"
+            value={password ?? undefined}
+            onChange={e => handlePasswordEntry(e.target.value)}
+          ></TextField.Root>
+          <Flex gap="2" py="2">
+            <Button onClick={handlePasswordSubmit}>Submit</Button>
+            <Button onClick={proceedToStart}>Cancel and Proceed to Map</Button>
```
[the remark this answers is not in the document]
mariogiampieri: I went for a both-and approach for now but not a hill i'd die on

app/src/app/components/Toolbar/PasswordPromptModal.tsx
```diff
@@ -0,0 +1,81 @@
...
+          </Flex>
+          <Text>{shareMapMessage ?? ''}</Text>
```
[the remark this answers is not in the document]
mariogiampieri: good call!

mariogiampieri:
> Two other semi-tangential things:
> 
> 1. Some bugs may be arising from keeping the cache assignments in sync with the backend. Because the parent child edge index makes the get_assignments endpoint pretty trivially fast, I'm tempted to cut out the local caching for stability and simplicity. Eg, always hit the server for assignments on load
> 2. In `bakcend/Dockerfile.dev` pls swap the image to `FROM python:3.12` which has the gcc pre req for the new packages

thanks for the note on 2)! updated. wondering for 1) if that involves essentially undoing the work you had previously done to set that up and additionally whether any of the work to set up the local copy of the table that you were describing this week is in conflict with that undoing in any way. i agree that removing that bit of complexity would simplify things nicely. that's definitely not the only culprit for some of the bugs so i'll keep revising in the meantime but lmk what you think!

nofurtherinformation:
> it looks like some of that streamlining was already done for the removal of `idb` so i'll start by updating from main and otherwise revising in the meantime but lmk what you think!

Ahh, sorry about the lack of clarity here -- I should have said a bit more at the meeting. Raph and I removed the `idb` integration which was the complexity / cause of some issues here. This is separate from the clientside demography table  -- you should be good with your most recent merge!

app/src/app/components/Topbar.tsx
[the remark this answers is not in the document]
nofurtherinformation: Hey yes for sure, here's a more detailed run down. 

Browser A
1. Make a map
2. Share (view only)

Browser B
3. Open share link (cannot draw / interact)
4. Refresh page with document id query parameter (not JWT) `?document_id=...`
5. Draw on map

Browser A
1. Refresh map
2. Has Browser B edits


I made a quick screen cap to try to reflect it -- if easier happy to pair up / hop on a call! 

https://github.com/user-attachments/assets/65b80629-4266-4d26-8b7e-2d734c00ebd6

l!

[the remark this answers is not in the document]
mariogiampieri: woohoo thank you! i was able to fix- had been having an issue identifying some conflicting db records that was allowing cases like this to slip through. should be addressed now!

app/src/app/utils/api/queries.ts
```diff
@@ -137,12 +133,22 @@ updateDocumentFromId.subscribe(mapDocument => {
     url.searchParams.delete('document_id');
     window.history.replaceState({}, document.title, url.toString());
   }
+  // TODO- this should really be a warning and not an error
+  if (mapDocument.data?.status === 'locked') {
```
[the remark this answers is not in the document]
mariogiampieri: and based on the above error identification -> fix, this should now work properly as well

raphaellaude (commented):
Started on review but will come back to it esp. after testing!

app/src/app/components/Toolbar/ToolUtils.tsx
```diff
@@ -34,38 +34,45 @@ export const useActiveTools = () => {
...
+      icon: HandIcon,
+      hotKeyAccessor: e => {
+        return e.code === 'KeyM';
+      },
     },
     {
       hotKeyLabel: 'P',
       mode: 'brush',
-      disabled: !mapDocument?.document_id,
+      disabled:
```
nofurtherinformation, line 54: H: This is interesting -- the hotkey listeners still work and allow actions when the map is locked. I think we could just hide the toolbar entirely when `mapDocument?.status === 'locked' ||
mapDocument?.genesis === 'shared'`?

nofurtherinformation: Users would only need to pan, never the other tools?

app/src/app/components/Toolbar/ShareMapsModal.tsx
```diff
@@ -0,0 +1,168 @@
...
+  const handleCreateShareLink = async () => {
+    const payload = {
+      document_id: mapDocument?.document_id,
+      password: password ?? null,
+      access_type: sharetype,
+    };
+
+    try {
+      // get the share link
+      sharePlan.mutate(payload).then(token => {
```
nofurtherinformation, line 55: PP: Slight preference for await syntax here

nofurtherinformation: I: When a user sets a password and generates a share link, then deletes the password from the modal and creates another share link, it remains password protected

app/src/app/components/Toolbar/ShareMapsModal.tsx
```diff
@@ -0,0 +1,168 @@
...
+        // copy to clipboard
+        const shareableLink = `${window.location.origin}?share=${token.token}`;
+        navigator.clipboard.writeText(shareableLink);
+
+        // Set link copied state
+        setLinkCopied(true);
+        setTimeout(() => setLinkCopied(false), 2000);
+      });
+    } catch (error) {
+      console.error('Error creating share link: ', error);
```
nofurtherinformation, line 65: I: `mapStore.setErrorNotification` potentially to inform users!

app/src/app/components/Toolbar/PasswordPromptModal.tsx
```diff
@@ -0,0 +1,90 @@
...
+    } else {
+      setPassword(null);
+    }
+  };
+
+  const proceedToStart = () => {
+    setPasswordPrompt(false);
+    setAppLoadingState('loaded');
+
+    const documentUrl = new URL(window.location.toString());
```
nofurtherinformation, line 50: H: The behavior with being able to edit after reload could be sidestepped by just leaving the token in the URL?

nofurtherinformation (commented):
A couple suggestions / Q's here!

app/src/app/utils/api/mutations.ts
```diff
@@ -124,3 +125,88 @@ export const document = new MutationObserver(queryClient, {
...
+        document_id: mapDocument?.document_id || '',
+        token: data.token,
+        status: data.access === 'edit' ? 'edit' : 'locked',
+      },
+    });
+    return data.token;
+  },
+});
+
+export const sharedDocument = new MutationObserver(queryClient, {
```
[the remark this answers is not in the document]
mariogiampieri: i *think* this is resolved now based on tests last night- in my tests now i do not see the duplicate queries

app/src/app/components/Toolbar/ShareMapsModal.tsx
mariogiampieri: good call!

app/src/app/components/Toolbar/ToolUtils.tsx
mariogiampieri: great catch! i think that's right and there is minimal (if any) additional value for being able to see the tools that are disabled anyway.

app/src/app/components/Toolbar/PasswordPromptModal.tsx
mariogiampieri: this does seem like a good strategy, but this should now be fixed via a better check on the backend- if you share a link with a password, open in a new browser, and refresh, you should now get the proper warning and also have the "Make a Copy" button

app/src/app/components/Toolbar/ShareMapsModal.tsx
mariogiampieri: good catch- for now I made it so once a pw is saved (or entered, without being saved) it cannot be updated. perhaps annoying but for now bypasses the overhead / logic of updating

app/src/app/components/Toolbar/PasswordPromptModal.tsx
nofurtherinformation: Nice!

app/src/app/components/Toolbar/ShareMapsModal.tsx
```diff
@@ -0,0 +1,234 @@
...
+          {/* share logic */}
+          {mapDocument?.genesis !== 'shared' ? (
+            <>
+              <Text size="2">
+                You can share your plan with others! Share as <b>View Only</b> to let others see and
+                copy your plan. <b>Share and make editable</b> to allow others to directly edit your
+                plan. You can optionally <b>set a password</b> to restrict who can interact with it.
+              </Text>
+              <Flex gap="2" className="flex-row items-center">
+                {!password && passwordDisabled ? null : (
```
nofurtherinformation, line 161: I: Two quick things -- this needs a little width adjustment and after you add a password it no longer displays:

Screenshot 2025-03-05 at 7 55 28 AM
Screenshot 2025-03-05 at 7 55 19 AM

app/src/app/utils/api/apiHandlers.ts
```diff
@@ -93,6 +93,18 @@ export interface DocumentObject {
   updated_at: string | null;
   extent: [number, number, number, number]; // [minx, miny, maxx, maxy]
   available_summary_stats: string[];
+  map_metadata: DocumentMetadata;
+  status: string;
+  genesis: string;
```
nofurtherinformation, line 98:
```suggestion
  status: 'locked' | 'unlocked' | 'checked_out';
  genesis: 'shared' | 'copied' | 'created';
```

nofurtherinformation (commented):
This looks solid and is working well! I flagged one display bug but otherwise I think this looks good.

app/src/app/components/Toolbar/ShareMapsModal.tsx
mariogiampieri: thanks and good catch on the width there! should be cleaner now. re: the bar disappearing, unfortunately that's on purpose- right now the logic is that *if* you set a password on initial link creation, it will remain and be visible at times, however if you create without a pw, the option to enter one disappears. this isn't the most satisfying UI but wanted to get the reliable path of lesser resistance working well first.

nofurtherinformation: Gotcha -- that makes sense!

backend/app/alembic/versions/2dbfabc0a613_add_user_session_model.py
```diff
@@ -0,0 +1,49 @@
+"""add user session model
```
raphaellaude, line 1: I: Before merging I'd recommend we merge all the revisions into a single big revision.

raphaellaude (commented):
Just one comment but otherwise this is looking good. Will UAT a bit more later today. Looking great!

raphaellaude (dismissed):
Looks great. What an effort! Great job. Still a few small things left to do / some UI patterns that are a bit confusing but let's merge and fix those in follow-ups

nofurtherinformation (approved):
This is looking and working great! Thank you for all your work on this!

PR #237, opened 2025-01-03
[bugfix] Stale loaded map ID on new document

nofurtherinformation:
We are now using a slightly redundant 'loadedMapId' piece of state to avoid re-running certain subscriptions and functions, especially when tabbing in and out.

A small bug occurs when you create a new map, then return to the immediately previous map via Recents. This PR makes the new document create mutation change the loadedMapId in the state, which resolves the issue


raphaellaude (approved):
lgtm

PR #238, opened 2025-01-03
[TEST] Minimal partial assignment

nofurtherinformation:
Follow up to #233


raphaellaude (approved):
lgtm – works like a charm

PR #243, opened 2025-01-11
Update fallback for null num_districts [fixes #205]

### LINKED ISSUE #205: No swatches show up if `num_districts` is null
The FE should have a sensible default

mapmeld:
I noticed the same bug as #205 some time ago. If I remember right, it was when we added `num_districts` to the DB, so older maps had `null ?? 0` districts. This updates the null count fallback to be 4 districts.


nofurtherinformation (approved):
lgtm

PR #251, opened 2025-01-18
Population bars while painting bug fixes (part 1)

nofurtherinformation:
There are a few bugs with population while painting that make it feel unreliable.

## Description
- Changes break/healing to be a non-temporal action so that population properly incorporates
- Adds a throttle to map click events to prevent double paint events at the block level

## Reviewers
- Primary: @mapmeld 
- Secondary: @raphaellaude 

## Checklist
- [ ] Population bars are stable when breaking/healing
- [ ] Population bars are stable when painting at the block level


raphaellaude (dismissed):
lgtm

raphaellaude (approved):
lgtm

PR #254, opened 2025-01-20
CSV and GeoJSON assignment exports

raphaellaude:
## Description

- Adds CSV and GeoJSON assignment exports

#### Discussion

There are probably lots of ways to do this, which will all have different advantages and disadvantages! Let's discuss. For now, I've chosen to implement this using postgres's `COPY` (via pyscopg2) because it supports [block-by-block stream reading/writing](https://www.psycopg.org/psycopg3/docs/basic/copy.html#copying-block-by-block), limiting memory overhead in the API and theoretically improving perf:

```python
with conn.cursor().copy(sql, params=params) as copy:
        _out_file = f"/tmp/{out_file_name}"
        with open(_out_file, "wb") as f:
            while data := copy.read():
                f.write(data)
```

An obvious downside of this approach is that it is much less extensible to other export types. One promising approach might be to use `pg_duckdb` in the future. However for the moment DuckDB spatial is incompatible with PostGIS. Still not impossible as we could do some shenanigans like `ST_AsText` or `ST_AsWKB` but I kinda wanted to avoid shenanigans for now.

I could imagine using geopandas, too. I chose _not_ to do so for now because of the memory implications of reading/copying large tables. Doing this for other export types is certainly not incompatible with the current implementation, we would just use a strategy pattern for the table-writing step.

Please discuss in the comments!

#### To-do

- [x] Support CSV exports of assignments
- [x] Support GeoJSON exports of assignments
- [x] Support block-only CSV exports of assignments
- [ ] ~Support block-only GeoJSON exports of assignments~ Not needed for now
- [x] Support district boundaries GeoJSON exports
- [x] Implement FE UI (thank you to @nofurtherinformation for the assist!)

## Reviewers

- Primary: @nofurtherinformation @mariogiampieri 
- Secondary: @mapmeld @bailliekova 

## Checklist


mapmeld:
I have used postgres's `\COPY` before and believe it's the right way to get CSV out (though I don't know background about what duckdb might do here)
Is the psycopg that you're using here pyscopg (v3) or pyscopg2? The line `conn.cursor().copy(sql, params=params)` looks like the new https://www.psycopg.org/psycopg3/docs/basic/copy.html but I can't get it working at the moment

For import, today I was thinking about a minimal page with an `<input type="file"/>` and a drag-and-drop UI. The frontend can use a JS library to read the CSV, create a blank map, and set its assignments in batches with the existing API. This avoids the issues of uploading the file, running a Python script async, checking for when it's finished, and reporting any errors.

raphaellaude:
> Is the psycopg that you're using here pyscopg (v3) or pyscopg2?

Yep, running `psycopg==3.1.18`!

> For import, today I was thinking about a minimal page with an `<input type="file"/>` and a drag-and-drop UI. The frontend can use a JS library to read the CSV, create a blank map, and set its assignments in batches with the existing API. This avoids the issues of uploading the file, running a Python script async, checking for when it's finished, and reporting any errors.

I think that would work well for precinct data but might struggle to heal blocks into precincts if a full-block dataset is uploaded. But perhaps it would be handled well by the FE healing logic, not sure.

raphaellaude:
> it exports NULL precincts

@mapmeld I was wondering what to do about that. You think that we should remove nulls? Include all unassigned?

mapmeld:
My opinion would be filter out the NULL precincts, no need to include unassigned precincts

nofurtherinformation:
Late to the convo here but I think this is a reasonable approach. I think it's worth clarifying what geospatial outputs are actually important here -- beyond tabular results, Dave's only outputs the overall district shapes, which may be lighter for computationally and data distribution.

raphaellaude:
> Dave's only outputs the overall district shapes

Quickly added those!

backend/app/sql/get_block_assignments.sql
```diff
@@ -0,0 +1,34 @@
...
+    FROM document.document
+    LEFT JOIN districtrmap
+    ON document.gerrydb_table = districtrmap.gerrydb_table_name
+    WHERE document.document_id = $1;
+
+    IF doc_districtrmap.gerrydb_table_name IS NULL THEN
+        RAISE EXCEPTION 'Table name not found for document_id: %', $1;
+    END IF;
+
+    IF doc_districtrmap.child_layer IS NULL THEN
```
nofurtherinformation, line 17: H: This works but may be confusing to users that block level downloads are only available for some plans. Will the default configuration in the full version be that everything is shatterable unless some exception?

backend/app/sql/get_block_assignments.sql
```diff
@@ -0,0 +1,34 @@
...
+        RAISE EXCEPTION 'Child layer is NULL for document_id: %. Block-level exports are not supported', $1;
+    ELSE
+        sql_query := format('
+            SELECT
```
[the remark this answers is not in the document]
raphaellaude: maybe – let's look at our postgres config and the number of processes first – these operations aren't _that_ heavy

backend/app/sql/get_block_assignments.sql
raphaellaude: I think in practice yeah–we'll almost only have shatterable maps

nofurtherinformation (approved):
Lgtm!

PR #260, opened 2025-01-23
Create document from block assignments CSV

mapmeld:
As discussed in #254 , this PR adds a page on `/uploader` where a user can select a CSV file, or drop a CSV file into the browser window, and create a map based on an existing gerrydb table. Since it is currently only frontend changes, it works as a separate PR

Blank or non-numeric zones are ignored. Big CSVs are uploaded at 2,000 rows/batch

Potential issues
- supporting and validating more CSV types (TSV, extra headers)
- possible to check on server or client that the first geoids have matches in the selected gerrydb table?
- moving finished map links to a sidebar, allowing multiple CSV uploads
- future auth integration


app/package.json
```diff
@@ -43,6 +43,7 @@
     "lodash": "^4.17.21",
     "maplibre-gl": "^4.4.1",
     "next": "^14.2.7",
+    "papaparse": "^5.5.1",
```
nofurtherinformation, line 46: H: The build is failing because the types for papaparse come from a separate package -- running `npm i -d @types/papaparse` or adding `@types/papaparse` to dev dependencies should resolve this

mapmeld: 👍

mapmeld:
OK, I did some testing and think this is ready for review.
- Allows user to upload multiple files (one at a time)
- Links appear on the right column, with the original CSV's filename
- Papaparse handles TSV properly
- Unusual rows (where right column is blank or not a number value) are skipped

Screenshot 2025-02-02 at 9 40 59 PM

app/src/app/uploader/page.tsx
```diff
@@ -0,0 +1,148 @@
...
+            rows.slice(rowCursor, rowCursor + ROWS_PER_BATCH).forEach(row => {
+              if (row.length == 2 && row[1] !== '' && !isNaN(Number(row[1]))) {
+                assignments.push({document_id, geo_id: row[0], zone: Number(row[1])});
+              }
+            });
+            patchUpdateAssignments(assignments).then(stepResult => {
+              setProgress(rowCursor + assignments.length);
+              rowCursor += ROWS_PER_BATCH;
+              if (rowCursor > results.data.length) {
+                setMapLinks([...mapLinks, {document_id, name: file.name}]);
```
nofurtherinformation, line 56: PP: On complete, could be nice to add the upload to user's recent maps, something like:

```ts
const upsertUserMap = useMapStore(state => state.upsertUserMap)
...
 upsertUserMap({
   mapDocument: response,
 })
```

app/src/app/uploader/page.tsx
```diff
@@ -0,0 +1,148 @@
...
+
+  const handleFileSelected = (event: React.ChangeEvent<HTMLInputElement>) => {
+    const file = event.target.files?.[0];
+    if (file) {
+      processFile(file);
+    }
+  };
+
+  return (
+    <div className="h-screen w-screen flex items-center justify-center bg-gray-100">
```
nofurtherinformation, line 83: Medium (?): Use Radix UI Themes component (`<Flex>`, `<Heading>`, `<Box>` etc) when possible to ensure styling consistency

app/src/app/uploader/page.tsx
```diff
@@ -0,0 +1,148 @@
...
+import {GerryDBViewSelector} from '../components/sidebar/GerryDBViewSelector';
+import {useMapStore} from '@/app/store/mapStore';
+import {Assignment, createMapDocument, patchUpdateAssignments} from '@/app/utils/api/apiHandlers';
+
+type MapLink = {
+  document_id: string;
+  name: string;
+};
+
+export default function Uploader() {
```
nofurtherinformation, line 14: Q: Should this function be directly available in the dropdown menu on the map page?

nofurtherinformation (changes requested):
Overall looking really and working nicely! One moderate refactor and a couple non-blocking comments, but then I think this is good to merge.

raphaellaude (changes requested):
Let's discuss this.

The approach of uploading blocks as assignments and letting the FE un-break the entire document in serial poses a number of challenges:

### 1. Very high DB load

image

This db load is from a single block map being loaded for Kansas (an export from Districtr itself).

### 2. Slow user experience

image

I think that writing an import endpoint would be _much_ more performant and support many more current loads.

nofurtherinformation:
@raphaellaude Agreed on all points for blocks level imports, which are the core use case. 

After a little more thought, we may want to wait until we have more full functionality to merge -- supporting only VTD imports may be confusing.

raphaellaude:
Oh and we should definitely link `/uploader` to the map page, right? Maybe under a new menu group in select map

image

raphaellaude:
> It looks like it's possible to use psycopg to COPY from a CSV string ( with cur.copy("COPY tmp_data(id, bla) FROM stdin (format csv, delimiter '|', quote '\"')")  in https://github.com/psycopg/psycopg/discussions/139 ).
The frontend here could be validating the file and then passing it to the server in the standardized format.

Yeah that sounds good for loading the data. Once it's in postgres, I think something like the following pseudo-query should then allow us to unbreak the the blocks:

```sql
with edge_assignments as (
  select
    doc.document_id,
    edges.parent_path,
    edges.child_path,
    doc.zone
  from parentchildedges edges
  left join document doc
  on doc.geoid = edges.child_path ),
unbroken_parents as (
  select
    e.parent_path as path,
    e.zone
  from edge_assignments e
  left join (
    select edges.parent_path, count(*) parent_count
    from edge_assignments ) pc
  using(parent_path)
  group by e.parent_path, e.zone
  having e.count(*) = parent_count )
select
  '{document id}' as document_id,
  unbroken_parents.path,
  unbroken_parents.zone
from
  unbroken_parents
union all
select
  doc.document_id,
  edges.parent_path as path,
  edges.zone
from
  edge_assignments edges
where edges.parent_path is not in (
  select parent_path
  from unbroken_parents )
```

backend/app/main.py
```diff
@@ -268,6 +270,31 @@ async def reset_map(document_id: str, session: Session = Depends(get_session)):
...
+):
+    d = data.model_dump()
+    csv_file = io.StringIO(d["assignmentTXT"])
+
+    session.execute(text("DROP TABLE IF EXISTS temploader"))
+    session.execute(text("""CREATE TABLE temploader (
+        geo_id TEXT,
+        zone INT
+    )"""))
+    session.connection().connection.cursor().copy_expert(
```
nofurtherinformation, line 285: H: This method does not appear to exist on `cursor`
```
2025-02-10T16:42:02.894 app[48e236dc3e9e58] ewr [info] File "/app/app/main.py", line 285, in upload_assignments

2025-02-10T16:42:02.894 app[48e236dc3e9e58] ewr [info] session.connection().connection.cursor().copy_expert(

2025-02-10T16:42:02.894 app[48e236dc3e9e58] ewr [info] ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

2025-02-10T16:42:02.894 app[48e236dc3e9e58] ewr [info] AttributeError: 'Cursor' object has no attribute 'copy_expert'
```

mapmeld: This is maybe a psycopg 2 / 3 issue, as I did end up with a new table?  I will try to get a blocks import -> shattered map example working, use a TEMP TABLE instead of this current system, and then figure out what function belongs here

nofurtherinformation: Ah, yes I think you're right. My local build (via docker-compose) and the dev preview should be on psycopg3

mapmeld: Does this latest commit work? I think it's the right way for psycopg3, it looks like it isn't bulk upload, but it's the same concept of a single COPY command 
References: https://www.psycopg.org/psycopg3/docs/basic/copy.html
https://www.psycopg.org/articles/2020/11/15/psycopg3-copy/

backend/app/main.py
```diff
@@ -277,20 +277,52 @@ async def upload_assignments(
...
+    # insert into actual assignments table
     session.execute(text("""
         INSERT INTO document.assignments (geo_id, zone, document_id)
         SELECT geo_id, zone, :document_id
         FROM temploader
     """), {"document_id":  document_id})
+
+    # find items to unshatter
+    results = session.execute(text("""
+        SELECT DISTINCT(SUBSTR(geo_id, 1, 11)) AS vtd, zone
```
nofurtherinformation, line 303: H: Unfortunately, VTD GEOIDs are not perfectly hierarchical with blocks. They diverge at the county level, so we can't rely on slicing the IDs to identify VTDs to heal/unshatter.

image

mapmeld: :+1: that makes sense. I also realized that I was doing this in reverse (un-shattering blocks, when on a fresh map I need to be shattering the VTDs which need block-level resolution)

backend/app/main.py
nofurtherinformation: This does work to populate the table with block assignments -- both locally and on the preview deploy!

mapmeld:
Alright, latest change does a few extra queries, maybe something could be consolidated here, but it does successfully import a mix of VTDs and blocks from a block-level CSV 
Screenshot 2025-02-11 at 7 57 15 PM

backend/app/main.py
```diff
@@ -268,6 +269,91 @@ async def reset_map(document_id: str, session: Session = Depends(get_session)):
...
+    # insert into actual assignments table
+    session.execute(text("""
+        INSERT INTO document.assignments (geo_id, zone, document_id)
+        SELECT geo_id, zone, :document_id
+        FROM temploader
+    """), {"document_id":  document_id})
+
+    session.commit()
+
+    return {"assignments_upserted": 1}
```
raphaellaude, line 354: H: I'd think we want a more informative return value. It's a bit misleading to say only a single assignment was inserted (also we're not upserting).

backend/app/main.py
```diff
@@ -268,6 +269,91 @@ async def reset_map(document_id: str, session: Session = Depends(get_session)):
...
+
+    # re-import VTDs that are fully uniform zone
+    vtds = []
+    with cursor.copy("COPY temploader (geo_id, zone) FROM STDIN") as copy:
+        for row in results:
+            vtd, zone = row
+            vtds.append(vtd)
+            copy.write_row([vtd, zone])
+    logger.info("uniform parents")
+    logger.info(vtds)
```
raphaellaude, line 321: I: Once we've loaded the assignments in via the temp table, I don't love performing another copy. Did you try adapting the query I previously suggested which performs all steps in a single query?

backend/app/main.py
```diff
@@ -268,6 +269,91 @@ async def reset_map(document_id: str, session: Session = Depends(get_session)):
     return {"message": "Assignments partition reset", "document_id": document_id}
 
 
+@app.patch("/api/upload_assignments")
+async def upload_assignments(
```
raphaellaude, line 273: H: Please add tests

backend/app/main.py
```diff
@@ -268,6 +269,91 @@ async def reset_map(document_id: str, session: Session = Depends(get_session)):
...
+        geo_id TEXT,
+        zone INT
+    )"""))
+    cursor = session.connection().connection.cursor()
+    with cursor.copy("COPY temploader (geo_id, zone) FROM STDIN") as copy:
+        for record in csv_rows:
+            if record[1] == "":
+                copy.write_row([record[0], None])
+            else:
+                copy.write_row([record[0], int(record[1])])
```
raphaellaude, line 292: H: add handling for bad inputs / types.

backend/app/main.py
```diff
@@ -268,6 +269,91 @@ async def reset_map(document_id: str, session: Session = Depends(get_session)):
...
+    """))
+    shatter = []
+    for row in results:
+        shatter.append(row[0])
+    logger.info("to shatter")
+    logger.info(shatter)
+
+    # insert into actual assignments table
+    session.execute(text("""
+        INSERT INTO document.assignments (geo_id, zone, document_id)
```
raphaellaude, line 347: H: Loading the rest of the assignments this way assumes that all the geo_ids are valid. I think at some stage we need to join the blocks against the parent_child_edges in order to determine that they are valid.

raphaellaude: PP: When bulk loading, specifying the partition in the table name `INSERT INTO document.assignments_{document_id} ...` will have a slight performance boost over inserting w/o the partition specified.

raphaellaude (changes requested):
Looking good! I think we need a few extra pieces before this is ready for prod:

- Block geoids are not validated. All the residual blocks in temp table–after parented blocks are removed/replaced by VTDs–are assumed to be valid. We should not assume that is the case.
- Input assumes the CSV has at least two columns which represent the geoids and zone. Invalid inputs are not handled by the endpoint since the data object is loosely typed as `list[list[str]]`.
- Backend tests are needed for the new endpoint, including different failure states for bad inputs. The backend should be a better friend to the client, providing informative error messages rather than 500s.
- Return value is misleading.

Otherwise, I still have a preference for the single query insert but this approach will work.

app/src/app/uploader/page.tsx
nofurtherinformation: Agreed!

app/src/app/uploader/page.tsx
```diff
@@ -0,0 +1,154 @@
...
+  const upsertUserMap = useMapStore(state => state.upsertUserMap);
+
+  const handleDragOver = (event: React.DragEvent) => {
+    event.preventDefault();
+  };
+
+  const processFile = (file: File) => {
+    if (!file) {
+      throw new Error('No file selected');
+      return;
```
nofurtherinformation, line 31: H: At least some file type validation would be good -- I accidentally dropped in a parquet and only got an internal server error response.

Additionally, some simple field / geoid validation would be good -- eg. if more than two columns should we ask users? Checking if GEOIDs all start with the expected state FIPS code?

app/src/app/uploader/page.tsx
```diff
@@ -0,0 +1,154 @@
...
+    }
+
+    Papa.parse(file, {
+      complete: results => {
+        setProgress(0);
+        setTotalRows(results.data.length);
+
+        createMapDocument({
```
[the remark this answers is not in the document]
nofurtherinformation: +1

backend/app/main.py
nofurtherinformation: H: About geoid validation, should we drop invalid/old geoids? What if someone uploads 2010 geoids?

backend/app/main.py
nofurtherinformation: +1

app/src/app/uploader/page.tsx
```diff
@@ -0,0 +1,154 @@
...
+import {useMapStore} from '@/app/store/mapStore';
+import {Assignment, createMapDocument, uploadAssignments} from '@/app/utils/api/apiHandlers';
+
+type MapLink = {
+  document_id: string;
+  name: string;
+};
+
+export default function Uploader() {
+  const [progress, setProgress] = useState<number>(0);
```
nofurtherinformation, line 15: Q: The progress bar doesn't seem to work in the current implementation, although the response time to process the query isn't very long. Do we need a progress bar?

nofurtherinformation (changes requested):
Thank you for all your work! This is working pretty nicely now! +1 to @raphaellaude suggestions and I added a few more, but this is moving in a great direction!

nofurtherinformation:
P.s. For code style I think this also needs the pre-commit hook installed/run!

app/src/app/uploader/page.tsx
mapmeld: I'll remove progress bar

backend/app/main.py
mapmeld: in the new flow this will return the document_id

backend/app/main.py
```diff
@@ -279,6 +280,130 @@ async def reset_map(document_id: str, session: Session = Depends(get_session)):
...
+    )
+    document_id = results.one()[0]
+
+    # process assignments CSV via temp table
+    session.execute(text("DROP TABLE IF EXISTS temploader"))
+    session.execute(
+        text("""CREATE TEMP TABLE temploader (
+        geo_id TEXT,
+        zone INT
+    )""")
```
nofurtherinformation, line 304: @bailliekova @raphaellaude If multiple requests hit this simultaneously will this create a race condition in the api or db where the first query's `temploader` table gets dropped?

bailliekova: I think it will; easy fix would be to append a random string to the temploader table name per load

backend/app/main.py
```diff
@@ -493,6 +494,254 @@ async def update_colors(
...
+                    SELECT COUNT(*) FROM {parent_child_table}
+                    WHERE parent_path = ANY(:geo_ids) OR child_path = ANY(:geo_ids)
+                    """),
+                    {"geo_ids": sample_geo_ids},
+                )
+                match_count = results.scalar()
+
+                if match_count == 0:
+                    # Clean up before raising exception
+                    session.execute(text(f"DROP TABLE IF EXISTS {temp_table_name}"))
```
raphaellaude, line 644: Temp tables are cleaned up at the end of the session or on commit anyway

backend/app/main.py
```diff
@@ -493,6 +494,254 @@ async def update_colors(
...
+        logger.error(f"Unexpected error in upload_assignments: {str(e)}")
+        # For the outermost exception handler, we don't need to specifically clean up
+        # the temp table as it's handled in the inner exception blocks
+        if not isinstance(e, HTTPException):
+            session.rollback()
+            raise HTTPException(
+                status_code=status.HTTP_500_INTERNAL_SERVER_ERROR,
+                detail=f"An unexpected error occurred: {str(e)}",
+            )
+        raise
```
raphaellaude, line 743: FastAPI will handle 500s for us so this is unnecessary

backend/app/main.py
```diff
@@ -493,6 +494,254 @@ async def update_colors(
...
+                    session.execute(
+                        text(f"""
+                        DELETE FROM {temp_table_name}
+                        WHERE geo_id = ANY(
+                            SELECT child_path FROM {parent_child_table}
+                            WHERE parent_path = ANY(:vtds)
+                        )
+                        """),
+                        {"vtds": vtds},
+                    )
```
raphaellaude, line 689: All of this can be done in one query replacing the temp table. No need to go back and forth with the API

backend/app/main.py
```diff
@@ -493,6 +494,254 @@ async def update_colors(
...
+
+            session.commit()
+
+            # Count how many assignments were actually created
+            count = session.execute(
+                text(
+                    "SELECT COUNT(*) FROM document.assignments WHERE document_id = :document_id"
+                ),
+                {"document_id": document_id},
+            ).scalar()
```
raphaellaude, line 715: Use `RETURNING` from insert query.

raphaellaude (changes requested):
Thanks for taking a pass at hardening. I think this looks more solid but leans pretty heavily on throwing exceptions. Errors like `Missing geo_id at row {i+1}` are going to be hard for users to debug–especially one at a time. Throwing 500s either doesn't help us much, too, as FastAPI will do so anyway if we encounter an error. We should try to define exceptions out of existence if we can. We'll probably need fuzzy heuristics for validation like: what share of blocks should be matched for it to be considered a valid upload? Other things like bulk reporting validation errors will make it easier for users to fix their data.

If it's alright with you I'll still pick this up and round it out.

backend/app/main.py
raphaellaude: Maybe not _one_ query but all in the db with another temp table yes

backend/app/main.py
raphaellaude: This is true but I added a `finally` to our session factory to ensure all sessions are closed after they drop out of scope. This ensures all temp tables are cleaned up

backend/app/main.py
```diff
@@ -268,6 +269,91 @@ async def reset_map(document_id: str, session: Session = Depends(get_session)):
     return {"message": "Assignments partition reset", "document_id": document_id}
 
 
+@app.patch("/api/upload_assignments")
```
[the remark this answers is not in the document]
raphaellaude: Ended up moving this logic into `/create_document`

backend/app/main.py
raphaellaude: Moved this so it's all db side

backend/app/main.py
raphaellaude: Actually... changed my mind completely and we should just be as permissive as possible and tell the user how many blocks we managed to match.

backend/app/main.py
raphaellaude: Added geoid validation against the parent child edges

backend/app/main.py
raphaellaude: If I'm reading the [docs](https://www.postgresql.org/docs/8.2/sql-createtable.html) right, we should be okay here as long as we clean up the session.

> If specified, the table is created as a temporary table. __Temporary tables are automatically dropped at the end of a session__, or optionally at the end of the current transaction (see `ON COMMIT` below). Existing permanent tables with the same name are not visible to the current session while the temporary table exists, unless they are referenced with schema-qualified names. Any indexes created on a temporary table are automatically temporary as well.

and

> Although the syntax of `CREATE TEMPORARY TABLE` resembles that of the SQL standard, the effect is not the same. In the standard, temporary tables are defined just once and automatically exist (starting with empty contents) in every session that needs them. PostgreSQL instead requires each session to issue its own `CREATE TEMPORARY TABLE` command for each temporary table to be used. __This allows different sessions to use the same temporary table name for different purposes__, whereas the standard's approach constrains all instances of a given temporary table name to have the same table structure.

Bolds mine

raphaellaude:
Changed the UI a bit for creating maps from `/map` to make clear that these are new documents only.

image

raphaellaude:
Success state test cases handled
- [x] Happy path
- [x] No data in CSV
- [x] String zones
- [x] No children -> parents
- [x] Some children -> parents
- [x] All children -> parents
- [x] Unshattered map input
- [x] Some empty zones
- [x] Too many zones

Failure state test cases handled
- [x] Duplicate geoids. Might be nice to eventually gracefully handle this case by skipping dupes?
- [ ] Assignments too big
- [ ] Districtr map not found

Other unimplemented nice to haves
- [ ] Report out number of import errors
- [ ] Report out number of skipped assignments
- [ ] Report out number of matched input geoids not just total assignments inserted
- [ ] Provide summary of block import for user in dialog (don't automatically close)

backend/app/main.py
raphaellaude: I guess I misread the docs. Anna was correct–we do need a unique temp table name 🤷🏻‍♂️

raphaellaude:
Thanks for the review @cdonnay ! I think I handled all of these cases correctly now. This doesn't have the detailed stats documenting the import yet so maybe we can do that in a follow-up

cdonnay:
Hi team! I tried my CSVs again and still was getting a lot of unassigned area issues. See below.
 
- when I upload a completely empty csv file, I still get the issue where the "unassigned area" evaluation thinks the entire state is one unit
Screenshot 2025-05-13 at 9 19 03 AM
- when i upload a csv with a  mix of null geo ids and zones, i get unassigned population but not an unassigned area
Screenshot 2025-05-13 at 9 20 02 AM
- same unassigned issue if i just have null geoids
- same unassigned issue if i have a null row
- unassigned single unit if i have a few non-int convertible str zones
Screenshot 2025-05-13 at 9 22 35 AM
-  too many zones also triggered the unassigned issue
- file that had too many rows (>1mil) was caught and i couldnt upload it, yay!

nofurtherinformation (approved):
great work!

Issue #246, opened 2025-01-13
Users should be able to swap district colors after drawing


nofurtherinformation:
Via 1/3 user feedback notes

nofurtherinformation:
#262

PR #262, opened 2025-01-24
User can change district colors

mapmeld:
As requested in #246 :
Right-clicking on a paint color opens up a color picker widget (many designs available from this library: https://casesandberg.github.io/react-color/ )

Screenshot 2025-01-24 at 3 21 36 PM

The color scheme is stored in `mapStore` , and can be undone / redone
Updating a district color repaints the mapped areas and number markers:

Screenshot 2025-01-24 at 3 21 28 PM

Remaining issues:
- ~color palette should use Districtr-friendly colors? More options / different widget?~
- ~should block users from making identical district colors~
- ~I wasn't able to test population charts locally~
- storing the new color scheme on the server


mapmeld:
Colors are now from the palette of Districtr colors and saved to the DB, so this PR is now ready to review.
It looks like I need to rebase

Screenshot 2025-01-30 at 3 49 00 PM

mapmeld:
Rebased. Updated color palette for the new UI, fixed compatibility with locking zones, and saving color changes caused by undo/redo

Screenshot 2025-02-02 at 10 34 54 PM

app/src/app/components/Toolbar/ColorPicker.tsx
```diff
@@ -115,21 +133,43 @@ export const ColorPicker = <T extends boolean>({
...
                   <RadioGroupIndicator />
+                  <div
+                    style={{
+                      display: i === colorSelectIndex ? 'block' : 'none',
+                      position: 'absolute',
+                      left: 10,
+                      top: 2,
+                    }}
+                  >
+                    <TwitterPicker
```
nofurtherinformation, line 166: H: I think this is a bit tricky being a right click option in the paint menu. Customizing colors is an advanced option so I'd suggest the following workflow:

1. "Customize colors" as an option in the top left hamburger menu
2. Open a modal with the current colors in the plan
3. On clicking color, the TwitterPicker opens up inside the modal just below the colors (maybe defaults to customizing zone 1?)

app/src/app/store/mapRenderSubs.ts
```diff
@@ -34,6 +35,20 @@ export const getRenderSubscriptions = (
...
   useHoverStore: typeof _useHoverStore
 ) => {
+  const addColorSchemeSub = useMapStore.subscribe<MapStore['colorScheme']>(
+    state => state.colorScheme,
+    colorScheme => {
+      removeZoneMetaLayers();
+      debouncedAddZoneMetaLayers({});
+      const {getMapRef, mapDocument, addVisibleLayerIds} = useMapStore.getState();
+      if (mapDocument) {
+        addBlockLayers(getMapRef(), mapDocument);
```
nofurtherinformation, line 45: H: This is a very expensive way to update the colors -- changing layer styles will be substantially easier once we merge in the react-map-gl refactor but I'd suggest just updating style properties instead of fully adding and removing layers

backend/app/alembic/versions/24ba32e7c978_add_color_scheme_column.py
```diff
@@ -0,0 +1,26 @@
...
+branch_labels: Union[str, Sequence[str], None] = None
+depends_on: Union[str, Sequence[str], None] = None
+
+
+def upgrade() -> None:
+    op.add_column('document.document', sa.Column('color_scheme', sa.ARRAY(sa.String()), nullable=True))
+
+
+def downgrade() -> None:
+    op.drop_column('document.document', 'color_scheme')
```
nofurtherinformation, line 26:
```suggestion
def upgrade() -> None:
    op.add_column(
        "document",  
        sa.Column("color_scheme", sa.ARRAY(sa.VARCHAR), nullable=True),
        schema="document"
    )


def downgrade() -> None:
    op.drop_column(
        "document",  
        "color_scheme",
        schema="document"
    )
```

I'm not sure if this is a psycog2/3 issue but the current implementation does not work on my end 
```

File: backend/app/main.py [hunk 4]
```diff
@@ -235,11 +236,34 @@ async def reset_map(document_id: str, session: Session = Depends(get_session)):
...
         )
     )
+    session.execute(text(f"UPDATE document.document SET color_scheme = NULL WHERE document_id = '{document_id}'"))
     session.commit()
 
     return {"message": "Assignments partition reset", "document_id": document_id}
 
 
+@app.patch(
+    "/api/update_colors/{document_id}",
```
nofurtherinformation, line 246: PP: `/api/document/{document_id}/update_colors` may be a more RESTful pattern

backend/app/main.py
```diff
@@ -235,11 +236,34 @@ async def reset_map(document_id: str, session: Session = Depends(get_session)):
...
+    response_model=ColorsSetResult,
```
nofurtherinformation, line 247: PP: Return color array for debugging purposes?

app/src/app/components/Toolbar/ColorPicker.tsx
```diff
@@ -115,21 +133,43 @@ export const ColorPicker = <T extends boolean>({
...
+                  value={String(i)}
+                  onContextMenu={e => {
+                    e.preventDefault();
+                    setColorSelectIndex(i);
+                  }}
+                >
                   <RadioGroupIndicator />
+                  <div
+                    style={{
+                      display: i === colorSelectIndex ? 'block' : 'none',
```
nofurtherinformation, line 160: Medium: Use radix ui themes components + tailwind

nofurtherinformation (changes requested):
Looking very good! 

This needs some merging with main and I added a few other suggestions. The biggest change I'd request is moving this into a modal from the top left menu dropdown which I think will be better for UX.

backend/app/alembic/versions/24ba32e7c978_add_color_scheme_column.py
nofurtherinformation: P.s. This seems to be where the build is failing!

mapmeld:
Rebased / rewrote the color-change UI to be in the dropdown menu and give these large squares

Screenshot 2025-03-03 at 1 57 35 PM

The color-change op is much more efficient with the new React map component

Color change is still an undo/redo-able action

I've accepted the suggested changes to the alembic migration. I am having some weird issues there so hopefully the new one has the right alembic config

nofurtherinformation:
Hey @mapmeld ! Overall this is looking good. I'd like to ask 2 UAT/feature changes:

- Make the UI in the modal a bit smaller on screen, its a bit heavy right now
- Pull in some of the code from this other PR https://github.com/districtr/districtr-v2/pull/290/files to extend the color palette beyond 40 colors and also programmatically generate colors (using some code based on an implementation by Peter)

mapmeld:
I made a little bit of progress, but I wasn't able to make the color swatches smaller, and have the selection radio UI and numbers enclosed in them
I tried switching the radio buttons to Radix's ToggleGroup but it seems to be a separate package? And it did not work as a drop-in replacement
Now that I am looking to do a git merge, I see that mapRenderSubs was removed and I don't know enough about the new scheme

nofurtherinformation (approved):
lgtm

PR #263, opened 2025-01-25
Bugfixes - UI Merge

nofurtherinformation:
A couple bugs came up after the UI revamp merge, this resolves:

- Unassigned not displaying correctly
- Issue with an infinite loop / max call stack exceeded issue

### DIFF
--- a/app/src/app/store/mapRenderSubs.ts
+++ b/app/src/app/store/mapRenderSubs.ts
@@ -61,8 +61,6 @@ export const getRenderSubscriptions = (
       }
 
       const layersToFilter = mapDocument?.child_layer ? CHILD_LAYERS : [];
-
-      if (mapDocument?.child_layer) layersToFilter.push(...CHILD_LAYERS);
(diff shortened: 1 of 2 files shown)


raphaellaude (approved):
lgtm

app/src/app/store/mapRenderSubs.ts
```diff
@@ -61,8 +61,6 @@ export const getRenderSubscriptions = (
       }
 
       const layersToFilter = mapDocument?.child_layer ? CHILD_LAYERS : [];
-
-      if (mapDocument?.child_layer) layersToFilter.push(...CHILD_LAYERS);
```
nofurtherinformation, line 65: This line was where the max call stack issue seemed to occur. I think (not 100% sure) that this was somehow sometimes it was recursively calling the spread operator and making a huge CHILD_LAYERS reference? Or every time you called this is was adding to CHILD_LAYERS since it is a reference?

PR #265, opened 2025-01-29
Load data script update

nofurtherinformation:
Fixes load data script

### DIFF
--- a/backend/load_data.py
+++ b/backend/load_data.py
@@ -66,16 +66,6 @@ def load_sample_data(config):
                 ]
             )
 
-        subprocess.run(
-            [
-                "python3",
-                "cli.py",
-                "create-parent-child-edges",
-                "--districtr-map",
-                view["gerrydb_table_name"],
-            ]
-        )
-
     for view in config["districtr_map"]:
         session = next(get_session())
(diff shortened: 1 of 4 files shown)


backend/load_data.py
```diff
@@ -21,6 +21,12 @@ def load_sample_data(config):
     specified data directory, and for each file, it runs a script to load the
     GerryDB view.
 
+    The order of adding views MUST be:
```
raphaellaude, line 24: Nice to document this, thank you!

backend/load_data.py
```diff
@@ -77,32 +83,41 @@ def load_sample_data(config):
...
-            subprocess.run(
-                [
-                    "python3",
-                    "cli.py",
-                    "create-parent-child-edges",
-                    "--districtr-map",
-                    view["gerrydb_table_name"],
-                ]
-            )
+            args = [
```
raphaellaude, line 86: PP: This all looks good. That said, these calls are all pretty thin wrappers around internal functions in utils. Down to merge this for now but perhaps we should leave a note to refactor out the use of the CLI from this script.

raphaellaude (approved):

PR #267, opened 2025-01-29
Local Caching

nofurtherinformation:
This PR adds clientside caching for gerrydb views and zone assignments for better user experience and lighter server loads.

## Description
- Adds a timestamp `updated_at` sent with update assignment patches to track versioning. This handles asynchronous changes but will not currently alert the FE to changes happening while they are active
- Uses indexeddb via (idb package) to cache gerrydb views, map state, and map state updated at timestamps
- The document table already has an `updated_at` property, so this does not change db schema
- The FE already gets the full document row, so it can check if its latest cached timestamp matches the database without pulling the full assignments
- The idb implementation to cache the state will store the assignments, shattered IDs and shatter mappings; the store state is very close to the in-memory representation to streamline reads and writes
- LocalStorage would work and may be easier, but with this approach would be limited to ~66 cached plans, which for power users could be hit

## Reviewers

- Primary: @raphaellaude 
- Secondary: @mariogiampieri 

## Checklist
- [x] Update timestamps
- [x] Cache map state in idb
- [x] Cache map state updated timestamps in idb
- [x] Cache gerrydb views in idb
- [x] Reload map state from idb conditionally
- [x] Reload gerrydb views from idb
- [x] Test if better to fully stringify or just convert to plain JS objects
- [x] Test perf implications
- [ ] User test / bug squash


nofurtherinformation:
After some testing, storing vanilla js objects and converting to map/set structures is faster than fully stringifying

backend/app/main.py
```diff
@@ -72,6 +72,24 @@ def get_session():
         yield session
 
 
+def update_timestamp(
+    session: Session,
+    document_id: str,
+    updated_at: str,
+    execute: bool = True,
+    commit: bool = False,
```
raphaellaude, line 80: I: Why these two params? Since you're passing the session from the caller, if the caller errors later all operations in the session's transaction will be rolled back. So you can safely execute this w/o worry.

raphaellaude (changes requested):
This works so so well and feels _way_ better for the user. One small change requested on the backend.

Other than that, I noticed that when the user reloads we load from the local cache OR the remote store. I'm wondering if there's a case for the AND approach: first load from the local store, then verify with the remote store. Since we're in the single user per map paradigm I figure this approach is fine, but wanted to discuss trade-offs. Thoughts?

backend/app/main.py
nofurtherinformation: Ah, that's a good call!

nofurtherinformation:
> This works so so well and feels _way_ better for the user. One small change requested on the backend.

Thank you!

> Other than that, I noticed that when the user reloads we load from the local cache OR the remote store. I'm wondering if there's a case for the AND approach: first load from the local store, then verify with the remote store. Since we're in the single user per map paradigm I figure this approach is fine, but wanted to discuss trade-offs. Thoughts?

That's a good question -- we take this approach for the map views caching (local but verify). I think my two questions here are:
- Is it worth saving the extra load / capacity on the backend?
- If the user paints before the backend assignments respond, how do we reconcile the state?

backend/app/main.py
```diff
@@ -72,6 +72,22 @@ def get_session():
         yield session
 
 
+def update_timestamp(
+    session: Session,
+    document_id: str,
+    updated_at: str,
+    execute: bool = True,
```
raphaellaude, line 79: Sorry if my previous suggestion wasn't clear. I would both params entirely and _always_ execute.

```suggestion
```

backend/app/main.py
```diff
@@ -72,6 +72,22 @@ def get_session():
...
+):
+    update_stmt = (
+        update(Document)
+        .where(Document.document_id == document_id)
+        .values(updated_at=updated_at)
+    )
+    if execute:
+        session.execute(update_stmt)
+        session.commit()
```
raphaellaude, line 88: And let the parent caller commit.

```suggestion
    session.execute(update_stmt)
```

raphaellaude (commented):
> If the user paints before the backend assignments respond, how do we reconcile the state?

Hmm yeah true that would be difficult. You've convinced me that we shouldn't pull.

Another approach might be to do a full document _push_ to the server using the local cache then?

nofurtherinformation:
> > If the user paints before the backend assignments respond, how do we reconcile the state?
> 
> Hmm yeah true that would be difficult. You've convinced me that we shouldn't pull.
> 
> Another approach might be to do a full document _push_ to the server using the local cache then?

I believe this is the case currently but will double check tomorrow!

raphaellaude (approved):
lgtm – we can leave the push pull discussion for another time / an issue rather than PR

backend/app/main.py
nofurtherinformation: Ahh thank you -- apologies for the misfire on my end :)

raphaellaude (approved):
apparently the above errors were from a local build so re-approving!

PR #269, opened 2025-01-31
Clientside data optimization

nofurtherinformation:
This PR is an experimental draft that works to resolve a few issues.

## Description
- map.queryRenderedFeatures (aka QRF) is slow
- We do not have a complete ID list on map load, for doing paint by county
- We do not have all small geometries, for painting
- We do not have all demographic data, for doing evaluation

This PR proposes the following solution:
- Pre-process bbox-based simplified selection areas of appropriate compactness (more on this later)
- Pre-process stats
- Consume stats and selection areas in the frontend through an RBush-based class

As of 1/31 this only works with Colorado VTDs. There is a button on the top of the menu to toggle RTree vs queryRenderedfeatures. The console will log the average time to query 1000 features in ms. The speed gains here on fast clients are minimal (~6ms average per 1000 features QRF vs 1ms RTree) but on slower machines this becomes pretty meaningful (~30ms QRF vs 3ms RTree). This is aside from the client information/data gains.

So far, Colorado's naive data representation for this is ~450kb once gzip'd and the RBush library is around 2kb gzip'd.

## Reviewers
- Primary:
- Secondary:

## Checklist
- [ ] Test options


nofurtherinformation:
Closing for now. Lower priority.

PR #279, opened 2025-02-08
React map gl refactor

nofurtherinformation:
Refactor to react map gl.

## Description
- Why: Better component API for managing complex layer mounting unmounting
- Refactors to react map gl
- Updates some data handling for the geometry worker and handles more tile stuff on the worker thread
- Tries to maintain mapRef API where possible, most everything is compatible 
- Cleans up subscriptions - we may be able to consolidate more

## Reviewers
- Primary: @raphaellaude 
- Secondary: @mapmeld


app/src/app/components/Toolbar/Settings.tsx
```diff
@@ -65,7 +65,7 @@ export const ToolSettings: React.FC = () => {
         value={[
           mapOptions.higlightUnassigned === true ? 'higlightUnassigned' : '',
           mapOptions.showPopulationTooltip === true ? 'showPopulationTooltip' : '',
-
+          mapOptions.showBlockPopulationNumbers === true ? 'showBlockPopulationNumbers' : '',
```
raphaellaude, line 68: nit

```suggestion
          mapOptions.showBlockPopulationNumbers ? 'showBlockPopulationNumbers' : '',
```

raphaellaude (commented):
Changes look good but I'm noticing some issues in the test app:

- "Show total population labels on blocks" doesn't work (was this ever implemented?)
- District numbers don't show
- Show/hide county names seems to remove only one set of county labels

image

Otherwise everything pretty much works as expected and the syntax is def an improvement

nofurtherinformation:
> Changes look good but I'm noticing some issues in the test app:
> 
> * "Show total population labels on blocks" doesn't work (was this ever implemented?)
> * District numbers don't show
> * Show/hide county names seems to remove only one set of county labels

Thank you @raphaellaude! 

The population labels on blocks and district numbers was related to an issue with Sentry annotating additional properties onto components, which did not play nicely with some of the component API. The new major release of `@sentry/nextjs` exposes an option to exclude components from this annotation.

County labels had some state + properties combos flipped (oops), but is now resolved in. addition to some general clean up of the map options and rendering.

Please re-review at your convenience! 

Screenshot 2025-02-13 at 2 29 24 PM
Screenshot 2025-02-13 at 2 29 54 PM

raphaellaude (changes requested):
looks good – there's still one county label layer visible, though, but then should be good to go!

image

nofurtherinformation:
This matches the current styling on main — I think that option had been “emphasize county labels” or “show prominent county labels” previously.

Wdyt about merging this as is but making a ticket to clarify how we want county labels to display?

raphaellaude (approved):

PR #288, opened 2025-02-27
Organize management commands

raphaellaude:
## Description

- Update management commands to include batch load data command
- Refactor load data script
- Pull GPKGs from S3 if they're not available
- Support data load config to support batch data updates
- Update tileset commands to support batch loading, too
- Create configs for new VAP data
    - Districtr maps
    - Tilesets

## Reviewers

- Secondary: @mariogiampieri

## Checklist


backend/management/load_data.py
```diff
@@ -0,0 +1,244 @@
...
+        session = next(get_session())
+        u = _create_districtr_map(
+            session=session,
+            name=view.name,
+            gerrydb_table_name=view.gerrydb_table_name,
+            parent_layer=view.parent_layer,
+            child_layer=view.child_layer,
+            tiles_s3_path=view.tiles_s3_path,
+            num_districts=view.num_districts,
+        )
```
nofurtherinformation, line 223: H: Add extent and available summary stats

nofurtherinformation (changes requested):
One small suggestion and just needs extent+summary stats and then good to go!

backend/management/load_data.py
raphaellaude: added!

raphaellaude:
There are two issues I'm seeing which are unrelated to this PR:

- Neither P1 nor P4 match the new gpkg's so there are no summary stats
- Weirdly, the population bars are not rendering, even though the total pop endpoint is returning data just fine. Have yet to investigate this...

image

nofurtherinformation (approved):
lgtm

PR #293, opened 2025-03-07
Smarter district number label placement

nofurtherinformation:
Small improvements to label placement for zone labels

## Description
- Makes the placement of labels _more_ random, which makes it seem like there is less of a "preference" in where labels
- Turns off zone labels when in block view and population numbers are turned on
- Adds a minimum buffer (50 pixels) when choosing location for a district label

## Reviewers
- Primary: @mariogiampieri 
- Secondary: @raphaellaude


app/src/app/components/Map/MetaLayers.tsx
```diff
@@ -91,11 +93,13 @@ const ZoneNumbersLayer = () => {
     }
   };
 
-  useEffect(handleUpdate, [showZoneNumbers, zoneAssignments]);
+  useLayoutEffect(handleUpdate, [showZoneNumbers, zoneAssignments]);
```
raphaellaude, line 96: nice

app/src/app/components/Map/MetaLayers.tsx
```diff
@@ -91,11 +93,13 @@ const ZoneNumbersLayer = () => {
...
 
   useEffect(() => {
     const map = getMapRef();
     if (map) {
+      map.loadImage('/lock.png')
+        .then(image => map.addImage('lock', image.data));
```
raphaellaude, line 102: PP: Might make sense to show an actual message at some point since a lock is a bit ambiguous.

app/src/app/utils/GeometryWorker/geometryWorker.ts
```diff
@@ -247,20 +236,27 @@ const GeometryWorker: GeometryWorkerClass = {
...
+      const minimumDistance = minBuffer ?? CENTROID_BUFFER_KM
+      // convert 30pixels
       Object.entries(this.previousCentroids).forEach(([zone, centroid]) => {
         if (!centroid.properties?.id || !centroid.properties.zone) return;
         const props = centroid.properties;
-        const currZone = this.activeGeometries[props.id].properties.zone;
-        if (currZone === null || (currZone !== props.zone && activeZones.includes(currZone)))
-          return;
+        const currZone = this.activeGeometries[props.id]?.properties?.zone;
+        if (!currZone || currZone !== props.zone && activeZones.includes(currZone)) return;
```
raphaellaude, line 245: PP: for complex logical statements like these, adding a comment or grouping terms into variables would be a bit clearer. Also my understanding of the code here is a bit iffy so pls rename this variable if you implement this idea!

```suggestion
        const currZoneIsPainted = currZone !== props.zone && activeZones.includes(currZone)
        if (!currZone || currZoneIsPainted) return;
```

app/src/app/utils/GeometryWorker/geometryWorker.types.ts
```diff
@@ -70,9 +70,10 @@ export type GeometryWorkerClass = {
    * @returns The centroids and dissolved outlines of the parsed features within the view.
    */
   getCentroidsFromView: (props: {
-    bounds: [number, number, number, number];
-    activeZones: number[];
-    fast?: boolean;
+    bounds: [number, number, number, number],
+    activeZones: number[],
+    fast?: boolean,
```
raphaellaude, line 75: PP: Prefer to avoid boolean flags (though of course I do them too!). Would prefer descriptive strategy pattern e.g.

```suggestion
    labelPlacementMethod?: LabelPlacementMethod,
```

where `LabelPlacementMethod` is an enum e.g.

```ts
enum LabelPlacementMethod {
    NonCollidingCentroids = (...props) => void, // whatever this should be
    CenterOfMass = (...props) => void, // whatever this should be
    // etc.
}
```

Sidebar: I've been thinking we should start to implement some feature flags so we can test these different ideas in prod w/o blocking PRs while allowing the features to bake for a bit.

raphaellaude (commented):
Happy to approve but wanted to discuss a few ideas first.

app/src/app/utils/GeometryWorker/geometryWorker.types.ts
raphaellaude: here's an issue for the feature flag idea #308

nofurtherinformation: Agreed on feature flags and avoiding boolean flags!

In this case passing a function directly to the `GeometryWorker` won't quite work since it gets serialized, but using specified strings and internal methods to the worker (eg. `getCentroidsFromView(bounds: number[], strategy: 'centerOfMass' | 'nonCollidingCentroids') should have the same effect!

app/src/app/components/Map/MetaLayers.tsx
nofurtherinformation: ty!

app/src/app/components/Map/MetaLayers.tsx
nofurtherinformation: Agreed, let's chat about this tomorrow with the group!

app/src/app/utils/GeometryWorker/geometryWorker.ts
nofurtherinformation: Agreed, good call

raphaellaude (approved):
looks good

PR #297, opened 2025-03-10
New gerrydb data schema

nofurtherinformation:
Incorporates new data from GerryDB, which is the new schema.

## Description
- Refactors P1 and P4 schemas to `TOTPOP` and `VAP` to be more semantic
- Cleans up some issues that came up in running the data pipeline
- Starts to incorporate election data, but currently the ambition is not to do all of it

## Reviewers
- Primary: @raphaellaude 
- Secondary: @bailliekova 

## Checklist
- [x] Update BE models
- [x] Update FE types
- [x] Test load data cli command on raw db state
- [x] Update tests
- [ ] Update docs


nofurtherinformation:
After some fussing around with Alembic downgrades from a populated database state to base, I gave up and just force dropped the districtr database and created a new one. There are a couple tweaks here but from a totally empty database `alembic upgrade head && python cli.py batch-create-districtr-maps ...` will one-shot build the database

nofurtherinformation:
Also, Texas breaks the memory limit. I upped `297-api` to 2 vcpus / 4gb ram on Fly

raphaellaude (dismissed):
lgtm

PR #306, opened 2025-03-13
Static site

nofurtherinformation:
A (big) PR that builds out the basic static site functionality and content management system. This first iteration focuses on two CMS constructs -- tags, for specific events of efforts, and places, for MGGG managed pages likes states or cities.

This PR does **not** handle data management and custom module ingestion.

## Description
- Based on team discussion, we chose to implement a database-oriented approach
- This approach uses a new schema `cms` with api endpoints to manage the content
- The core requirements of the schema are: draft and publish content workflow, rich text, multi-language support
- Each content type has a table, allowing for particular properties as needed
- The API is designed for future work to enable auth that bifurcates permissions, specifically editing, creating, and publishing
- The cms tables have a draft and published JSONB schemas that use TipTap as a content description language
- Currently, content is sent to the client and rendered, but it is the hope to render all content serverside as HTML
- This PR also adds from refactors to frontend matters, specifically API handlers which now live in their own files. This is more verbose, but given side effects and subscriptions, this is a much more stable and isolated way to organize these. An issue that came up, for example, when importing the api handler to get gerry db views from a static page, something related to shattering threw an error because the map wasn't instantiated. This organization is also better for code splitting across different pages.

Big todos and next steps:
1. Auth
2. Clean up styles
3. Build out custom components for frontend / pages 
4. Populate site with additional content

## Reviewers
- Primary: @raphaellaude  
- Secondary: @mariogiampieri 

## Checklist
- [x] Added CMS DB schema
- [x] Added endpoints for: create, update, publish, delete, list content, get piece of content
- [x] Added tests for each endpoint
- [x] Added pages to view lists of Tags and Places
- [x] Added tag and place page
- [x] Added CMS editor page for Tags and Places
- [x] Workflow from place or tag page to created map

Screenshot 2025-03-27 at 5 10 50 PM
Screenshot 2025-03-27 at 5 24 31 PM
Screenshot 2025-03-27 at 5 11 17 PM
Screenshot 2025-03-27 at 5 12 01 PM


backend/app/cms/main.py
```diff
@@ -0,0 +1,241 @@
+import uuid
+from typing import List
+from fastapi import APIRouter, Depends, HTTPException, status, Query
+from sqlmodel import Session, select
+from sqlalchemy.exc import MultipleResultsFound, NoResultFound
+from sqlalchemy.sql import literal_column
+import logging
+
+from app.core.db import engine
+from app.models import (
```
raphaellaude, line 10: I: Since you've already put these routes in a separate module I'd move these models into the cms module, too

backend/app/cms/main.py
```diff
@@ -0,0 +1,241 @@
...
+    """Create a new CMS content entry"""
+    # Check if content with same slug and language already exists
+    CmsModel = CMS_MODEL_MAP[type]
+    existing = session.exec(
+        select(CmsModel)
+        .where(CmsModel.slug == data.slug)
+        .where(CmsModel.language == data.language)
+    ).first()
+
+    if existing:
```
raphaellaude, line 48: I: Having a patch endpoint will be useful

backend/app/cms/main.py
```diff
@@ -0,0 +1,241 @@
...
+
+
+@router.post(
+    "/content/{type}",
+    response_model=CmsContentCreate,
+    status_code=status.HTTP_201_CREATED,
+)
+async def create_cms_content(
+    type: CMSContentTypesEnum,
+    data: CmsContentCreate,
```
raphaellaude, line 36: I: The content type might make more sense as a field on the `CmsContentCreate` model

backend/app/cms/main.py
```diff
@@ -0,0 +1,241 @@
...
+    #         literal_column(CMSModel.districtr_map_slugs.name) == districtr_map_slugs
+    #     )
+
+    query = query.offset(offset).limit(limit)
+    results = session.exec(query).all()
+
+    return results
+
+
+@router.get("/content/{type}/{slug}", response_model=CMSContentPublicWithLanguages)
```
raphaellaude, line 109: H: Across all of these routes I would be a little more verbose in your naming conventions. Also, putting two params next to each other could lead to conflicts. You'll need to ensure all routes are unique.

backend/app/cms/main.py
nofurtherinformation: Sounds good!

backend/app/cms/main.py
nofurtherinformation: Agreed

backend/app/cms/main.py
nofurtherinformation: Agreed

backend/app/cms/main.py
nofurtherinformation: That makes sense. How about something like this, echoing the main API structure?

- `POST /api/cms/create_content` (content type internal to body)
- `PATCH /api/cms/content/{content_id}/update`
- `POST /api/cms/content/{content_id}/publish`

- `GET /api/cms/list/{content_type}`
- `GET /api/cms/get/{content_type}/{slug}`

raphaellaude: It's fine if they're repeated for POST/PATCH/GET/DELETE etc. but you'll run into issues with `{content_type}/{slug}`. How about?

- `POST   /api/cms/content` (all body)
- `PATCH  /api/cms/content` (all body)
- `POST   /api/cms/content/{content_id}/publish`
- `DELETE /api/cms/content/{content_id}`
- `GET    /api/cms/content/{content_id}`
- `GET    /api/cms/content/type/{content_type}` (to list content)
- `GET    /api/cms/content/{content_id}/type/{content_type}`

nofurtherinformation: Sounds good - thanks!

nofurtherinformation: I landed on something close to this, but except for getting content we do need to use the type and slug since the front end request will not have the content ID

raphaellaude (commented):
A few notes (happy for these to be handled in follow-up PRs or here, up to you):

- CMS editor should maybe disable save until after any edits have been made. Save for the edit button in the content list (since you're already editing!)
- PP: `Content with slug 'my-super-awesome-map' and language 'LanguageEnum.ENGLISH' already exists` error is great to have but the language enum bit could be cleaned up
- I: If you edit and then publish while editing, the content list does not refresh properly (see screenshot). I think this is because the `/list` and `/publish` events fire simultaneously so are in a race condition.
- H: I was able to publish a map with an empty slug (and another with an empty title, for that matter). I've pushed a backend change to address this but still needs to be handled on the FE.
- PP: The content list reordering on save is a little bit jarring. I would order based on `created_at` so that the order is stable
- Q: Should the slug be hard-coded after the initial edit? Users can always create a new map, but they can't fix broken links they've sent out

image

Images work nicely! :)

image

nofurtherinformation:
@raphaellaude thanks very much for this review! I just pushed some changes that I believe resolve these issues.

raphaellaude (approved):
lgtm! nicely done!

PR #309, opened 2025-03-14
Misc bugfixes

nofurtherinformation:
A potpourri of fixes 

## Description
- Fix connected components useQuery state issues
- Fix error on color palette migration repeat, issue when adding column in sqlalchemy that already existed
- Handle color palette error better
- Keep county FIPS code when switching to plan of same state
- Fix paint brush cursor location
- Clarify difference from ideal population labels on population bars with stroke (for dark backgrounds) and explicit "+" for greater than ideal
- When ideal population is a fraction (eg. 1280.5) no longer show +/-1 
- Remove vestigial population tooltip
- Correctly clear out old zone assignments when changing documents of the same tileset
- Fix display issue on top-to-bottom deviation not showing even when all zones are painted, or calculated incorrectly with unassigned population as a zone

## Reviewers
- Primary: @raphaellaude 
- Secondary: @mariogiampieri


backend/app/alembic/versions/4b0aec5f8350_save_color_palette.py
```diff
@@ -20,6 +20,12 @@
 
 
 def upgrade() -> None:
+    # check if column exists
+    conn = op.get_bind()
+    inspector = sa.engine.reflection.Inspector.from_engine(conn)
+    columns = inspector.get_columns("document", schema="document")
+    if "color_scheme" in [column["name"] for column in columns]:
+        return
```
raphaellaude, line 28: What is the root of the issue here? This looks pretty inelegant

app/src/app/components/sidebar/MapValidation/ZoomToConnectedComponents.tsx
```diff
@@ -35,33 +34,22 @@ export default function ZoomToConnectedComponents({
...
           </IconButton>
         </Tooltip>
       </Flex>
-      {showZoom && isLoading && <Spinner />}
+      {showZoom && !data && !error && <Spinner />}
       {showZoom && error && (
         <Blockquote color="red">Error fetching connected components</Blockquote>
       )}
-      {!!(showZoom && isFetched && data) && (
+      {!!(showZoom && !isLoading && data) && (
```
raphaellaude, line 52: I: This is definitely looking better, thank you! The only issue I'm seeing now is when the number of connected components has changed, in some cases the cached / previous number of connected components is shown before being updated shortly thereafter.

raphaellaude (commented):
Looks good but I'd like to figure out this migration thing before merging!

backend/app/alembic/versions/4b0aec5f8350_save_color_palette.py
nofurtherinformation: So, this is purely devex -- when starting up docker compose and 'alembic upgrade head' after you've already run the is migration, I was seeing an error get thrown because the column already existed. 

This is low priority so if this needs more thought we can scrap this change and handle it later

app/src/app/components/sidebar/MapValidation/ZoomToConnectedComponents.tsx
nofurtherinformation: Will take a look this morning!

backend/app/alembic/versions/4b0aec5f8350_save_color_palette.py
raphaellaude: Ah got it – if you've already run the migration then the migration version should be set in the `alembic_version` table and the migration won't be applied.

```
districtr=# select * from alembic_version;
 version_num
--------------
 6d83465440a6
(1 row)
```

nofurtherinformation: I thought that was the case, but it was still running for reasons I can't explain 🤷🏼 perhaps swapping between branches it got mixed up!

app/src/app/components/sidebar/MapValidation/ZoomToConnectedComponents.tsx
nofurtherinformation: Resolved in ce0e54d6f6b6e97ed438d2fe0845d92911bf1692

raphaellaude (approved):
lgtm

PR #310, opened 2025-03-14
Handle legacy map requests

nofurtherinformation:
A quick fix to fix errors 

## Description
- Returns demography requests with 410 

## Reviewers
- Primary: @raphaellaude 
- Secondary:


raphaellaude (dismissed):

raphaellaude (approved):
Looks good. At some point we'll likely want a solution that allows legacy maps to remain supported but this is good for now

PR #314, opened 2025-03-17
Create thumbnails endpoint

mapmeld:
Use GeoPandas to query the postgres database and render a PNG. Example:

image

- Connects directly to Postgres query (no toGeoJSON / ST_AsText go-between)
- Uses `color_scheme` from document or default colors
- Renders unassigned and NULL zones in #CCCCC light grey

The matplotlib import and `plt.close()` line is a little obscure, I believe it was in the original thumbnail code to remove the plot from memory and prevent redrawing the map over and over

I'm returning the base64 png as a JSON string, because we used to call this and save the string directly to the DB sort-of async, i.e. not expecting to return the image to the client right away. We could rework this to return the PNG binary file, or upload to S3


mapmeld:
Rebased, moved thumbnail code into its own file, and added code which I think connects us to S3

- going to `/document_id/thumbnail` uses S3 head to check if the thumbnail exists and fetches it; if it's not there it fetches the map image and returns it as a `StreamingResponse`
- a PUT request to `/document_id/put_thumbnail` would tell the server to generate or re-generate the map (we can call this when clicking "Share" or from a cronjob to update recent maps)
- if S3 fails, it should either stream or redirect to a fallback image. For now I redirected to the path to the megaphone image

Notes:
- did not test S3 or shattered maps
- we could reorganize the fetch thumbnail code to try fetching and regenerate the map if any error occurs
- put_thumbnail and the S3 upload steps could be run as a child process to make the response faster

mapmeld:
I made a comment in https://github.com/districtr/districtr-v2/pull/333 - that added thumbnail headers, but there are issues making cards for individual maps unless that is on a `page.tsx`

mapmeld:
Updated the S3 and database queries, and confirmed it is now working on shattered maps with a mix of VTDs and blocks (green area is blocks)

image

The way this works is giving geopandas all geometry of the parent layer (assigned or not, so we get the whole state), union with assigned blocks from the child layer

backend/app/main.py
```diff
@@ -1083,6 +1085,28 @@ async def get_projects(
...
+    session: Session = Depends(get_session),
+):
+    try:
+        img_io = fetch_thumbnail(session, document_id)
+        return StreamingResponse(content=img_io, media_type="image/png")
+    except:
+        return RedirectResponse(url="/home-megaphone.png")
+
+
+@app.put("/api/document/{document_id}/put_thumbnail", status_code=status.HTTP_200_OK)
```
nofurtherinformation, line 1101: H: A `POST` request would be more appropriate since the request body is not consumed. Also, `api/document/{document_id}/create_thumbnail` or just `api/document/{document_id}/thumbnail` follows existing patterns better

backend/app/main.py
```diff
@@ -1083,6 +1085,28 @@ async def get_projects(
...
+async def put_thumbnail(
+    *,
+    document_id: str,
+    session: Session = Depends(get_session),
+):
+    generate_thumbnail(session, document_id)
+    return {}
```
nofurtherinformation, line 1108: H: Something returned here might be helpful

backend/app/thumbnails.py
```diff
@@ -0,0 +1,137 @@
...
+    "#FF80AB",
+    "#D81B60",
+    "#26A69A",
+    "#FFEA00",
+    "#6200EA",
+]
+
+
+# receives the document_id and returns whether a thumbnail exists on S3
+def thumbnail_exists(document_id: str) -> bool:
```
nofurtherinformation, line 59: H: docstring

backend/app/thumbnails.py
```diff
@@ -0,0 +1,137 @@
...
+    assert s3, "S3 client is not available"
+    object_information = s3.head_object(
+        Bucket=THUMBNAIL_BUCKET, Key=f"thumbnails/{document_id}",
+    )
+    return object_information["ResponseMetadata"]["HTTPStatusCode"] == 200
+
+
+# receives the current session and requested document_id
+# returns io stream of generated image
+def generate_thumbnail(session: Session, document_id: str) -> io.BytesIO:
```
nofurtherinformation, line 70: H: docstring

backend/app/thumbnails.py
```diff
@@ -0,0 +1,137 @@
...
+    GROUP BY zone
+    """
+    if child_layer is not None:
+        sql += f"""UNION
+        (SELECT ST_Union(geometry) AS geom, zone
+        FROM "document.assignments_{document_id}" assigned
+        INNER JOIN gerrydb.{child_layer} blocks ON blocks.path = assigned.geo_id
+        WHERE zone IS NOT NULL
+        GROUP BY zone)"""
+    conn = conn = session.connection().connection
```
nofurtherinformation, line 102: H: I may be misunderstanding what you are trying to do here, but possible just `conn = session.get_bind().raw_connection()` ?

backend/app/main.py
```diff
@@ -1083,6 +1085,28 @@ async def get_projects(
     ).all()
     return gerrydb_views
 
+@app.get("/api/document/{document_id}/thumbnail", status_code=status.HTTP_200_OK)
+async def get_thumbnail(
```
nofurtherinformation, line 1089: H: What is the use case you are thinking for this? I think we can just use the CDN path (eg `https://tilesets1.cdn.districtr.org/thumbnails/f3c43f6b-b338-4475-b686-70268b9f16d8`) with a fallback default image for the frontend uses eg:
```
<object
  type="image/png"
  data="https://somedomain.com/image.png"
  width="150"
  height="150"
  aria-label="This image should exist, but alas it does not"
>
  <img src="/path/to/fallback.png" alt="Fallback image" width="150" height="150" />
</object>
```
Would this get called somewhere else?

backend/app/thumbnails.py
```diff
@@ -0,0 +1,137 @@
...
+            ContentType="image/png",
+        )
+    except:
+        logger.info("Could not upload thumbnail")
+
+    return pic_IObytes
+
+
+# fetch current thumbnail or generate one
+def fetch_thumbnail(session: Session, document_id: str) -> io.BytesIO:
```
nofurtherinformation, line 130: See above re: the get thumbnail endpoint

backend/app/main.py
```diff
@@ -1083,6 +1085,28 @@ async def get_projects(
...
+):
+    try:
+        img_io = fetch_thumbnail(session, document_id)
+        return StreamingResponse(content=img_io, media_type="image/png")
+    except:
+        return RedirectResponse(url="/home-megaphone.png")
+
+
+@app.put("/api/document/{document_id}/put_thumbnail", status_code=status.HTTP_200_OK)
+async def put_thumbnail(
```
nofurtherinformation, line 1102: H: This endpoint is pretty heavy. I tried it with a texas map and saw about 15 seconds of blocking time. Would [FastAPI background tasks](https://fastapi.tiangolo.com/tutorial/background-tasks/#add-the-background-task) or better yet a lambda or cron job make sense here? CC @raphaellaude

backend/app/thumbnails.py
```diff
@@ -0,0 +1,137 @@
...
+        color_scheme = DISTRICT_COLORS
+
+    def coloration(row):
+        if row["zone"] is None or math.isnan(row["zone"]):
+            return "#CCCCCC"
+        else:
+            return color_scheme[int(row["zone"]) % len(color_scheme)]
+
+    sql = f"""
+    SELECT ST_Union(geometry) AS geom, zone
```
nofurtherinformation, line 90: Q: This union is heavy. Are there any performance gains using `ST_Collect` to make feature collections and then letting the API layer deal with rendering?

nofurtherinformation, line 90: Q: This union is heavy. Are there any performance gains using `ST_Collect` to make feature collections and then letting the API layer deal with rendering?

nofurtherinformation (changes requested):
The logic here works and the output thumbnail maps look good! 

The main things we need to add to this are:
- Tests and docstrings
- Handling the thumbnail generation in a non-blocking way

There are a few other suggestions here but those are the big two -- let me know if you want to pair up / chat about any of this!

nofurtherinformation:
P.s. After some thinking and chatting with @raphaellaude , here are some possible strategies that might be good instead of running this on an API endpoint:
- Running a cron job to build the thumbnails using this workflow (geopandas + matplotlib) daily with maps that have changed
- Running a cron job using Selenium (or similar) daily with maps that have changed
- Painting the maps clientside and uploading images via presigned S3

backend/app/main.py
mapmeld: ✅

backend/app/thumbnails.py
```diff
@@ -0,0 +1,137 @@
...
+    plt.close(geoplt.figure)
+    pic_IObytes.seek(0)
+
+    # write to S3
+    try:
+        s3 = settings.get_s3_client()
+        assert s3, "S3 client is not available"
+        s3.put_object(
+            Bucket=THUMBNAIL_BUCKET,
+            Key=f"thumbnails/{document_id}",
```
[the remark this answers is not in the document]
mapmeld: changed 3 lines with `Key=` to have `.png` ✅

backend/app/main.py
mapmeld: OK since it's accessible through that bucket URL that sounds like what I want to do instead of copying.
I'm thinking ahead to the OpenGraph thumbnail URL, where we want a GET path, a request means we want to generate that thumbnail, and if the S3 image doesn't exist we can redirect to a fallback image.
So I think what I want to do is, redirect to S3 or redirect to the fallback.

backend/app/main.py
mapmeld: I'll switch to background task

backend/app/main.py
```diff
@@ -1084,6 +1086,32 @@ async def get_projects(
...
 
+@app.get("/api/document/{document_id}/thumbnail", status_code=status.HTTP_200_OK)
+async def get_thumbnail(
+    *,
+    document_id: str,
+    session: Session = Depends(get_session),
+):
+    try:
+        if thumbnail_exists(document_id):
+            return RedirectResponse(url=f"https://tilesets1.cdn.districtr.org/thumbnails/{document_id}.png")
```
nofurtherinformation, line 1097: H: Please use an environment variable and settings `from app.core.config import settings` instead of hardcoding the base CDN url

backend/app/main.py
```diff
@@ -1084,6 +1086,32 @@ async def get_projects(
...
+
+
+@app.post("/api/document/{document_id}/thumbnail", status_code=status.HTTP_200_OK)
+async def make_thumbnail(
+    *,
+    document_id: str,
+    background_tasks: BackgroundTasks,
+    session: Session = Depends(get_session),
+):
+    background_tasks.add_task(generate_thumbnail, session=session, document_id=document_id)
```
nofurtherinformation, line 1111: Q: Are there any options to log errors with background tasks?

backend/app/main.py
```diff
@@ -1084,6 +1086,32 @@ async def get_projects(
...
+        else:
+            return RedirectResponse(url="/home-megaphone.png")
+    except:
+        return RedirectResponse(url="/home-megaphone.png")
+
+
+@app.post("/api/document/{document_id}/thumbnail", status_code=status.HTTP_200_OK)
+async def make_thumbnail(
+    *,
+    document_id: str,
```
nofurtherinformation, line 1107: H: Please validate that document_id exists in the db before firing off background task

backend/app/thumbnails.py
```diff
@@ -0,0 +1,142 @@
+import geopandas
```
nofurtherinformation, line 1: PP: `thumbnails.py` might make more sense in `app/core` than `app` for code organization purposes; please move the file and adjust imports

nofurtherinformation (changes requested):
A few small comments here to help improve stability and debugging

backend/app/thumbnails.py
mapmeld: 👍

backend/app/main.py
mapmeld: Do I need to add this to settings? I will use `{settings.CDN_URL}` for now

backend/app/main.py
mapmeld: From some googling, [the logger should work](https://stackoverflow.com/questions/75145424/fastapi-starlette-how-to-handle-exceptions-inside-background-tasks). I'll upgrade logger.info to `logger.error` calls for the existing sections

backend/app/core/thumbnails.py
```diff
@@ -0,0 +1,143 @@
...
+    LEFT JOIN "document.assignments_{document_id}" assigned ON geos.path = assigned.geo_id
+    GROUP BY zone
+    """
+    if child_layer is not None:
+        sql += f"""UNION
+        (SELECT ST_Collect(geometry) AS geom, zone
+        FROM "document.assignments_{document_id}" assigned
+        INNER JOIN gerrydb.{child_layer} blocks ON blocks.path = assigned.geo_id
+        WHERE zone IS NOT NULL
+        GROUP BY zone)"""
```
raphaellaude, line 115: H: There's a UDF that will handle all this for you. [Here's some code from district boundary exports](https://github.com/districtr/districtr-v2/blob/main/backend/app/exports/main.py#L73-L80)
```suggestion
    sql = f"""WITH geos AS ( SELECT * FROM get_zone_assignments_geo('{document_id}'::UUID) )
        SELECT
            zone::TEXT AS zone,
            ST_Union(geometry) AS geometry
        FROM geos
        GROUP BY zone
        """
```

raphaellaude (changes requested):

backend/app/main.py
```diff
@@ -965,6 +969,50 @@ async def get_projects(
...
+):
+    try:
+        if thumbnail_exists(document_id):
+            return RedirectResponse(
+                url=f"{settings.CDN_URL}/thumbnails/{document_id}.png"
+            )
+        else:
+            return RedirectResponse(url="/home-megaphone.png")
+    except:
+        return RedirectResponse(url="/home-megaphone.png")
```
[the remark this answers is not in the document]
mapmeld: I think this is assuming that thumbnails have file system errors like local images, when `thumbnail_exists` checks S3

backend/app/core/thumbnails.py
nofurtherinformation: If we are making thumbnails I would assume these are complete maps?

PR #319, opened 2025-03-20
Transition to separate parent and child tilesets

nofurtherinformation:
Due to performance reasons, we need to separate out the parent and child tilesets. 

## Description
- This will optimize rendering and selection speed.
- Ideally we can migrate with as minimal changes as possible
- Rest of the API should remain the same

## Reviewers
- Primary: @mariogiampieri 
- Secondary: @raphaellaude 

## Checklist
- [ ] Change FE layer configs
- [ ] Probably rename layers and sources, sanely 
- [ ] Update map events and interactions as needed
- [ ] Upload required data
- [ ] Update tilesets pipeline
- [ ] Drop tilesets column ?

### DIFF
--- a/app/src/app/components/Map/DemographicLayer.tsx
+++ b/app/src/app/components/Map/DemographicLayer.tsx
@@ -3,7 +3,6 @@ import {
   BLOCK_HOVER_LAYER_ID_CHILD,
   BLOCK_LAYER_ID,
   BLOCK_LAYER_ID_CHILD,
-  BLOCK_SOURCE_ID,
   LABELS_BREAK_LAYER_ID,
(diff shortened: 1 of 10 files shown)


raphaellaude:
Feels maybe slightly faster but it's hard to tell. WDYT?

mariogiampieri:
I'm also testing side by side in Texas shattering  etc. and can't entirely tell a perf difference. Are there certain situations that you encountered where the difference was more pronounced to test on?

nofurtherinformation:
Let's chat with Moon today -- I was previously able to find some cases where zooming out after being very zoomed in saw some delayed pop-in, but I'm not able to replicate just now

PR #323, opened 2025-03-24
Fixes #322 - Faster connected component bboxes

### LINKED ISSUE #322: Connected component contiguity performance
It's distractingly slow. Need better solution

raphaellaude:
## Description

- Improves #322 , addressing perf (and db OOM issues). Better but still not an awesome ux imo
- Does this mainly by (1) moving the reprojection to the last step so it's only run on the final bbox coords and (2) not dipping back into the db and just iterating over the results in python. Not a terribly attractive/satisfying implementation tbh but more than twice as fast in most cases and <1s

## Reviewers

- Primary: @mariogiampieri @nofurtherinformation 
- Secondary: @districtr/districtr-reboot 

## Checklist

### DIFF
--- a/backend/app/alembic/versions/b04f5bed1e48_faster_connected_component_bboxes.py
+++ b/backend/app/alembic/versions/b04f5bed1e48_faster_connected_component_bboxes.py
@@ -33,17 +33,6 @@ def upgrade() -> None:
             sql = f.read()
         op.execute(sa.text(sql))
 
-    get_gerrydb_tables = sa.text(
-        "SELECT table_name FROM INFORMATION_SCHEMA.TABLES WHERE table_schema = 'gerrydb'"
-    )
-    conn = op.get_bind()
-    gerrydb_tables = conn.execute(get_gerrydb_tables).scalars().all()
-
-    for table_name in gerrydb_tables:
-        op.execute(
-            sa.text(f"CREATE INDEX ON gerrydb.{table_name} USING GIST (geometry)")
-        )
-
 
 def downgrade() -> None:
     op.execute(
--- a/backend/app/alembic/versions/b04f5bed1e48_faster_connected_component_bboxes.py
+++ b/backend/app/alembic/versions/b04f5bed1e48_faster_connected_component_bboxes.py
@@ -33,9 +33,21 @@ def upgrade() -> None:
             sql = f.read()
         op.execute(sa.text(sql))
 
+    get_gerrydb_tables = sa.text(
+        "SELECT table_name FROM INFORMATION_SCHEMA.TABLES WHERE table_schema = 'gerrydb'"
+    )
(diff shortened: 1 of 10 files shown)


backend/app/sql/get_block_zone_assignments_bboxes.sql
```diff
@@ -0,0 +1,33 @@
...
+    WHERE document.document_id = $1;
+
+    IF doc_districtrmap.districtr_map_slug IS NULL THEN
+        RAISE EXCEPTION 'Table name not found for document_id: %', $1;
+    END IF;
+
+    IF doc_districtrmap.child_layer IS NULL THEN
+        RAISE EXCEPTION 'Child layer is NULL for document_id: %. Block-level exports are not supported', $1;
+    ELSE
+        sql_query := format('
```
nofurtherinformation, line 20: H: This query doesn't filter zones current, which could improve performance. Locally, I'm seeing ~2x gains by adding the `WHERE` clauses here.

```suggestion
 sql_query := format($f$
            SELECT
                assignments.geo_id,
                assignments.zone,
                Box2D(blocks.geometry) AS bbox
            FROM get_block_assignments($1::UUID, $2) assignments
            JOIN gerrydb.%I blocks
            ON blocks.path = assignments.geo_id
            WHERE assignments.zone IS NOT NULL
            AND assignments.zone = ANY($2)
        $f$, doc_districtrmap.child_layer);
```

backend/app/alembic/versions/b04f5bed1e48_faster_connected_component_bboxes.py
```diff
@@ -0,0 +1,41 @@
...
+
+
+# revision identifiers, used by Alembic.
+revision: str = "b04f5bed1e48"
+down_revision: Union[str, None] = "518ab28c5fd6"
+branch_labels: Union[str, Sequence[str], None] = None
+depends_on: Union[str, Sequence[str], None] = None
+
+
+def upgrade() -> None:
```
nofurtherinformation, line 23: H: We should add spatial indices to the block tables. Eg.

```
CREATE INDEX ON gerrydb.{child_layer} USING GIST (geometry);
```

nofurtherinformation (changes requested):
Overall looks very good! I tooled around a bit locally and with adding a spatial index and refining the query a bit and (unscientifically) I'm seeing a 2-5x speed increase. 

The spatial index costs disk space, but I think that is a small price to pay. Try out the tweak on the query and spatial index and let me know what you think -- happy to merge as-is too

backend/app/sql/get_block_zone_assignments_bboxes.sql
raphaellaude: nice

PR #329, opened 2025-04-11
Better zone number labels

nofurtherinformation:
This PR improves the zone label positioning.

## Description
- Previous work to position zone labels used geometric dissolves / unions and centroids, which was slow (2000-3000ms), but pretty good
- To optimize this, we started using random points (~20-100ms), but that was kind of annoying when points on or near the edges were selected
- This approach instead compromises by making a low res raster of the map and choosing a valid point based on the raster image; this is pretty much as good as the first method, but much faster (~30-60ms)

## Reviewers
- Primary: @raphaellaude 
- Secondary:

## Checklist
- [x] Add raster shenanigans to the web worker
- [x] Fallback to random point
- [x] Remove dead code

### DIFF
--- a/app/src/app/components/Cms/ContentEditor/ContentEditor.tsx
+++ b/app/src/app/components/Cms/ContentEditor/ContentEditor.tsx
@@ -1,4 +1,4 @@
-import React, { useEffect, useState } from 'react';
+import React, {useEffect, useState} from 'react';
 import {useCmsFormStore} from '@/app/store/cmsFormStore';
 import {LANG_MAPPING} from '@/app/utils/language';
 import {Box, Button, Flex, Grid, Heading, Select, Text, TextField} from '@radix-ui/themes';
@@ -20,21 +20,29 @@ export const ContentEditor: React.FC = () => {
   const [contentHasChanged, setContentHasChanged] = useState(false);
 
   useEffect(() => {
-    if (!editingContent && formData?.content.slug && formData?.content.language && formData?.content.title) {
+    if (
+      !editingContent &&
+      formData?.content.slug &&
+      formData?.content.language &&
+      formData?.content.title
+    ) {
       setContentHasChanged(true);
     } else if (editingContent) {
-      const currentEditContent = editingContent.content.draft_content || editingContent.content.published_content;
+      const currentEditContent =
+        editingContent.content.draft_content || editingContent.content.published_content;
       if (!currentEditContent) {
         setContentHasChanged(false);
       } else {
-        const contentChanged = JSON.stringify({title: currentEditContent.title, body: currentEditContent.body}) !== JSON.stringify({title: formData?.content.title, body: formData?.content.body});
+        const contentChanged =
+          JSON.stringify({title: currentEditContent.title, body: currentEditContent.body}) !==
+          JSON.stringify({title: formData?.content.title, body: formData?.content.body});
         setContentHasChanged(contentChanged);
       }
     } else {
       setContentHasChanged(false);
     }
   }, [editingContent, formData]);
-  
+
   return (
     <Flex direction="column" gapY="4" p="6" className="bg-white shadow rounded-lg">
       <Heading as="h2" className="text-xl font-semibold">
--- a/app/src/app/utils/GeometryWorker/geometryWorker.ts
+++ b/app/src/app/utils/GeometryWorker/geometryWorker.ts
@@ -183,16 +183,11 @@ const GeometryWorker: GeometryWorkerClass = {
       bboxGeom,
     };
   },
-  async computeCenterOfMass(
-    geojson,
-    bounds,
-    width = 256,
-    height = 256
-  ){
+  async computeCenterOfMass(geojson, bounds, width = 256, height = 256) {
     const canvas = new OffscreenCanvas(width, height);
     const ctx = canvas.getContext('2d', {willReadFrequently: true});
     if (!ctx) return null;
-  
+
     // Calculate bounds for rendering
     const [minX, minY, maxX, maxY] = bounds;
     const scaleX = width / (maxX - minX);
@@ -201,13 +196,13 @@ const GeometryWorker: GeometryWorkerClass = {
     ctx.fillStyle = `rgb(255,0,0)`;
     ctx.strokeStyle = `rgb(255,0,0)`;
     ctx.lineWidth = 1;
-  
+
     for (const feature of geojson.features) {
       const geom = feature.geometry;
       const coords = getCoords(feature);
       const polygons: GeoJSON.Polygon['coordinates'] =
         geom.type === 'MultiPolygon' ? coords.flat() : coords;
-  
+
       for (const ring of polygons) {
         ctx.beginPath();
         ring.forEach(([x, y], i) => {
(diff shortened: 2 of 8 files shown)


app/src/app/components/Cms/ContentEditor/ContentEditor.tsx
```diff
@@ -1,4 +1,4 @@
-import React, { useEffect, useState } from 'react';
```
raphaellaude, line 1: What do the changes in this file have to do with the better zone number labels?

raphaellaude: I'd move them to a separate PR

app/src/app/utils/GeometryWorker/geometryWorker.ts
```diff
@@ -238,35 +183,137 @@ const GeometryWorker: GeometryWorkerClass = {
...
     };
   },
-  getCentersOfMass(bounds, activeZones) {
+  async computeCenterOfMass(
+    geojson,
+    bounds,
+    width = 256,
+    height = 256
+  ){
+    const canvas = new OffscreenCanvas(width, height);
```
raphaellaude, line 192: Nice use of `OffscreenCanvas`! Browser compatibility seems good enough ([mozilla](https://developer.mozilla.org/en-US/docs/Web/API/OffscreenCanvas#browser_compatibility), [canIuse](https://caniuse.com/offscreencanvas))

app/src/app/utils/GeometryWorker/geometryWorker.ts
```diff
@@ -238,35 +183,137 @@ const GeometryWorker: GeometryWorkerClass = {
...
+    const ctx = canvas.getContext('2d', {willReadFrequently: true});
+  
+    // Calculate bounds for rendering
+    const [minX, minY, maxX, maxY] = bounds;
+    const scaleX = width / (maxX - minX);
+    const scaleY = height / (maxY - minY);
+    if (!ctx) return null;
```
raphaellaude, line 199: Q/nit: any reason you call this after doing some work and not right after getting the ctx a few lines up?

app/src/app/utils/GeometryWorker/geometryWorker.ts
```diff
@@ -238,35 +183,137 @@ const GeometryWorker: GeometryWorkerClass = {
...
+      }
+    }
+    const imageData = ctx.getImageData(0, 0, width, height).data;
+    let sumX = 0,
+      sumY = 0,
+      count = 0;
+    const validPixels = [];
+    for (let y = 0; y < height; y++) {
+      for (let x = 0; x < width; x++) {
+        const i = (y * width + x) * 4;
```
raphaellaude, line 231: Q: What is the * 4 for?

app/src/app/utils/GeometryWorker/geometryWorker.ts
```diff
@@ -238,35 +183,137 @@ const GeometryWorker: GeometryWorkerClass = {
...
+          count++;
+          validPixels.push({x, y});
+        }
+      }
+    }
+  
+    if (count === 0) return null;
+  
+    let centerX = sumX / count;
+    let centerY = sumY / count;
```
raphaellaude, line 244: Super smart

app/src/app/utils/GeometryWorker/geometryWorker.ts
```diff
@@ -238,35 +183,137 @@ const GeometryWorker: GeometryWorkerClass = {
...
+  
+    const idx = (Math.floor(centerY) * width + Math.floor(centerX)) * 4;
+    const isValidCenter = imageData[idx] === 255;
+  
+    // Fallback: choose a random pixel inside the district
+    if (!isValidCenter) {
+      const fallback = validPixels[Math.floor(Math.random() * validPixels.length)];
+      centerX = fallback.x;
+      centerY = fallback.y;
+    }
```
raphaellaude, line 254: The numbers jump around a lot when no valid centers are found on each rerender, which is a bit distracting. I'm wondering if there's a more deterministic way of doing this. E.g. pick the median valid pixel? `Math.floor(validPixels.length / 2)`

image

app/src/app/utils/GeometryWorker/geometryWorker.ts
```diff
@@ -343,14 +390,19 @@ const GeometryWorker: GeometryWorkerClass = {
...
+  async getCentroidsFromView({
+    bounds,
+    activeZones,
+    strategy = 'non-colliding-centroids',
+    minBuffer,
+  }) {
     switch (strategy) {
       case 'center-of-mass':
-        return this.getCentersOfMass(bounds, activeZones);
+        return await this.getCentersOfMass(bounds, activeZones);
```
raphaellaude, line 401: Q: Since you've exposed the width and height as part of the function, is there any sense in using those params? Or is the 256 square canvas just fine?

raphaellaude (approved):
Definitely a huge improvement. Very smartly done. I have a couple comments / suggested improvements but not a big deal if we want to merge

app/src/app/components/Cms/ContentEditor/ContentEditor.tsx
nofurtherinformation: These just missed prettier in the last PR -- we can split them out if you'd prefer!

app/src/app/utils/GeometryWorker/geometryWorker.ts
nofurtherinformation: thank you!

app/src/app/utils/GeometryWorker/geometryWorker.ts
nofurtherinformation: Good catch! Resolved in 7816d3f0e2de0a36a04becd67b995c2b1207b6b6

app/src/app/utils/GeometryWorker/geometryWorker.ts
nofurtherinformation: Each pixels gets 4 positional int values (R G B A) so this multiplier gets us to the first channel -- I added some clarifying comments in 6d56fe596e663452fafaea3420aa18f136a1c4cf  !

app/src/app/utils/GeometryWorker/geometryWorker.ts
nofurtherinformation: ty!!

app/src/app/utils/GeometryWorker/geometryWorker.ts
nofurtherinformation: Great catch, agreed. 

I added a slight variation on this suggestion in 357b073778aab9e4ba9675b90e6d236dad055434, which checks if any of the adjacent pixels to the geometric center work, and if not falls back to the median

Most of the cases where the center fails is due to geometry imprecision or jagged boundaries and such, or it is actually a circular / irregular shape

app/src/app/utils/GeometryWorker/geometryWorker.ts
nofurtherinformation: Added in 8ecc3a5f6975975d0b6c5a9675fa8f978760bc42

nofurtherinformation:
Thanks very much for this review @raphaellaude ! I've added some clean up and clarifications, please re-review at your convenience!

raphaellaude (approved):
lgtm

PR #330, opened 2025-04-12
Assorted UI fixes to map

mapmeld:
1. Undo and Redo tooltips should appear separately

Screenshot 2025-04-12 at 4 55 45 PM

2. I don't know if this is better fixed on the DB side, but the Alaska map should not zoom out to the whole world. I wasn't able to test this locally

3. As discussed on Slack, it's hard to get some small blocks on the WV map, I believe this is because we deactivate painting while the cursor/center is outside the interactive blocks, so you could be swiping across the map and not getting as good coverage on the state border. This event is still useful to remove highlighting after your mouse fully leaves the state / map / window

### DIFF
--- a/app/src/app/components/Map/Map.tsx
+++ b/app/src/app/components/Map/Map.tsx
@@ -1,5 +1,5 @@
 'use client';
-import type {LngLatBoundsLike, MapLayerEventType} from 'maplibre-gl';
+import type {MapLayerEventType} from 'maplibre-gl';
 import maplibregl from 'maplibre-gl';
 import 'maplibre-gl/dist/maplibre-gl.css';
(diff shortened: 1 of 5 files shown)


nofurtherinformation:
Thanks for these fixes @mapmeld !

Great catch on #1. 

For #2, this is on the database side when we generate the DistrictrMap views -- we should fix this there and avoid one-off fixes in the components. 

For #3, the issue is the small geometries do not render when zoomed out. I think we should keep the current painting behavior. If we don't interrupt the painting event when it leaves the canvas, it can create the situation where the brush is stuck on

nofurtherinformation (approved):
lgtm! thank you!

PR #332, opened 2025-04-16
Contiguity tool loads custom colorScheme

### LINKED ISSUE #313: Custom colors aren't applied to contig
Image

mapmeld:
Fixes #313
Loads colorScheme from mapStore to allow custom colors

I did a `git grep` to check if other files still use `colorScheme` from `constants/colors `, and others already have been updated to `import {colorScheme as DefaultColorScheme}`


raphaellaude:
image

This change only seems to apply to the first swatch

mapmeld:
I don't have contiguity environment to test it locally. I'm thinking that the square could be set to an invalid color, but I don't know how. If the problem was that `colorScheme` was blank or wasn't updated then I don't know how it could update one district and not the next one

mapmeld:
ok, contiguity should be colorful again 👍 
issue here is `(row.zone % colorScheme.length) - 1` wasn't a problem with the long list of colors, but after shortening in the custom colors tool, the last district would go back to a -1 index. With the order of % and - switched it should be resolved

raphaellaude (approved):
lgtm

PR #333, opened 2025-04-16
Use maps metadata in title/description

mapmeld:
This replaces NextJS's static metadata feature with its `generateMetadata` async function, allowing the title and description of the page and its social media / opengraph preview to use a map's name, description, and number of districts.

- I suggest the Districtr paintbrush logo as the social media image until we add #314 to make individual thumbnail URLs
- Replaces description "Districtr reboot" with "Create districting maps"
- On OpenGraph where we already have siteName "Districtr 2.0" and the Districtr logo, the title can be "Get Started" on general pages like the homepage or /map
- On individual maps the title would be "3 districts - NAME" or "3 districts - Shared Map"
- On individual maps the description would be from map metadata or "Create districting maps"

Notes:
- removes a duplicate type definition for `MapDocument`
- very open to reworking how we use name / num_districts and other phrasing, this is mostly to make better metadata
- uses `num_districts` and not the number of districts drawn by the user
- I didn't decide how to get the state name from layer names

### DIFF
--- a/app/src/app/layout.tsx
+++ b/app/src/app/layout.tsx
@@ -4,7 +4,6 @@ import './globals.css';
 import {Theme} from '@radix-ui/themes';
 import '@radix-ui/themes/styles.css';
 import {getDocument} from '@/app/utils/api/apiHandlers/getDocument';
-import {isArray} from 'lodash';
 
 const inter = Inter({subsets: ['latin']});
 
@@ -23,7 +22,7 @@ const DISTRICTR_LOGO = [
 export async function generateMetadata({searchParams}: MetadataProps): Promise<Metadata> {
   const resolvedParams = await searchParams;
   const document_id = resolvedParams?.document_id ?? '';
-  const singularDocumentId = isArray(document_id) ? document_id[0] : document_id;
+  const singularDocumentId = Array.isArray(document_id) ? document_id[0] : document_id;
   if (!singularDocumentId) {
     return {
       title: 'Districtr 2.0',
--- a/app/src/app/layout.tsx
+++ b/app/src/app/layout.tsx
@@ -3,13 +3,54 @@ import {Inter} from 'next/font/google';
 import './globals.css';
 import {Theme} from '@radix-ui/themes';
 import '@radix-ui/themes/styles.css';
+import {getDocument} from '@/app/utils/api/apiHandlers/getDocument';
+import {isArray} from 'lodash';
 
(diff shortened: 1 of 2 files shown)


app/src/app/layout.tsx
```diff
@@ -3,14 +3,55 @@ import {Inter} from 'next/font/google';
...
+    url: 'https://districtr-v2-frontend.fly.dev/_next/image?url=%2Fdistrictr_logo.jpg&w=1920&q=75',
+    width: 1136,
+    height: 423,
+  },
+];
+
+export async function generateMetadata({searchParams}: MetadataProps): Promise<Metadata> {
+  const resolvedParams = await searchParams;
+  const document_id = resolvedParams?.document_id ?? '';
+  const singularDocumentId = isArray(document_id) ? document_id[0] : document_id;
```
nofurtherinformation, line 26: Nit: I'd like to phase out the lodash dependency at some point, `Array.isArray` should do the trick here

app/src/app/utils/api/apiHandlers/types.ts
```diff
@@ -1,20 +1,6 @@
 import {NullableZone} from '@constants/types';
 import {SummaryTypes} from '../summaryStats';
 
-export interface DocumentObject {
```
nofurtherinformation, line 4: Nice catch -- thank you!

nofurtherinformation (changes requested):
Nice work! One small nit but otherwise I think this is a great starting point for SSR page metadata

app/src/app/layout.tsx
mapmeld: Makes sense 👍

nofurtherinformation (approved):
lgtm!

mapmeld:
Update on the OpenGraph tags: this is in layout.tsx so we don't have access to document_id ( https://nextjs.org/docs/app/api-reference/file-conventions/layout#layouts-do-not-receive-searchparams ) and it is always showing the Getting Started text.
Also the Districtr paintbrush logo is cropped to lose the D https://www.opengraph.xyz/url/https%3A%2F%2Fdistrictr-v2-frontend.fly.dev

The function can also go into page.tsx, but that wouldn't work in `(interactive)map/page.tsx` because it has client-side components / `use client`
Also since this is happening on the server then I'm unclear how it will work with the share links

I'll look up whether other sites use the ID in the URL, or organize page.tsx differently, but unfortunately having a bit of a setback

PR #334, opened 2025-04-17
Parquet tabular data

nofurtherinformation:
Draft: refactor demography and tabular data to use parquet instead of postgres

## Description
- Refactor pipelines to be a bit more ergonomic + add dockerization
- Add pipeline to build parquet files that can be partially fetched
- Frontend updates to handle data sets that may more partial and incomplete, like elections
- Add some more customizability in how we express choropleth variables 

The new format is described below and this description is also in the readme under `pipelines`.

> Tabular data is expected to be output as parquet files withthe following requirements:
> - Parquet, long format with these columns: `path` (geoid), `column_name` (metric in that row), and `value`
> - Compressed (default ZSTD level 12)
> - Rows ordered by parent geography (VTD). The first rows must be 
> - Key value metadata with the following information:
>   - `column_list`: String list of all available columns in the long format
>   - `length_list`: A dictionary of parent IDs and row ranges as an array `[number,number]`. Must include `parent`, which has the base parent data.
> 
> _Why long?_: Long format is much better for partial reads of the parquet, and with compression ends up a bit more efficient in transfer size
>
> _Does it stay long format in browser?_: Nope! The queries and column derives are faster as wide. So, the parquert gets partially read using `hyparquet` on a WebWorker thread which rotates the data to a wide format.
>
> _Doesn't this key-value metadata get chunky?_: A little. Texas, as an example of a big state, gets to be around 300kb. It's not ideal, but not terrible.

## Reviewers
- Primary: @raphaellaude 
- Secondary:

## Checklist
- [x] Rough out
- [x] Add dockerization to pipeline module
- [x] Refactor common pipeline code to `core` module
- [x] Add tabular pipeline module
- [x] Run pipeline to generate data
- [x] FE - Add parquet worker thread that uses Hyparquet to partially read data
- [x] FE - refactor to allow for per-column available datasets
- [x] FE - some type cleanup and simplification reflecting the above
- [x] FE - add new expressions (anonymous functions to derive viz columns) and fixed ranges 
- [x] FE - added linear scale legend for custom scales
- [ ] Add pipeline tests?
Screenshot 2025-04-18 at 12 39 28 PM


pipelines/README.md
```diff
@@ -6,4 +6,39 @@ Most Input datasets come from the gerrydb project.
...
+
+### Tabular
+
+Tabular data is expected to be output as parquet files withthe following requirements:
+- Parquet, long format with these columns: `path` (geoid), `column_name` (metric in that row), and `value`
+- Compressed (default ZSTD level 12)
+- Rows ordered by parent geography (VTD). The first rows must be 
+- Key value metadata with the following information:
+  - `column_list`: String list of all available columns in the long format
+  - `length_list`: A dictionary of parent IDs and row ranges as an array `[number,number]`. Must include `parent`, which has the base parent data.
```
raphaellaude, line 39: H: I would see if we can rely on the default parquet metadata if you haven't already. Things like column names are [by default included in the parquet metadata](https://duckdb.org/docs/stable/data/parquet/metadata.html#parquet-schema). Same goes for the length list; [column page stats](https://github.com/apache/parquet-format/blob/master/PageIndex.md) are by default implemented in `1.11.0`. I worry by using our own versions of this metadata we are leaking information from our pipeline into the FE. This coupling could come back to bite!

app/src/app/utils/ParquetWorker/parquetWorker.ts
```diff
@@ -0,0 +1,138 @@
...
+    }
+    const url = `${process.env.NEXT_PUBLIC_S3_BUCKET_URL}/tabular/${view}.parquet`;
+    const byteLength = await byteLengthFromUrl(url).then(Number);
+    const file = await asyncBufferFromUrl({url, byteLength});
+    let metadata = (await parquetMetadataAsync(file)) as ExtendedFileMetaData;
+    metadata['rows'] = JSON.parse(
+      metadata['key_value_metadata']
+        ?.find(f => f.key === 'length_list')
+        ?.value?.replaceAll("'", '"') ?? '{}'
+    );
```
raphaellaude, line 22: would love to see if we can get around having to do this by using the default parquet metadata

app/src/app/store/demography/constants.ts
```diff
@@ -0,0 +1,113 @@
...
+  'Attorney General 2022': ['ag_22_dem', 'ag_22_rep'],
+  'Attorney General 2018': ['ag_18_dem', 'ag_18_rep'],
+  'Governor 2022': ['gov_22_dem', 'gov_22_rep'],
+  'Governor 2018': ['gov_18_dem', 'gov_18_rep'],
+  'Senate 2022': ['sen_22_dem', 'sen_22_rep'],
+  'Senate 2018': ['sen_18_dem', 'sen_18_rep'],
+  'Senate 2016': ['sen_16_dem', 'sen_16_rep'],
+  'Presidential 2020': ['pres_20_dem', 'pres_20_rep'],
+  'Presidential 2016': ['pres_16_dem', 'pres_16_rep'],
+} as const;
```
raphaellaude, line 24: Nice. So much better to be able to define these groupings in the FE.

raphaellaude (changes requested):
Lots of great progress. It's working well in the FE but I have a couple of high level comments I'd love to discuss:

### 1. Pipeline -> FE coupling

We're leaking information from the pipeline to the FE in the parquet metadata (the `columns` and `length_list` KVs). I would see if we can get away with using only the default parquet metadata rather than the KV store. The current choice ties us to a very particular implementation that could be hard to back out of.

### 2. Pipeline package design

Now that we're making the pipelines package a bit more full featured, now is the time to think a bit bigger about the package design (generalizing / interfaces / CLI).

I'm thinking that some of the create / replace / upload logic is going to be pretty repetitive across the different module pipelines. We might want to settle on a standard interface each module can implement; e.g. if all pipeline modules implement create and upload methods, then a shared `Manager` dependency can manage config reading, batch uploads, and replace logic.

### 3. API deprecation

We should clean up the API, deleting the views and population totals logic.

### 4. API views

Currently we're defining views in the FE, which seems quite flexible / easy. It does however tie the views to a deploy. It seems like data updates requiring FE changes was a pain point of Districtr v1. We might want to consider these the default views but implement a new way of defining views in the API/db so that users can define their own views. This can happen (if at all) in a separate PR after discussion w/ Moon.

app/src/app/utils/ParquetWorker/parquetWorker.ts
nofurtherinformation: Done in ce14b0c10fac3f8efd79d45313ed089fd532d2e0!

pipelines/configs/tabular_vap_and_election.yml
```diff
@@ -0,0 +1,43 @@
+name: vap and election batch dataset load
```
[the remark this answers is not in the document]
nofurtherinformation: Resolved in 847ca627de2bfac293e6b97dcd15d419e80456ae

pipelines/README.md
```diff
@@ -6,4 +6,39 @@ Most Input datasets come from the gerrydb project.
 
 Structure:
 
+- `core`: Shared configuration and models
+- `configs`: YAML configs for batch scripts
```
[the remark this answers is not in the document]
nofurtherinformation: Resolved in 847ca627de2bfac293e6b97dcd15d419e80456ae

pipelines/__main__.py
```diff
@@ -0,0 +1,25 @@
+"""
+Main entry point for the pipelines package.
```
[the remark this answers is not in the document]
nofurtherinformation: Agreed, I've simplified this in a3731191f98d61bcda375af1ce851052b57c9095 and we just have one core CLI that gets imported and we add click groups

app/src/app/store/demography/constants.ts
nofurtherinformation: Thank you!! I've formalized the types a bit and noted the next steps needed in this issue: https://github.com/districtr/districtr-v2/issues/337

nofurtherinformation:
@raphaellaude This is pretty much ready for re-review. 

Key updates since the last review:
1. Swapped to standard parquet page index / row statistics metadata instead of custom KV metadata
2. Simplified the pipelines CLI (although not fully abstracted as is needed)
3. Tore out old tabular data functions and API and created migration to drop 
4. Added voter data to evaluation panel (and some UI tweaks based on feedback)

raphaellaude (approved):
This looks great! Nice job. I added a couple of commits to finish up the API and pipeline clean-up:
- Removed `/total_pop` endpoint and `ZonePopulation` model it used
- I hope you don't mind but I didn't think the pipelines package needed to be an executable so I simplified things by moving everything to a single `python cli.py ...` entrypoint

PR #335, opened 2025-04-18
Bugfix: load data

nofurtherinformation:
From a fresh install, the load data script had a couple breaking errors:

1. When making gerrydb views, the views not committed before building which cause the geo index to fail and mess up other stuff down the line
2. When makign the districtrmap view, the view was not committed before trying to make parent child edges


mapmeld (approved):
✅ lgtm , it looks like these would resolve the issues that I saw

PR #336, opened 2025-04-23
Create individual map metadata on page.tsx

mapmeld:
After #333, I found that layout.tsx doesn't have access to the query parameters (including our `document_id=`) in the serverside rendering code which generates metadata tags. https://nextjs.org/docs/app/api-reference/file-conventions/layout#layouts-do-not-receive-searchparams

To display a title, description, and eventually thumbnail for individual maps, that needs to happen in a serverside version of `page.tsx`.
The PR accomplishes this by moving the existing page code to `component.tsx`, and having the `page.tsx` just be a wrapper around that component.
I also made a change to help the Docker container get the API_URL

Ongoing issue with `getDocument`: the existing function won't run on a server because it uses `useMapStore` to get the current user ID, and passes that to the server. I don't know the context where we wanted to do that, can we get user ID from cookies and sessions?


nofurtherinformation (approved):
I made two small tweaks and merged in main for sanity, but I think this looks good to go!

PR #340, opened 2025-04-25
Save share reprised

mariogiampieri:
Reprises some save and share behavior. 
- [x] change map status from a button to a badge
- [x] add separate map status, "Scratch Work Only" and make the default condition
- [x] split active map modal from recent maps;
- [x] active map properties available from button "Save/Status" on menu bar
- [x] update UI on Save/Status modal
- [x] Make plan name + module show up in top bar (e.g. **My cool map** _MA Congressional Map using 2020 census geographies_)
    - [x] map name
    - [x] module name - need to be sure to get human readable version
- [x] Share link is not right anymore - just goes to home page. Missing the “/map” part of the url 
- [x] Share editable requires a password. Button should not be clickable without it 
- [x] “Copy Link” instead of “Generate Link” and 
- [ ] link should be visible when it’s visible- need to ask Dylan plan on link shortening 
- [x] zone changes are not checking if the plan is locked- it should check if it’s locked first, and if it is locked, to not transact 
- [x] Make the name clickable to open the save 
- [x] all maps, including what is loaded, should be available in recent 
- [x] there should be no default status and instead it should be blank 
- [x] current map should show up in recent maps, right now it does not. In recent table it should say - 
- [x] recent map- “Active” and “Ready to Load” instead of just “active”? << this
- [x] “remove from list” instead of “delete” - and make a modal that makes it actually prompt you
	- [x] delete should actually delete, it’s not properly removing entries from table

## Description

- Item 1

## Reviewers

- Primary:
- Secondary:

## Checklist

### DIFF
--- a/backend/tests/test_thumbnails.py
+++ b/backend/tests/test_thumbnails.py
@@ -1,6 +1,6 @@
 import os
 import pytest
-from tests.constants import FIXTURES_PATH
+from tests.constants import FIXTURES_PATH, USER_ID
 from unittest.mock import patch
 
 
@@ -30,7 +30,9 @@ def document_id_with_assignments(client, document_id):
                     "geo_id": "200979691001108",
                     "zone": 2,
                 },
-            ]
+            ],
+            "updated_at": "2023-10-01T00:00:00Z",
+            "user_id": USER_ID,
         },
     )
     assert response.status_code == 200


backend/app/alembic/versions/2f8e63322e06_merge_stats_metadata.py
```diff
@@ -0,0 +1,24 @@
+"""merges update_stats and metadata revisions
```
nofurtherinformation, line 1: Q: Is this revision needed?

backend/app/main.py
```diff
@@ -312,19 +311,33 @@ async def create_document(
...
+        if len(assignments) > 0:
+            updated_at = update_timestamp(session, document_id)
+        session.commit()
+        return {
+            "assignments_upserted": len(assignments),
+            "updated_at": data["updated_at"],
+        }
+    else:
+        raise HTTPException(
+            status_code=status.HTTP_403_FORBIDDEN,
```
nofurtherinformation, line 338: Low: `409 Conflict` Might be better

app/src/app/components/Toolbar/RecentMapsModal.tsx
```diff
@@ -174,22 +167,62 @@ const RecentMapsRow: React.FC<{
...
-        {!active && (
+      <Table.Cell py=".5rem" justify="center">
+        {!active ? (
           <Button
             onClick={() => onSelect(data)}
             variant="outline"
             className="box-content size-full rounded-xl hover:bg-blue-200 inline-flex transition-colors"
           >
-            Load
+            Ready to Load
```
nofurtherinformation, line 177: Q: Maybe just 'Load' -- is it ever not ready to load in this state?

app/src/app/components/Toolbar/RecentMapsModal.tsx
```diff
@@ -174,22 +167,62 @@ const RecentMapsRow: React.FC<{
...
+              className="box-content mx-2 size-full rounded-xl hover:bg-blue-200 inline-flex transition-colors"
+            >
+              Active
+            </Button>
+            <Button
+              disabled
+              variant="outline"
+              className="box-content mx-2 size-full rounded-xl hover:bg-blue-200 inline-flex transition-colors"
+            >
+              Unload
```
nofurtherinformation, line 193: Q: What does / should this do?

app/src/app/components/Toolbar/RecentMapsModal.tsx
```diff
@@ -174,22 +167,62 @@ const RecentMapsRow: React.FC<{
...
             className="box-content size-full rounded-xl hover:bg-blue-200 inline-flex transition-colors"
           >
-            Load
+            Ready to Load
           </Button>
+        ) : (
+          <div style={{}}>
+            <Button
+              disabled
+              variant="outline"
```
nofurtherinformation, line 183: Medium: ghost instead of outline to indicate no interaction
```suggestion
              variant="ghost"
```

app/src/app/components/Toolbar/SaveMapModal.tsx
```diff
@@ -0,0 +1,309 @@
...
+  const mapDocument = useMapStore(store => store.mapDocument);
+  const status = useMapStore(store => store.mapStatus?.status);
+  const userMaps = useMapStore(store => store.userMaps);
+  const upsertUserMap = useMapStore(store => store.upsertUserMap);
+  const [mapNameIsSaved, setMapNameIsSaved] = React.useState(false);
+  const [groupNameIsSaved, setGroupNameIsSaved] = React.useState(false);
+  const [tagsIsSaved, setTagsIsSaved] = React.useState(false);
+  const [descriptionIsSaved, setDescriptionIsSaved] = React.useState(false);
+  const [shareStateIsSaved, setShareStateIsSaved] = React.useState(false);
+  const [password, setPassword] = React.useState<string | null>(null);
```
nofurtherinformation, line 53: H: Reset relevant state on map document switch?

nofurtherinformation (changes requested):
Overall code looks solid. Some Q's and also here are some UAT comments from Chris:

> when i accessed a frozenurl of my map, i had to reload the page once to get the paint to load (but the pop bar was correct on first load)
> when i made a copy of my frozen map i had to reload the page to see the updated title at the top
> after sharing my map as frozen, i tried to share as editable but the link always shared the frozen map
> password persisted to other maps, ie switching from FL to CO had my same password in the share box. i think it should reset to empty.
> even when starting from fresh map module, share with password only shares as frozen

app/src/app/components/Toolbar/SaveMapModal.tsx
mariogiampieri: good catch- handled now in added `useEffect` below

app/src/app/components/Toolbar/RecentMapsModal.tsx
mariogiampieri: yes good call

app/src/app/components/Toolbar/RecentMapsModal.tsx
mariogiampieri: actually plugged in now and unloads a document!

app/src/app/components/Toolbar/RecentMapsModal.tsx
```diff
@@ -1,15 +1,15 @@
...
+import {Button, Flex, Text, Table, Dialog, Box, Separator, Popover} from '@radix-ui/themes';
 import {usePathname, useSearchParams, useRouter} from 'next/navigation';
 import {DocumentObject} from '@utils/api/apiHandlers/types';
 import {styled} from '@stitches/react';
 import {useTemporalStore} from '@/app/store/temporalStore';
+import {unlockMapDocument} from '@/app/utils/api/apiHandlers/unlockMapDocument';
 
 const DialogContentContainer = styled(Dialog.Content, {
-  maxWidth: '75vw',
+  maxWidth: '60vw',
```
nofurtherinformation, line 12: PP: On mobile increase this to 90-95vw?

app/src/app/components/Toolbar/SaveMapModal.tsx
```diff
@@ -0,0 +1,327 @@
...
+  gap: '1rem',
+});
+
+const DialogContentContainer = styled(Dialog.Content, {
+  maxHeight: 'calc(100vh-2rem)',
+  maxWidth: '60vw',
+  overflowY: 'auto',
+});
+
+export const SaveMapModal: React.FC<{
```
nofurtherinformation, line 34: Low: At some point in the future we may want to refactor some of this state to a custom hook or zustand store

backend/requirements.txt
```diff
@@ -51,7 +51,7 @@ filelock==3.17.0
     # via virtualenv
 fiona==1.10.1
 geoalchemy2==0.15.2
-geopandas==1.0.1
+geopandas==0.14.4
```
nofurtherinformation, line 54: Q: Is this a local env issue ?

mariogiampieri: i did have an env issue and noticed there were different versions in `backend` and `pipelines` and set both to 0.14, would be just as happy embracing 1.*

app/src/app/components/Toolbar/RecentMapsModal.tsx
mariogiampieri: good catch, implemented on save modal also

backend/requirements.txt
nofurtherinformation: Lets do 1.* -- I'm not sure how the thumbnail functions will work with 0.14.4 instead and would rather break the pipelines :)

nofurtherinformation (approved):
lgtm -- nice work!

backend/app/alembic/versions/2f8e63322e06_merge_stats_metadata.py
raphaellaude: err probably not since it's not doing anything! let's clean this up...

backend/app/main.py
```diff
@@ -313,19 +313,33 @@ async def create_document(
 
 @app.patch("/api/update_assignments")
 async def update_assignments(
-    data: AssignmentsCreate, session: Session = Depends(get_session)
+    data: dict = Body(...), session: Session = Depends(get_session)
```
raphaellaude, line 316: H: no validation on input. Why not add `user_id` to `AssignmentsCreate`?

backend/app/models.py
```diff
@@ -141,13 +141,20 @@ class ParentChildEdges(TimeStampMixin, SQLModel, table=True):
...
+
+
 class DistrictrMapMetadata(BaseModel):
     name: Optional[str] | None = None
     group: Optional[str] | None = None
     tags: Optional[list[str]] | None = None
     description: Optional[str] | None = None
     event_id: Optional[str] | None = None
-    is_draft: bool = True
+    draft_status: Optional[DocumentDraftStatus] = DocumentDraftStatus.scratch
```
raphaellaude, line 157: Nice. Love swapping a bool for an enum / strategy.

Just noticed these are all `Optional` and `Optional[T] | None` is redundant. Slight preference for repo convention of `T | None`

backend/app/sql/update_metadata_draft_status.sql
```diff
@@ -0,0 +1,20 @@
+-- migrations/sql/udf_update_metadata_draft_status.sql
+
+CREATE OR REPLACE FUNCTION update_metadata_draft_status(map_metadata JSONB)
```
raphaellaude, line 3: tbh this kind of json headache / shenanigans is why I've recommended we stick to fields in tables / stricter schema – especially since we're strictly typing the JSON anyway

raphaellaude (commented):
A few comments / thoughts for possible follow-up work.

I'd recommend we continue to strongly type the body for `/api/update_assignments` by adding `user_id` to `AssignmentsCreate` instead of loosely accepting `Body(...)`

PR #345, opened 2025-04-29
Maps can be organized into groups

mapmeld:
As discussed on the Slack, we want a MapGroup table to organize DistrictrMaps. The current homepage DistrictrMaps all fall into a States group.

Other maps will be browseable in the future by group - for example an organization wants to organize their DistrictrMaps across one or more states. With a many-to-many table, maps can be part of multiple groups.

I confirmed that `cli.py create-districtr-map` without any changes will put the map in the States group, and adding `--group-slug` put it somewhere else

Note:
- Auto-increment gets messed up if I INSERTing the first group row with a set ID in the migration. I could maybe set it to 0 or -1?
- Delete/cascade not being set in the migration - I wasn't sure the right way to make the join table dependent
- There is no warning if the `--group-slug` doesn't exist; the DistrictrMap is created without a row in the join table
- You can browse a group by ID, like `/api/gerrydb/views?group=10`, but this isn't readily accessible if we are using the slug
- Unclear how we will create groups (from CLI?)

### DIFF
--- a/backend/app/alembic/versions/dc0216fef023_create_map_groups.py
+++ b/backend/app/alembic/versions/dc0216fef023_create_map_groups.py
@@ -37,7 +37,7 @@ def upgrade() -> None:
             ["districtrmap.uuid"],
         ),
     )
-    op.execute(sa.text("INSERT INTO map_group (id, name, slug) VALUES (1, 'States', 'states')"))
+    op.execute(sa.text("INSERT INTO map_group (name, slug) VALUES ('States', 'states')"))
     op.execute(sa.text("INSERT INTO districtrmaps_to_groups (group_id, districtrmap_uuid) SELECT 1, uuid from districtrmap"))
 
 def downgrade() -> None:
--- a/backend/app/alembic/versions/dc0216fef023_create_map_groups.py
+++ b/backend/app/alembic/versions/dc0216fef023_create_map_groups.py
@@ -1,0 +1,45 @@
+"""create map groups
+
+Revision ID: dc0216fef023
+Revises: d38d0f766dc5
+Create Date: 2025-04-29 15:50:59.583224
+
+"""
+from typing import Sequence, Union
+
+from alembic import op
+import sqlalchemy as sa
+from sqlalchemy.dialects import postgresql
+
+# revision identifiers, used by Alembic.
+revision: str = 'dc0216fef023'
+down_revision: Union[str, None] = 'd38d0f766dc5'
+branch_labels: Union[str, Sequence[str], None] = None
+depends_on: Union[str, Sequence[str], None] = None
+
+
+def upgrade() -> None:
+    op.create_table('map_group',
+        sa.Column('id', sa.Integer(), primary_key=True, autoincrement=True),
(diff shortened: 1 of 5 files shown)


backend/app/main.py
```diff
@@ -952,11 +953,17 @@ async def update_districtrmap_metadata(
 async def get_projects(
     *,
     session: Session = Depends(get_session),
+    group: int = Query(default=1),
```
raphaellaude, line 956: H: I would use the group slug since it's a bit more human readable

backend/app/alembic/versions/dc0216fef023_create_map_groups.py
```diff
@@ -0,0 +1,45 @@
...
+branch_labels: Union[str, Sequence[str], None] = None
+depends_on: Union[str, Sequence[str], None] = None
+
+
+def upgrade() -> None:
+    op.create_table('map_group',
+        sa.Column('id', sa.Integer(), primary_key=True, autoincrement=True),
+        sa.Column('name', sa.String(), unique=True, nullable=False),
+        sa.Column('slug', sa.String(), unique=True, nullable=False),
+    )
```
raphaellaude, line 26: The map group doesn't seem to have a model associated with it. We should make sure all tables have a model to match

raphaellaude (commented):
Looking good. Pls don't forget to add tests incl. adding a new group.

> Auto-increment gets messed up if I INSERTing the first group row with a set ID in the migration. I could maybe set it to 0 or -1?

This would be easier if you created a new autoincrementing id for the many to many table and just set a unique constraint on the foreign keys together

> There is no warning if the `--group-slug` doesn't exist; the DistrictrMap is created without a row in the join table

Maybe set a default (as you've done) and make the foreign keys non-nullable. That way the constraint is handled in the db.

> Unclear how we will create groups (from CLI?)

Yes, since we don't want users to be able to create groups

backend/app/alembic/versions/dc0216fef023_create_map_groups.py
mapmeld: ✅

backend/app/main.py
mapmeld: ✅

backend/app/models.py
```diff
@@ -302,6 +302,13 @@ class ColorsSetResult(BaseModel):
     colors: list[str]
 
 
+class MapGroup(SQLModel, table=True):
+    __tablename__ = "map_group"
+    id: int = Field(primary_key=True)
+    name: str = Field(nullable=False)
+    slug: str = Field(nullable=False)
```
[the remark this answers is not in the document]
mapmeld: ✅  I'll use the slug as the unique primary key to resolve the join issue

mapmeld:
I pushed commits which make the group slug its primary key, and add a CLI command to name a group.
If the slug is not specified, the script. generates a lowercase a-z slug based on the name

raphaellaude (approved):
lgtm

raphaellaude (approved):
lgtm

PR #346, opened 2025-05-01
Election data improvements

nofurtherinformation:
## Description
Improves the election stat display based on feedback
- Adds a toggle when in overlay mode to view
- Changes display to be lean Republican/Democrat instead of two percentages
- Cleans up evaluation panel configuration to remove show by count / share buttons when not used

## Reviewers
- Primary: @mariogiampieri 
- Secondary:


app/src/app/components/sidebar/DemographicMapPanel.tsx
mariogiampieri, line 1: this works great and is a nice feature! when testing i got the page into a state where the option/alt context items were left up after cycling through either with keying 'x' or clicking in the demographics ui (don't remember exact sequence) but i can't reproduce. it then went away after i clicked again back on the map. just flagging that there may be some edge cases affecting behavior, **but** the features introduced here seem to work perfectly otherwise

app/src/app/store/demography/evaluationConfig.ts
```diff
@@ -42,29 +42,17 @@ export const VAPColumnConfig: EvalColumnConfiguration<SummaryStatConfig['VAP']>
...
-  {column: 'sen_18_rep', label: '2018 Sen (R)'},
-  {column: 'sen_18_dem', label: '2018 Sen (D)'},
-  {column: 'sen_16_rep', label: '2016 Sen (R)'},
-  {column: 'sen_16_dem', label: '2016 Sen (D)'},
-
-  {column: 'ag_22_rep', label: '2022 AG (R)'},
-  {column: 'ag_22_dem', label: '2022 AG (D)'},
-  {column: 'ag_18_rep', label: '2018 AG (R)'},
-  {column: 'ag_18_dem', label: '2018 AG (D)'},
+// TODO FIX typing
```
mariogiampieri, line 45: non-blocking but a reminder

mariogiampieri (approved):
great work, ux feels intuitive and user friendly. minor comments but nothing blocking

Issue #343, opened 2025-04-28
Contiguity polish


nofurtherinformation:
Contiguity components may not be responding correctly; reporting the wrong number of components and zoom to wrong areas / not helpful areas

raphaellaude:
Would love to track specific instances of this happening to get a better idea of the issues

mariogiampieri:
![Image](https://github.com/user-attachments/assets/15fdbbdb-e175-4949-af61-0051218e811a)

Per my comments on #347, see the above where I can create a new disconnected component and see the number in the dropdown update immediately, but have the indicator above not sync properly. I was able to reproduce similar behavior by adding and erasing components on the map.

nofurtherinformation:
Resolved in #354

PR #354, opened 2025-05-07
Fe Polish Part 2

### LINKED ISSUE #276: When no recent maps, recents doesn't open – should be greyed out

nofurtherinformation:
Additional FE polishes.

## Description
- Resolves #276 
- Works on #343 - adds state to track when a zone has been painted, and if it has changed since the last contiguity check the icon becomes a question mark to avoid presenting stale data
 
## Reviewers
- Primary: @raphaellaude 
- Secondary:

### DIFF
--- a/app/src/app/components/sidebar/MapValidation/ContiguityDetail.tsx
+++ b/app/src/app/components/sidebar/MapValidation/ContiguityDetail.tsx
@@ -5,31 +5,38 @@ import {Blockquote, Flex, IconButton, Spinner, Text, Tooltip} from '@radix-ui/th
 import {useQuery} from '@tanstack/react-query';
 import {queryClient} from '@utils/api/queryClient';
 import {useEffect, useState} from 'react';
-import {ChevronDownIcon, ChevronUpIcon, CrossCircledIcon} from '@radix-ui/react-icons';
+import {
+  CheckCircledIcon,
+  ChevronDownIcon,
+  ChevronUpIcon,
+  CrossCircledIcon,
+  DashIcon,
+  QuestionMarkCircledIcon,
+} from '@radix-ui/react-icons';
 
-interface ZoomToConnectedComponentsProps {
+interface ContiguityDetailProps {
   zone: number;
-  contiguity: number;
-  updateTrigger: number | string | null;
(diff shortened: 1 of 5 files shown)


app/src/app/components/sidebar/MapValidation/ContiguityDetail.tsx
```diff
@@ -38,26 +45,43 @@ export default function ZoomToConnectedComponents({
...
+        {isOutOfSync ? (
+          <QuestionMarkCircledIcon color="gray" />
+        ) : contiguity === 1 ? (
+          <CheckCircledIcon color="green" />
+        ) : (
+          <CrossCircledIcon color="red" />
+        )}
+        <Text color="gray" className={`${isOutOfSync ? 'opacity-35' : ''}`}>
+          {contiguity} component{contiguity > 1 ? 's' : ''}
+        </Text>
```
raphaellaude, line 77: Love this – I think it'll buy us some grace when we know the results are stale. This seems to be working mostly well except sometimes when refetching is briefly displays the old data as fresh then updates it. We should perhaps set the last updated in the on success for the contig endpoint?

raphaellaude (commented):
one small UAT comment otherwise looks good

app/src/app/components/sidebar/MapValidation/ContiguityDetail.tsx
nofurtherinformation: Ah good catch -- I see what you mean. I'll integrate this today and ping back here

nofurtherinformation: Also thank you!

nofurtherinformation: Resolved in 810ef92f4df36f554f5dbcc2b64c6022365c7609 and cc1da336a315a683e00f70a10415393b91154600 - useQuery has a nice stateful `dataUpdatedAt` property which we can just use instead of having some duplicated state. UAT wise this seems very nice and solid now!

raphaellaude (approved):
looks good

PR #361, opened 2025-05-10
Thumbnail cleanup

raphaellaude:
## Description

- [x] Move thumbnail endpoints to separate module / router
- [x] Replace thumbnails mock test with test that creates the file
- [x] Fix off by one error in map coloring

Other things I did while debugging
- [x] Update prod API s3 credentials (images were being sent to the old bucket)

## Reviewers

- Primary: @mapmeld @nofurtherinformation 
- Secondary: @mariogiampieri 

## Checklist


backend/app/core/config.py
```diff
@@ -137,6 +137,14 @@ def get_s3_client(self):
             **kwargs,
         )
 
+    @computed_field
+    @property
+    def cnd_url(self) -> str | None:
+        if self.CDN_URL is not None:
+            return self.CDN_URL
+
+        return self.R2_BUCKET_NAME
```
mapmeld, line 146: These aren't equivalent (CDN_URL is a navigable URL). Is this primarily to have a fallback if CDN_URL is missing?

raphaellaude: yep

raphaellaude: but yeah I see, the the fallback should be 
```suggestion
        return f"https://{self.R2_BUCKET_NAME}.s3.amazonaws.com"
```

mapmeld (approved):
changes/reorganization makes sense to me

PR #363, opened 2025-05-12
Frozen document share fix

nofurtherinformation:
Update password prompt logic: only when locked and can edit
Update logic to when to display document_id: only when edit access and locked


mariogiampieri (approved):
lgtm!

PR #365, opened 2025-05-12
Add page for groups at /group/[slug]

mapmeld:
Now that groups are in the database, we can represent them on a static page, with clickable links to open a blank map.

It's mostly copied from the place/[slug] code, I just added an image. The image will be a map preview or a fallback image

Screenshot 2025-05-11 at 11 05 15 PM

But maps are currently associated with a document UUID - how to associate a DistrictrMap with a thumbnail?
My thought is: save to the same bucket with `{districtrmap_slug}.png`. I have to review the thumbnail code to see if we can pass an option here (include_null_assignments=True) or if it's too connected to the document JOIN to use the same query/same code.


mapmeld:
Changed this to a single API call, showing all of the group's visible DistrictrMaps, and stand-in images for now 

TBD: some styling, blank maps as thumbnails, pagination and search

Screenshot 2025-05-13 at 3 58 01 PM

backend/app/main.py
```diff
@@ -1222,3 +1217,57 @@ async def get_survey_results(
         return FileResponse(
             path=_out_file, media_type=media_type, filename=out_file_name
         )
+
+
+@app.get("/api/group/{group_slug}", response_model=MapGroupResponse)
```
nofurtherinformation, line 1222: H: I think we should use the `api/gerrydb/views` endppint with the group argument

nofurtherinformation (changes requested):
FE looks good but I think we should use the existing views endpoint

backend/app/main.py
mapmeld: ok I just pushed a commit for that. I changed the params for `getAvailableDistrictrMaps` and the sort order of the response to be alphabetical; hope those changes makes sense

nofurtherinformation (dismissed):
lgtm!

nofurtherinformation (dismissed):
lgtm :)

nofurtherinformation (approved):
Nice work !

PR #380, opened 2025-05-14
Remove double status button

nofurtherinformation:
This snuck by during the block import


cdonnay (approved):
Looks good on the user side!

Issue #379, opened 2025-05-13
Font across website


nofurtherinformation:
We should use nunito
Various UI areas may be impacted

nofurtherinformation:
Resolved in #381

Issue #377, opened 2025-05-13
Slow image loading on "Rules" page


cdonnay:
The images on the "Rules of redistricting" page are really slow to load.

nofurtherinformation:
Resolved in #381

Issue #374, opened 2025-05-13
Changes to footer of site


cdonnay:
- [x] Change "Explore maps" to "Map"
- [x] Change "Places" to "List"
- [x] Change the order of links in "Learn more" to match the header of the site
- [x] Change "Get in touch" to "Big picture"

cdonnay:
Closed with #381

Issue #373, opened 2025-05-13
Self contained beta landing page


cdonnay:
- [x] Under "Districtr is a free browser-based tool for drawing districts and mapping your community." add "This is a beta release of Districtr 2.0. We have a limited selection of states available for mapping, and are accepting user feedback via the gren "Feedback" button in the bottom right corner."
- [x] Next to "You can draw districts." add "(in limited states)"
- [x] Next to "You can draw your community." add "(coming soon)"
- [x] "Import an existing plan or community map (coming soon)" should just be removed entirely. That's part of a broader UX discussion of how we want people to get to a map.

cdonnay:
Closed with #381

PR #381, opened 2025-05-14
UI fixes and copy updates

nofurtherinformation:
## Description 
- #372 
- #373 
- #374
- #377 
- #379

## Reviewers
- Primary: @cdonnay 
- Secondary:

### DIFF
--- a/app/package-lock.json
+++ b/app/package-lock.json
@@ -68,6 +68,7 @@
         "react-map-gl": "^7.1.3",
         "react-relative-time": "^0.0.9",
         "react-resizable": "^3.0.5",
+        "sharp": "^0.34.1",
         "simple-statistics": "^7.8.7",
         "superjson": "^2.2.1",
         "tiptap": "^1.32.2",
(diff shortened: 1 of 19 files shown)


cdonnay (changes requested):
Everything looks good to me! One new edit; we have new copy for the updates page.

**After many years of faithful service and thousands of community-based maps drawn, Districtr is getting a makeover in 2025!**

New/notable features include:

- Block-level detail: the ability to take a precinct and "shatter” it into its component Census blocks. This lets you draw maps very tight population deviation.
- Hotkeys/keyboard shortcuts: P to paint, M to move, etc.
- Save/share: You can share in an editable or a view-only version.
- Customizable district colors.
- Map validation:
     - Contiguity check lets you zoom to components.
     - Find unassigned units.
- Draggable and resizable toolbar.
- Import/export: we already had block assignment exports, but now we have improved block assignment imports.
- Demographics and elections: new viewing options let you see choropleths (shaded or colored maps) side-by-side or overlaid. There is also adjustable binning for the shading.
- Locking individual districts.

cdonnay (approved):

PR #385, opened 2025-05-14
Group page fix

nofurtherinformation:
## Description
- Use env variable for the api request
- Small FE cleanup

## Reviewers
- Primary: @mapmeld 
- Secondary:


mapmeld (approved):
My bad . Thanks for the fix

PR #391, opened 2025-05-15
Bugfix: stale document

nofurtherinformation:
There is an edge case where upon returning to tab that has slept, the subscription re-subscribes and loads stale data. There is never a case where we want stale document data to load in temporarily, since it has `updated_at` and `access`/`edit` state attached to it, which can change on re-focus

## Description
- Remove stale time from map document query
- Remove placeholder data from map document query

## Reviewers
- Primary: @raphaellaude 
- Secondary:


mapmeld (approved):
I don't have a good understanding of the page idle issue, but after reading https://github.com/TanStack/query/discussions/1685 I would be in favor of merging this as a fix

PR #392, opened 2025-05-15
Topbar improvements

nofurtherinformation:
This PR aims to iterate on the topbar UI with a few key enhancements

## Description
- Clarify the publish status (scratch, in progress, ready) vs the editing status (read only, in use, needs password, currently editing)
- Streamline common updates (changing map name, update publish status)
- Condense advanced functions (map details, sharing) to make more room on the top bar

## Reviewers
- Primary: @mariogiampieri 
- Secondary: @raphaellaude 

## Checklist
- [x] Refactor state to zustand store
- [x] Condense save/share to single modal
- [x] Refactor components to smaller sections of modal
- [x] Add editable header
- [x] Separate out publish and editing statuses
- [ ] UX review with Moon + Chris


app/src/app/components/Toolbar/SaveShareModal/SaveShareModal.tsx
```diff
@@ -0,0 +1,83 @@
...
+}> = ({open, onClose}) => {
+  const mapDocument = useMapStore(state => state.mapDocument);
+  const mapMetadata = useMapMetadata(mapDocument?.document_id);
+  const [innerFormState, setInnerFormState] = useState<DocumentMetadata>(
+    mapMetadata ?? DEFAULT_MAP_METADATA
+  );
+  const [linkCopied, setLinkCopied] = useState(false);
+  const isEditing = useMapStore(
+    state => state.mapDocument?.access === 'edit' && state.mapDocument?.status === 'checked_out'
+  );
```
mariogiampieri, line 25: nice, more clear

app/src/app/components/Topbar/MapTitleDisplay.tsx
```diff
@@ -0,0 +1,101 @@
...
+    );
+  }
+  if (editing) {
+    return (
+      <Popover.Root>
+        <Popover.Trigger>
+          <Flex align="center" gapX="1" direction="row" className="cursor-pointer">
+            <Tooltip content={_mapName} className={`${isTruncated ? 'w-full' : 'w-fit'}`}>
+              <Flex align="center" gapX="1" direction="row">
+                <Text size="2">{mapName || '(Edit Map Title)'}</Text>
```
mariogiampieri, line 59: PP: should read "(Edit Map Name)" for consistency across

app/src/app/components/Topbar/MapTitleDisplay.tsx
```diff
@@ -0,0 +1,101 @@
...
+                <Pencil1Icon />
+              </Flex>
+            </Tooltip>
+          </Flex>
+        </Popover.Trigger>
+        <Popover.Content>
+          <Flex direction="column" gap="2">
+            <Text size="2">Edit Map Name</Text>
+            <TextArea
+              value={mapTitleInner}
```
mariogiampieri, line 69: I: this value and not the placeholder is set to "(Edit Map Name)" so the user has to highlight and delete that in the course of setting a new name. Would be slightly more convenient otherwise

app/src/app/store/saveShareStore.ts
mariogiampieri, line 1: nice, this definitely makes sense as it's own store at this point

mariogiampieri (changes requested):
looks great overall! would be good to address the user experience in setting the map name for the first time but otherwise lgtm

app/src/app/components/Topbar/MapTitleDisplay.tsx
nofurtherinformation: Resolved in d1cf9036ffa37d9a66c2f798445830d4b4fe970b

app/src/app/components/Topbar/MapTitleDisplay.tsx
nofurtherinformation: Good catch -- resolved in d4ada26b3aa69f9fd694a8909ca40f63c940165d

app/src/app/store/saveShareStore.ts
nofurtherinformation: Thank you!

app/src/app/components/Toolbar/SaveShareModal/SaveShareModal.tsx
nofurtherinformation: Thank you! This could be refactored in the `featureFlagStore` we've been discussing to encapsulate logic in one place a bit better

nofurtherinformation:
Thanks @mariogiampieri ! I cleaned up those issues and this is ready for re-review

PR #393, opened 2025-05-17
refactor save share as separate module

raphaellaude:
## Description

- Item 1

## Reviewers

- Primary:
- Secondary:

## Checklist

### DIFF
--- a/backend/app/core/dependencies.py
+++ b/backend/app/core/dependencies.py
@@ -1,6 +1,9 @@
 from fastapi import Depends, HTTPException
 from sqlmodel import select, Session
-from app.models import Document
+from app.models import Document, DocumentPublic, DistrictrMap
+from app.save_share.models import DocumentShareStatus, DocumentEditStatus
+from sqlalchemy.sql.functions import coalesce
+from app.save_share.locks import check_map_lock
 from sqlalchemy.exc import NoResultFound
 from app.core.db import get_session
 import logging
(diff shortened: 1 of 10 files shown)


nofurtherinformation (commented):
I just gave this a quick read through but I think this approach is looking great. 

Next step that should be in probably a second PR: I think the big UX improvement (aside from backend hardening) would be making the frozen link route param be the serial ID `/map/1` -- also when maps are marked as ready for review they automatically get a serial id that can be accessed serially

nofurtherinformation:
Apropos the above and thumbnails, what if we name thumbnails with the serial ID `districtr-thumbnail-1.png` and only allow thumbnail generation for published maps?

raphaellaude:
@nofurtherinformation @mariogiampieri @mapmeld this is ready for review though the thumbnails aren't fully obscured yet / are leaking the document id.

This seems much more secure though so feeling p good.

raphaellaude:
[Here's a shared frozen map](https://districtr-v2-393-app.fly.dev/map?share=eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJ0b2tlbiI6IjVkNmM5YWY5LWRlZjQtNDI5Ni04OWFkLTYxNTFhYzcyOTI5ZCIsImFjY2VzcyI6InJlYWQiLCJwYXNzd29yZF9yZXF1aXJlZCI6ZmFsc2V9.Hkca2q7i7c3MdrBpihyffINVVFiEXe4Xus4oKTvfXUc), try to get edit access to the original!

app/src/app/components/Toolbar/ZonePicker.tsx
```diff
@@ -13,7 +14,7 @@ export function ZonePicker() {
   };
 
   return (
-    <div>
+    <div style={access === 'read' ? {pointerEvents: 'none', opacity: 0.5} : {}}>
```
nofurtherinformation, line 17: Medium: This could be refactored since it appears in multiple places

```typescript
const AccessBoundary: React.FC<{children: React.ReactNode}> = ({children}) => {
  const access = useMapStore(state => state.mapStatus?.access)
  return <Box className={access === 'read' && 'pointer-events-none opacity-50'}>{children}</Box>
}

```

app/src/app/utils/api/apiHandlers/getDocument.ts
```diff
@@ -3,16 +3,16 @@ import {DocumentObject} from './types';
...
-      .post(`${process.env.NEXT_PUBLIC_API_URL}/api/document/${document_id}`, {
-        user_id: userID,
-      })
-      .then(res => {
-        return res.data;
-      });
-  } else {
-    throw new Error('No document id found');
+  if (userID) {
+    url += `?user_id=${userID}`;
```
nofurtherinformation, line 12: H: For better URL safety use `const url = new URL(${process.env...` and `url.searchParams.set('user_id', userID)` + `axios.get(url.toString())`

nofurtherinformation:
> the UI doesn't pick up that it's now an editable map. I'd assume this should all work with subscriptions, though. @nofurtherinformation would you mind taking a look?

On it!

PR #394, opened 2025-05-19
Add map_type enum column to DistrictrMaps

mapmeld:
As discussed on Slack on the 10th, this PR would help us turn UI features on/off when making local redistricting maps for a town/city/county

Server / DB:
- add a map_type column, enum has values 'default' or 'local'
- return map_type as a string in mapDocument

Frontend, when we are in a local map:
- hide the county paintbrush checkbox
- hide county and tribal boundary options from the settings menu
- in Map Validation, show the unassigned areas tab by default, instead of contiguity

I will ask if we want to hide contiguity (since local maps often have enclaves) and state/national election results, but the core changes should be here; let me know if you think we should modify any other UI.


backend/app/models.py
```diff
@@ -89,6 +89,7 @@ class DistrictrMap(TimeStampMixin, SQLModel, table=True):
     # when you create the view, pull the columns that you need
     # we'll want discrete management steps
     visible: bool = Field(sa_column=Column(Boolean, nullable=False, default=True))
+    map_type: str | None = None
```
raphaellaude, line 92: I: If it's not nullable we should be able to remove `| None = None`

backend/app/alembic/versions/af62f0e0276b_add_map_type_enum.py
```diff
@@ -0,0 +1,35 @@
...
+def upgrade() -> None:
+    maptype_enum.create(op.get_bind(), checkfirst=True)
+    op.add_column(
+        "districtrmap",
+        sa.Column("map_type", maptype_enum, nullable=False, server_default="default"),
+    )
+
+
+def downgrade() -> None:
+    op.drop_column("mytable", "map_type")
```
raphaellaude, line 34: H: Wrong table name

```suggestion
    op.drop_column("districtrmap", "map_type")
```

app/src/app/utils/api/apiHandlers/types.ts
```diff
@@ -56,6 +56,7 @@ export interface DocumentObject extends StatusObject {
   extent: [number, number, number, number]; // [minx, miny, maxx, maxy]
   map_metadata: DocumentMetadata;
   color_scheme: string[] | null;
+  map_type: 'default' | 'local' | null;
```
raphaellaude, line 59: Q: Do we want to allow nulls?

app/src/app/components/BrushControls.tsx
```diff
@@ -5,16 +5,19 @@ import PaintByCounty from './Toolbar/PaintByCounty';
 import {ZonePicker} from './Toolbar/ZonePicker';
 export const BrushControls = () => {
   const activeTool = useMapStore(state => state.activeTool);
+  const mapType = useMapStore(state => state.mapDocument?.map_type);
```
raphaellaude, line 8: I: By using the `mapType` value inside these components, we may create a ton of complexity / conditional logic. What do you think about taking the following approach:
- New store for feature flags
- Manage all feature flags as subscriptions on the map type
- Only consume descriptive boolean feature flags in components, e.g. `{ paintByCounty ?? <PaintByCounty /> }` rather than `{ mapType !== 'local' ?? <PaintByCounty /> }`

```suggestion
  const paintByCounty = useFeatureFlagStore(state => state.paintByCounty);
```

This design would have the added advantage as also working for user-set feature flags / preferences. This could be useful in cases where we want to test out a new feature but not turn it on for any map types by default.

raphaellaude (changes requested):
Looking good. Couple small things and an idea

app/src/app/components/BrushControls.tsx
nofurtherinformation: +1 

For updating state we can subscribe to `mapStore.mapDocument` and update `featureFlagStore` on change, separating the concerns of any feature logic there

app/src/app/components/Toolbar/Settings.tsx
```diff
@@ -135,40 +136,44 @@ export const ToolSettings: React.FC = () => {
...
-              })
-            }
-          >
-            Emphasize county names
-          </CheckboxGroup.Item>
-          <CheckboxGroup.Item value="2" disabled>
-            Show tribes and communities
-          </CheckboxGroup.Item>
-        </CheckboxGroup.Root>
+        {mapType !== 'local' ? (
```
nofurtherinformation, line 139: Nit: Small preference for `!(mapType === 'local') && ...` syntax (or `!!` syntax when looking for true) which has the same safety of a ternary operator but doesn't have the trailing `null` fallback

app/src/app/utils/api/apiHandlers/types.ts
mapmeld: :+1:

app/src/app/components/Toolbar/Settings.tsx
mapmeld: :+1:

app/src/app/components/BrushControls.tsx
mapmeld: ok this was my first time adding in a store / subscription but I think this is working

backend/app/models.py
```diff
@@ -253,7 +253,7 @@ class DocumentPublic(BaseModel):
     genesis: str | None = None
     access: DocumentShareStatus = DocumentShareStatus.edit
     color_scheme: list[str] | None = None
-    map_type: str | None = None
+    map_type: str = "default"
```
raphaellaude, line 256: this should really come from the db. Are there cases where the map type is unset and needs a default?

```suggestion
    map_type: str
```

raphaellaude (dismissed):
lgtm

another quick thought: you might want to update the `create-districtr-map` and `update-districtr-map` commands in order to specify / update the map type.

nofurtherinformation (changes requested):
This looks good to me -- one small nit if we can combine the migrations into one since they are pretty connected but then I think this is good to go!

nofurtherinformation (approved):

PR #399, opened 2025-05-28
List public Documents made with a specific DistrictrMap

mapmeld:
Adds an API endpoint and a page to browse recent `Document`s for a given `DistrictrMap` slug

The API endpoint limits results to members of a public DistrictrMap (i.e. if NYC hasn't been published yet, the API won't return any results)

TODO: I think that the API endpoint should also limit results to `shared = True` and `access = 'read'`, but I don't understand how these are represented in queries

2nd todo: thumbnails?


raphaellaude:
> @mapmeld did you want a review for this? The PR is marked as draft still

Or did you assign it to me to finish?

Issue #405, opened 2025-06-26
New defaults


cdonnay:
From our testing session on 6/26/25, Moon would like the following defaults changed:

- [x] floating toolbar should start in the sidebar, and users should have to choose to make it floating
- [x] distance to ideal district size should start turned on

PR #406, opened 2025-06-26
New defaults

nofurtherinformation:
## Description
- #405 
- #414 
- Adds smarter snapping/rotating logic to the toolbar 

## Reviewers
- Primary: @mapmeld 
- Secondary: @raphaellaude

### DIFF
--- a/app/src/app/components/Toolbar/Settings.tsx
+++ b/app/src/app/components/Toolbar/Settings.tsx
@@ -94,7 +94,7 @@ export const ToolSettings: React.FC = () => {
               })
             }
           >
-            Show total population labels on all geometries
+            Show population labels on whole precincts
           </CheckboxGroup.Item>
           <CheckboxGroup.Item
             value="showBlockPopulationNumbers"
@@ -105,7 +105,7 @@ export const ToolSettings: React.FC = () => {
             }
             disabled={!mapDocument?.child_layer}
           >
-            Show total population labels on blocks
+            Show population labels on exposed blocks
           </CheckboxGroup.Item>
           <CheckboxGroup.Item
             value="showPaintedDistricts"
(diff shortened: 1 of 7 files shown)


nofurtherinformation:
Smart -- yes that makes sense! Also, we have this set up to remember your preferences (sidebar or map location, draggable or fixed)

raphaellaude (commented):
Personally not the biggest fan of the default being draggable but it's ok!

It looks like you've got some white div background bleeding through 

Image

nofurtherinformation:
> Personally not the biggest fan of the default being draggable but it's ok!

The new default keeps it in the sidebar, so draggable / toolbar on the map in general has a layer of opt-in!

> It looks like you've got some white div background bleeding through
> 
> Image

Good catch -- this is an issue on main, too, so good to get this cleaned up!

mapmeld:
1. With any size toolbar I am getting this info below the toolbar, where it gets cut off at the bottom of the screen
Screenshot 2025-07-08 at 9 33 57 AM

2. When I am dragging the toolbar, it snaps to the left edge of the map, I am OK with this as a design. It's only unexpected if I drag it slightly right and it snaps back (when I first turned on draggable toolbar, it was in the top left corner, so an experimental drag snapped back to the left edge and rotated it...)

3. With vertical toolbar, it's not too hard to hit the move button, but the rotate and return-to-toolbar buttons are tricky, I think because of the border-radius, I'm touching the map before I touch the button
Screenshot 2025-07-08 at 9 40 03 AM
Screenshot 2025-07-08 at 9 40 47 AM

nofurtherinformation:
Thank you Nick!

> 1. With any size toolbar I am getting this info below the toolbar, where it gets cut off at the bottom of the screen

This should now be resolved.

> 2. When I am dragging the toolbar, it snaps to the left edge of the map, I am OK with this as a design. It's only unexpected if I drag it slightly right and it snaps back (when I first turned on draggable toolbar, it was in the top left corner, so an experimental drag snapped back to the left edge and rotated it...)

I tweaked the tolerances for this to be less aggressive with snapping and autorotation, let me know what you think!

> 3. With vertical toolbar, it's not too hard to hit the move button, but the rotate and return-to-toolbar buttons are tricky, I think because of the border-radius, I'm touching the map before I touch the button

I adjusted the button styling to be more compact and less rounded; let me know if this makes it easier!

Issue #407, opened 2025-06-26
Tooltip disagrees with pop labels


cdonnay:
When shattering precincts, we observed that sometimes the pop tooltip gave a different value than the pop labels.

- sometimes the tooltip showed 0 pop while the label was non-zero
- sometimes the tooltip showed a number that was twice the label, and then eventually changed to match the label

Observed in Philly and KS. Chris has screen recordings of the Philly occurrence.

nofurtherinformation:
This is mostly resolved in #408, and was due to Maplibre getting duplicate features selected in some edge cases (eg. selecting 2 copies of a geometry where it is split)

PR #408, opened 2025-06-26
deduplicate selected features

nofurtherinformation:
## Description
- #407

## Reviewers
- Primary: @mariogiampieri 
- Secondary: @mapmeld

### DIFF
--- a/app/src/app/utils/helpers.ts
+++ b/app/src/app/utils/helpers.ts
@@ -15,6 +15,7 @@ import {MapStore, useMapStore} from '../store/mapStore';
 import {NullableZone} from '../constants/types';
 import {demographyCache} from './demography/demographyCache';
 import {DocumentMetadata} from '@utils/api/apiHandlers/types';
+import {onlyUniqueProperty} from './arrays';
 
 /**
  * PaintEventHandler
@@ -429,11 +430,13 @@ export const checkIfSameZone = (
  * The function returns an array of features that pass all the filtering criteria.
  */
 const filterFeatures = (
-  features: MapGeoJSONFeature[],
+  _features: MapGeoJSONFeature[],
   filterLocked: boolean = true,
   additionalFilters: Array<(f: MapGeoJSONFeature) => boolean> = [],
   allowOutsideCaptiveIds: boolean = false
 ) => {
+  // first, dedupe
+  const features = _features.filter(onlyUniqueProperty('id'));
   const {
     activeTool,
     captiveIds,


nofurtherinformation:
The second failure case appears when a geometry is split and we get two instances of the geometry queried.

app/src/app/utils/arrays.ts
```diff
@@ -1,7 +1,38 @@
...
+ * A filter function that returns true only for the first occurrence of each unique value in an array.
+ *
+ * @param {unknown} value - The current element being processed
+ * @param {number} index - The index of the current element
+ * @param {unknown[]} self - The array being filtered
+ * @returns {boolean} - True if this is the first occurrence of the value, false otherwise
+ */
 export const onlyUnique = (value: unknown, index: number, self: unknown[]) => {
   return self.indexOf(value) === index;
 };
```
raphaellaude, line 12: Based on a bit of googling it looks like the best time complexity we can get is `O(n^2 / 2)`. I think you're probably hitting that with this solution but not 100% sure.

That just raises the question – why not use lodash's [uniqBy](https://lodash.com/docs/4.17.15#uniqBy) or vendor their implementation?

nofurtherinformation: Good catch, I was assuming that overall perf here would be negligible due to the operation but I was wrong, here are the numbers I'm seeing:

```
Current implementation
25.303960391200416 'ms average dedupe time'
26.907920793141468 'ms average dedupe time'
30.707920792669352 'ms average dedupe time'
35.99009901344186 'ms average dedupe time'
36.908910888256415 'ms average dedupe time'
``` 

So, the current is quite slow, especially since this operation runs so often. I tried lodash and a quick implementation aimed at perf and here's what I saw:

```
Lodash uniqBy
1.4257425713067007 'ms average dedupe time'
1.7455445528030396 'ms average dedupe time'
2.298019802806401 'ms average dedupe time'
2.4732673256704123 'ms average dedupe time'


fastUniqBy Implementation (Vanilla for loop + set)
0.21287128948929288 'ms average dedupe time'
0.1326732605990797 'ms average dedupe time'
0.24653465263914354 'ms average dedupe time'
0.297029703560442 'ms average dedupe time'
```

In this case either is a big improvement, but squeezing out 1-2ms per paint operation is probably worth it! Radash is a little faster than Lodash too, but I don't think we need that dependency for now

raphaellaude (approved):

PR #409, opened 2025-06-27
Point selection fallback strategy

nofurtherinformation:
When selecting small geographies from an outer zoom, often they get missed. We've played around with a handful of strategies, but a possible solution is to simply have representative points for all geographies always present; or at least all parent geos and child geos as needed.

## Description
- [Temp] Pulls out the point data from gpkgs for a test case (Kansas)
- Uses a point layer as an additional selection layer

## Reviewers
- Primary: WIP
- Secondary:

## Checklist
- [ ] Add point selection layers
- [ ]Tweak existing selection logic
- [ ] Test performance
- [ ] Update pipeline

### DIFF
--- a/app/src/app/components/Map/PointSelectionLayer.tsx
+++ b/app/src/app/components/Map/PointSelectionLayer.tsx
@@ -1,0 +1,81 @@
+import {BLOCK_LAYER_ID, BLOCK_LAYER_ID_CHILD, BLOCK_POINTS_LAYER_ID, BLOCK_POINTS_LAYER_ID_CHILD} from '@/app/constants/layers';
+import {useLayerFilter} from '@/app/hooks/useLayerFilter';
+import {useMapStore} from '@/app/store/mapStore';
+import { asyncBufferFromUrl, parquetReadObjects } from 'hyparquet';
+import React, {useEffect, useRef, useState} from 'react';
+import {Layer, Source} from 'react-map-gl/dist/esm/exports-maplibre';
+
+const generateGeojson = (pointData: any[], mapDocument: MapDocument, child: boolean) => {
+  const source = child ? BLOCK_LAYER_ID_CHILD : BLOCK_LAYER_ID;
+  const sourceLayer = child ? mapDocument?.child_layer : mapDocument?.parent_layer;
+  return {
+    type: 'FeatureCollection',
+    features: pointData.map((d: any) => ({
+      type: 'Feature',
+      geometry: {type: 'Point', coordinates: [d.x, d.y]},
+      properties: {
+        path: d.path,
+        total_pop_20: parseInt(d.total_pop_20),
+        __source: source,
+        __sourceLayer: sourceLayer,
+      },
+    })),
+  } as GeoJSON.FeatureCollection<GeoJSON.Point>;
+};
+
+export const PointSelectionLayer: React.FC<{child?: boolean}> = ({child = false}) => {
+  const fullData = useRef<any[] | null>(null);
+  const data = useRef<GeoJSON.FeatureCollection<GeoJSON.Point> | null>(null);
+  const mapDocument = useMapStore(state => state.mapDocument);
+  const brokenChildIds = useMapStore(state => state.shatterIds.children);
+  const layerFilter = useLayerFilter(child);
+  const sourceID = 'SELECTION_POINTS' + (child ? '_child' : '');
+  const [dataHash, setDataHash] = useState<string | null>(null);
+
+  useEffect(() => {
+    const fetchData = async () => {
+    const layer = child ? mapDocument?.child_layer : mapDocument?.parent_layer;
+    if (layer) {
+      const url = `${process.env.NEXT_PUBLIC_S3_BUCKET_URL}/dev/${layer}_points.parquet`;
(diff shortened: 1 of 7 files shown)


app/src/app/components/Map/PointSelectionLayer.tsx
```diff
@@ -0,0 +1,81 @@
+import {BLOCK_LAYER_ID, BLOCK_LAYER_ID_CHILD, BLOCK_POINTS_LAYER_ID, BLOCK_POINTS_LAYER_ID_CHILD, EMPTY_FT_COLLECTION} from '@/app/constants/layers';
+import {useLayerFilter} from '@/app/hooks/useLayerFilter';
+import {useMapStore} from '@/app/store/mapStore';
+import { asyncBufferFromUrl, parquetReadObjects } from 'hyparquet';
+import React, {useEffect, useRef, useState} from 'react';
+import {Layer, Source} from 'react-map-gl/dist/esm/exports-maplibre';
```
raphaellaude, line 6: Q: why use the esm export?

app/src/app/components/Map/PointSelectionLayer.tsx
```diff
@@ -0,0 +1,81 @@
...
+  return (
+    <Source id={sourceID} type="geojson" promoteId="path" data={data.current || EMPTY_FT_COLLECTION}>
+      <Layer
+        id={child ? BLOCK_POINTS_LAYER_ID_CHILD : BLOCK_POINTS_LAYER_ID}
+        source={sourceID}
+        filter={layerFilter}
+        type="circle"
+        paint={{
+          'circle-radius': 1,
+          'circle-color': '#000000',
```
raphaellaude, line 76: Just in dev that we need to see these, correct? For debugging it could be nice to set a dev feature flag to false which sets the layer visibility

app/src/app/components/Map/PointSelectionLayer.tsx
```diff
@@ -0,0 +1,81 @@
...
+  const brokenChildIds = useMapStore(state => state.shatterIds.children);
+  const layerFilter = useLayerFilter(child);
+  const sourceID = 'SELECTION_POINTS' + (child ? '_child' : '');
+  const [dataHash, setDataHash] = useState<string | null>(null);
+
+  useEffect(() => {
+    const fetchData = async () => {
+    const layer = child ? mapDocument?.child_layer : mapDocument?.parent_layer;
+    if (layer) {
+      const url = `${process.env.NEXT_PUBLIC_S3_BUCKET_URL}/dev/${layer}_points.parquet`;
```
raphaellaude, line 39: I'm not seeing any requests to this endpoint, though the points are showing up 🤷🏻

app/src/app/components/Map/PointSelectionLayer.tsx
```diff
@@ -0,0 +1,81 @@
...
+      const file = await asyncBufferFromUrl({ url }) // wrap url for async fetching
+      const parquetData = await parquetReadObjects({
+        file,
+        columns: ['path', 'x', 'y', 'total_pop_20'],
+      }) as any;
+      if (child) {
+        fullData.current = parquetData;
+        setDataHash(new Date().toISOString());
+      } else {
```
raphaellaude, line 48: I: It's a little confusing for this use effect to do one of two things depending on whether the layer is a parent/child. IMO this use effect should do all the work of generating the geojson or only fetching the data instead of splitting the work across two use effects.

app/src/app/components/Map/PointSelectionLayer.tsx
```diff
@@ -0,0 +1,81 @@
...
+      }
+    }
+  }, [child, brokenChildIds, dataHash]);
+
+  return (
+    <Source id={sourceID} type="geojson" promoteId="path" data={data.current || EMPTY_FT_COLLECTION}>
+      <Layer
+        id={child ? BLOCK_POINTS_LAYER_ID_CHILD : BLOCK_POINTS_LAYER_ID}
+        source={sourceID}
+        filter={layerFilter}
```
raphaellaude, line 72: Q: Does this filter the parent layer, too?

raphaellaude (commented):
Overall definitively better captures tiny geometries when painting. Nice thinking! Just a few small comments / mostly questions

app/src/app/components/Map/PointSelectionLayer.tsx
nofurtherinformation: Intellisense not quite intelli enough (or sensing enough) -- resolved in 7863d5bf535fb2332ce6ecfe490b416a7ac29539

app/src/app/components/Map/PointSelectionLayer.tsx
nofurtherinformation: Yes, that's right -- and agreed!

app/src/app/components/Map/PointSelectionLayer.tsx
nofurtherinformation: This has been cleaned up to:
1. Partially load child data as needed
2. Load through the parquet worker (for shared/general file handling)
3. Have some nicer abstractions/cleaner separation of concerns

raphaellaude (approved):
lgtm nicely done

PR #415, opened 2025-07-07
Save+share using public ids instead of tokens

nofurtherinformation:
This PR refactors save/share functionality to incorporate a public sequential ID (accessed when a map is moved off of 'scratch work' or shared with a frozen link. Additionally, it moves toward sharing in one of the three forms:
-`map/{public id}` - Read only public map
-`map/edit/{document id}` - Editable map
-`map/{public id}?pw=true` - Password required share link

## Reviewers

- Primary: @raphaellaude 
- Secondary:

## Checklist
- [x] Creating a map takes you to `/map/edit/foo`
- [x] Sharing a link frozen yields `/map/{public id}`
- [x] Sharing a link with password yields `/map/{public id}?pw=true`
- [x] Password unlocking a pw link redirects to the edit ID
- [x] Public maps can be copied
- [x] In-use (locked/checked out maps can be copied)
- [x] Thumbnails generate from the edit ID and generate to the public ID


backend/app/main.py
```diff
@@ -809,36 +794,15 @@ async def get_connected_component_bboxes(
 
 @app.put("/api/document/{document_id}/metadata", status_code=status.HTTP_200_OK)
 async def update_districtrmap_metadata(
-    document_id: str,
-    metadata: dict = Body(...),  # TODO: Type this properly
+    metadata: DocumentMetadata,
```
raphaellaude, line 797: This and the `.values(map_metadata=metadata.model_dump(exclude_unset=True))` fix below patches an important security issue in which arbitrary data could be added to the document metadata.

backend/app/models.py
```diff
@@ -235,7 +232,7 @@ class AssignmentsResponse(SQLModel):
     geo_id: str
     zone: int | None
     parent_path: str | None
-    document_id: str
+    # document_id: str
```
raphaellaude, line 235: Don't leak document id. Not really needed in response since the client already knows which document it's working with

.gitignore
```diff
@@ -85,3 +85,7 @@ pipelines/districtr_pipelines.egg*
 docker-compose.override.yml
 app/.env.docker
 pipelines/build/*
+
+# editors
+.vscode/
+.zed/
```
raphaellaude, line 91: was doing some debugging with debugger!

backend/app/save_share/main.py
```diff
@@ -136,91 +137,51 @@ async def share_districtr_plan(
...
                 "password_hash": hashed_password,
             },
-        )
-        session.commit()
-    return {"document_id": document.document_id, "public_id": document.public_id}
+        ).scalar_one()
 
+        session.commit()
 
-@router.post("/api/share/load_plan_from_share", response_model=DocumentPublic)
```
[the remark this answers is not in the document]
nofurtherinformation: Resolved via adding the password functionality to `/checkout`, which was already mostly there

nofurtherinformation:
I think this is ready for review. Some key updates:
- The assignment reload on tab focus is now resolved; tanstack had a behavior to auto re-fetch on focus which caused the behavior and has been flagged to not refetch unless we do it manually
- The refocus UX has been further cleaned up with some additional logic to prevent flashing of progress indicators
- Metadata now comes from the mapDocument rather than userMaps, which caused public views to not have the correct metadata
- Custom cursor icons are fixed, this was a filepath issue 
- I removed the publish requirement for now. I think this leads to some tricky edge cases and may not ultimately be better UX
- Thumbnails on the preview are being fussy, but I think this is just an environment issue uploading to the wrong s3 bucket
- Added better error text when going to a document that doesn't exist

nofurtherinformation:
Good catches on the PW field collapsing and redirect -- I'll update those today

raphaellaude:
> Lgtm! Nice work

going to merge to unblock some downstream work

Issue #417, opened 2025-07-09
Indiana local map title


cdonnay:
The title bar of an Indiana local map does not have the name of the locality next to the map title.

<img width="870" height="194" alt="Image" src="https://github.com/user-attachments/assets/b1ebc106-5624-4ac1-af0a-bc8879dea341" />

<img width="1128" height="229" alt="Image" src="https://github.com/user-attachments/assets/84b6a363-32cf-42f4-aedc-e83e960e7786" />

PR #418, opened 2025-07-21
Load source geography name (state / town name) for top bar

mapmeld:
The top bar of the map shows `DistrictrMap.name`, which is the name of the state/city/source geography of this map, as seen in  #417

The name was extracted from `mapViews` so our local maps (not being part of the public list) were not showing a name.
I also saw some quirks with loading a blank state map, where it couldn't get the name from `mapViews` until I started painting and assignments got saved.

This PR returns the name in the `mapDocument` object we use for `num_districts` and other DistrictrMap info. For Indiana maps it might still be a little different because the states' names include a number of districts ("Florida Congressional (28)") and Indiana does not.

### DIFF
--- a/app/src/app/components/Topbar/MapHeader.tsx
+++ b/app/src/app/components/Topbar/MapHeader.tsx
@@ -12,11 +12,6 @@ export const MapHeader: React.FC = () => {
   const mapDocument = useMapStore(state => state.mapDocument);
   const mapMetadata = useMapMetadata(mapDocument?.document_id);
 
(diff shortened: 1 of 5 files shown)


nofurtherinformation (changes requested):
Thank you @mapmeld this looks great! 

One small naming thing, would it be more consistent to call this "map_module"? I think that is the most consistent internal term we use (although not always)

nofurtherinformation (approved):
Nice fix -- thank you!

PR #420, opened 2025-07-25
Stop chart from rendering before parent is ready

nofurtherinformation:
Small fix to silence warning

When loading into a map, the chart would try to render before the parentSize hook has the width, firing some SVG warnings.


raphaellaude (approved):
seems like build failure was just an intermittent fly.io thing 🙄 so lgtm

Issue #104, opened 2024-09-25
Tag pages, galleries, etc


mariogiampieri:

nofurtherinformation:
I think this is done but let's double check as part of 2.2.2

PR #429, opened 2025-08-05
Map galleries component

nofurtherinformation:
#104 

## Description
- Adds a more ergonomic backend to query a list of plans/map documents
- Adds a FE component 
- Adds CMS bindings for the component
- This component will NOT ssr

## Reviewers
- Primary: @mapmeld 
- Secondary:

## Checklist
- [x] Add `/api/documents/list` endpoint
- [x] List endpoint can query based on  given IDs or given tags
- [x] List endpoint is paginated
- [x] Test list endpoint
- [x] Component for map gallery
- [x] CMS bindings for map gallery component
- [x] FE rendering for map gallery component

Tag/place page:
Screenshot 2025-08-05 at 1 32 22 PM

CMS editor:
Screenshot 2025-08-05 at 1 32 38 PM


mariogiampieri:
Q/I: Wondering if it's a Fly issue but I get 403s for each of the thumbnails that are loaded:

image

It looks like there's supposed to be a fallback image in cases where images aren't available or don't load based on the screenshot above.

mariogiampieri:
H: The gallery component has a blinking cursor when the user clicks into it. It appears the user can edit the gallery content inline, but the update content button doesn't activate and pressing delete removes the entire gallery. The gallery should only be editable via the gear icon and not inline in the wyswig editor or the form should be more robust.

nofurtherinformation:
Ty @mariogiampieri ! I've fixed the issue with the fallback image (in the future, we need to set up a CRON job to generate thumbnails). We'll want a future PR with some CMS fixes

backend/app/main.py
```diff
@@ -550,6 +550,53 @@ async def get_document_object(
         )
 
 
+@app.get("/api/documents/list")
+async def get_document_list(
```
mapmeld, line 554: Please add a test for this endpoint

nofurtherinformation: Ok we're good now

mariogiampieri (approved):
Fallback image works fine now and understand re: having a CMS-focused PR in the future to handle those issues.

PR #435, opened 2025-08-20
Make use of index in uploader query

mapmeld:
This has two changes:
- modifies the WHERE query so we are searching edges with our index (much faster, possible to upload the California block CSV)
- handles a frontend bug which popped up as undefined

I'm OK to drop the frontend change if it seems too weird


nofurtherinformation (approved):
Lgtm

PR #436, opened 2025-08-20
Add a comment list component

mapmeld:
Submitting as a draft PR for now

- API endpoint `/api/comments/list` can accept `document_id=__` , `tag=___` , or both to return recent comments and their associated tags
- tests for new endpoint
- Component shows a list of comments with tags:
Screenshot 2025-08-11 at 12 10 41 AM

**Todos**
- paging / infinite scroll if there are dozens of comments
- including a map link, or thumbnail image of a mp


mapmeld:
ok I am marking this for review and will plan on adding pagination / infinite scroll in a future update

PR #437, opened 2025-08-20
Add district map level comment and context info

nofurtherinformation:
We need additional information at the map module level to clarify a few things.

## Description
- Optional comment that specifies additional info
- Adds data source, parent geography, and child geography to specify (on hover over map module)

## Reviewers
- Primary: @mapmeld 
- Secondary: @raphaellaude 

Screenshot 2025-08-20 at 1 38 11 PM
Screenshot 2025-08-20 at 1 38 03 PM


app/src/app/components/Topbar/MapHeader.tsx
```diff
@@ -1,5 +1,5 @@
 'use client';
-import {Text, Flex} from '@radix-ui/themes';
+import { Flex} from '@radix-ui/themes';
```
mapmeld, line 2: minor style / spacing fix
can we add this to pre-commit / lint rules?

mapmeld: doesn't need to be in this PR, just I have noticed it comes up often

nofurtherinformation: Yeah, the precommit should be running prettier but doesn't seem to do so consistently

PR #440, opened 2025-08-21
Optimize parquet loading

nofurtherinformation:
This PR works on optimizing tabular and point parquet loading for plans that have a large number of broken parent units (VTDs / CBGs). Currently, the browser needs to fire off a large number of requests to get all the required data. This is fine with a couple dozen broken units, but gets painful and slow as you crest over many dozens (or hundreds, thousands) in some plans. 

## Description
- Set thresholds of number of broken parent units, after which the whole file should be fetched (the benefit of partial fetching is outweighed by the number of requests)
- Set a threshold for a parquet being too big to plausibly download
- Load the file into [OPFS](https://developer.mozilla.org/en-US/docs/Web/API/File_System_API/Origin_private_file_system)
- Run queries against that 

## Reviewers
- Primary:
- Secondary:

## Checklist
- [x] Add cache utilities to parquet worker
- [x] Update parquet worker types
- [ ] UAT
- [ ] Consider `navigator.connection.effectiveType` for making decisions on download strategies?


nofurtherinformation:
This is a bit fussier than I had hoped; for now closing this in favor of public pre-unioned views.

PR #446, opened 2025-08-22
Workflow to create district unions from ready-to-share documents, for use in thumbnail generation, future analytics, and public views

nofurtherinformation:
This PR adds a workflow for storing pre-computed district unions with demographic data aggregation. When users unlock map documents marked as "ready to share", the backend will automatically generates unioned geometries by district zone and aggregates demographic statistics from the associated gerrydb tables. The thumbnail pipeline runs after this, using the already unioned geometries. This improves thumbnail generation performance and provides ready-to-use unioned districts for future analytics. Next PR should use these for frontend display of public maps.

## Description
- Database Schema
  - Added district_unions partitioned table in document schema
  - Indexed by auto-incrementing ID and by document_id for optimal performance and maintenance
  - Stores unioned MultiPolygon geometries grouped by zone
  - Includes JSONB demographic_data column for flexible demographic storage
-  Model Updates
  - Added DistrictUnions SQLModel with partitioning configuration
  - Enhanced imports for geoalchemy2 and JSON support
- Demographic Aggregation
  - `update_or_select_district_unions` handles on-demand fetching of evaluation stats and workflow to generate stats, if non-existent or outdated
  - Dynamically discovers numeric columns in gerrydb tables via information_schema
  - Builds JSON aggregation using SUM() for all demographic fields (population, voting data, etc.)
  - Joins on path field between zone assignments and demographic tables
  - Gracefully handles documents without associated gerrydb tables

! Why JSONB !
Yep it's a new JSON column. Because each gerrydb view has dynamic columns (eg election stats) we'd need either per map module or per document views. I'm open to this but my instinct was to follow the existing `document` schema

! Why ~Partitioned Table~ a table and not Materialized View !
Given that this is a heavy workflow and needs unique logic to validate per-map document (eg avoid running if map was not updated since last district unioning), the materialized view logic seems pretty tricky to maintain and the materialized view benefits are not worth the added complexity vs a simple table 

## Reviewers
- Primary: @raphaellaude 
-Secondary:

## Checklist
- [x] Create a schema for unioned districts and aggregate demographic stats
- [x] Create functions to handle running these updates
- [x] Add a background task workflow 
- [x] Update thumbnails using unioned geos, if pre-processed
- [x] Optimize and avoid duplicate work
- [x] Update tests


backend/app/models.py
```diff
@@ -282,3 +283,15 @@ class DistrictrMapsToGroups(SQLModel, table=True):
...
         )
     )
+
+
+class DistrictUnions(TimeStampMixin, SQLModel, table=True):
+    __table_args__ = ({"postgresql_partition_by": "LIST (document_id)"},)
+    __tablename__ = "district_unions"  # pyright: ignore
+    metadata = MetaData(schema=DOCUMENT_SCHEMA)
+    document_id: str = Field(sa_column=Column(UUIDType, primary_key=True))
+    zone: int = Field(primary_key=True)
```
raphaellaude, line 293: H: document id should be a foreign key.

PP: I'd personally prefer an autoincrementing integer primary key with a secondary index on the document id, rather than a compound primary key.

backend/app/save_share/main.py
```diff
@@ -47,9 +55,210 @@ def verify_password(password: str, hashed_password: str) -> bool:
...
+        logger.info(
+            f"Updated district unions for document {document_id} in {time() - time0} seconds"
+        )
+
+    except Exception as e:
+        session.rollback()
+        logger.error(
+            f"Failed to update district unions for document {document_id}: {str(e)}"
+        )
+        raise
```
raphaellaude, line 223: rollback will occur when the session falls out of scope anyway but this is generally good to have sooner than later, esp. if you're continuing to use the same transaction. anyway it's a good reminder to me that we probably need a slightly better / standardized pattern for rollbacks

backend/app/main.py
```diff
@@ -88,9 +92,14 @@
...
-async def cleanup_expired_locks():
+async def cleanup_expired_locks(
+    background_tasks: BackgroundTasks,
+):
     session = next(get_session())
-    _cleanup_expired_locks(session=session, hours=1)
+    document_ids = _cleanup_expired_locks(session=session, hours=1)
+    _bulk_update_district_stats(
+        session=session, document_ids=document_ids, background_tasks=background_tasks
+    )
```
raphaellaude, line 102: Q/H: what in your view are the trade-offs of performing unions on cleanup rather than on read? Wouldn't this approach mean districts are usually 1 hour out of date? The most frequent access pattern is probably a user that wants to look at their own map after they publish it – at which point the lock will not have released yet, right? Could we instead update the unions on read, regenerating the unions if the timestamp of the last union update is < the timestamp of the last assignment update? Timestamps have issues in multi-leader replicated dbs but since we're running a single node set-up the timestamp approach would probably be fine.

nofurtherinformation: I was going back and forth on this point. I'm game to try on read and see how it goes!

backend/app/alembic/versions/ca985583b5df_add_demographic_data_to_district_unions.py
```diff
@@ -0,0 +1,34 @@
+"""add_demographic_data_to_district_unions
+
+Revision ID: ca985583b5df
+Revises: dc7408074929
+Create Date: 2025-08-22 14:53:41.206948
```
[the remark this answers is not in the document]
nofurtherinformation: Resolved in 2d86cf3e8f00165b8c02ae44ab823ce055441396

backend/app/alembic/versions/dc7408074929_create_district_union_view.py
```diff
@@ -0,0 +1,62 @@
...
+        sa.Column("zone", sa.Integer, nullable=False),
+        sa.Column("geometry", Geometry("MULTIPOLYGON", 4326), nullable=False),
+        sa.Column(
+            "created_at", sa.TIMESTAMP(timezone=True), server_default=sa.text("NOW()")
+        ),
+        sa.Column(
+            "updated_at", sa.TIMESTAMP(timezone=True), server_default=sa.text("NOW()")
+        ),
+        sa.PrimaryKeyConstraint("document_id", "zone"),
+        postgresql_partition_by="LIST (document_id)",
```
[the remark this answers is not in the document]
nofurtherinformation: Resolved in c48b0f7a0a3d43d0894e384bf020fd56d3ee853c

backend/app/models.py
nofurtherinformation: Resolved in ce4643eacb435d96f355344dd8c7e2dbbdd78ddc

backend/app/utils.py
```diff
@@ -360,3 +362,148 @@ def add_districtr_map_to_map_group(
 
     if autocommit:
         session.commit()
+
+
+def update_or_select_district_stats(
```
raphaellaude, line 367: PP: given this isn't a general util, slight preference for moving this to a new module

raphaellaude (approved):
only nits. lgtm

PR #447, opened 2025-08-22
Comment check review options and portal

nofurtherinformation:
Review portal and API endpoints

## Description
- Adds an admin only page to screen false positives and false negatives 
- Tweaks moderation tests and admin list endpoint

## Reviewers
- Primary: @raphaellaude 
- Secondary:

## Checklist
- [x] Tweak admin list and API endpoints
- [x] Add admin pages and workflows
- [x] Comments can be filtered by tag, place, zip, and status
- [x] Comments can be full approved / rejected, or individual parts can be approved/rejected

Screenshot 2025-09-05 at 2 37 41 PM

### DIFF
--- a/backend/tests/test_comments.py
+++ b/backend/tests/test_comments.py
@@ -1032,7 +1032,6 @@ class TestCommentListEndpoints:
 
         response = client.post("/api/comments/submit", json=form_data)
         assert response.status_code == 201
-        handle_full_submission_approve(client, response.json())
 
         # Test admin endpoint (auth is mocked in conftest.py)
         response = client.get("/api/comments/admin/list")


backend/tests/utils.py
```diff
@@ -4,3 +4,17 @@ def string_to_bool(booly: str):
 
 async def fake_verify_recaptcha(*args, **kwargs):
     return True
+
+
+async def fake_verify_auth(*args, **kwargs):
+    return {
+        "sub": "test_user",
+        "scope": [
+            "create:content",
```
raphaellaude, line 13: H: it doesn't look like this is being used

backend/app/comments/main.py
```diff
@@ -313,6 +319,8 @@ def get_comments_base_query(
     public_id: int | None,
     moderation_threshold: float = MODERATION_THRESHOLD,
     exclude_rejected: bool = True,
+    moderate: bool = True,
+    admin_columns: bool = False,
```
raphaellaude, line 323: PP/I: This function is overloaded and its interface is quite complex. IMO this function should be a base query – as in the function name – which sort of implies it doesn't handle special cases like admins. To achieve what you're doing here, you should be able to add columns in the admin endpoint with [`stmt.add_columns()`](https://docs.sqlalchemy.org/en/20/core/selectable.html#sqlalchemy.sql.expression.Select.add_columns).

This pattern should also allow us to implement the comment filtering for the two different users a bit more simply in each endpoint rather than having to handle both cases in a single function.

raphaellaude (commented):
Admin UI looks good. Couple backend comments

raphaellaude (approved):

PR #456, opened 2025-09-15
Test nextjs api route

nofurtherinformation:
WIP / Test -- seeing if nextjs api routes on fly are performant. They were not locally but I want to double check if this is docker/node being fussy.


nofurtherinformation:
Not performant. Closing this

PR #457, opened 2025-09-16
test speed with pg duckdb

raphaellaude:
## Description

- Tested locally and improved `/stats` endpoint by about half.
- Spent some time trying to figure out deploy but haven't gotten it yet. Feels like it would be a decent lift so this is probably not worth it...

## Reviewers

- Secondary:

## Checklist

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -45,7 +45,6 @@ next-env.d.ts
 pnpm-lock.yaml
 
 # data
-db/*
 assets/
 **/public/**/*.parquet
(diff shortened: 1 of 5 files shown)


raphaellaude:
We can keep it open for now! I'll see if I can't get it to work. One of the big value adds IMO would be in pipelining data into the DB.

nofurtherinformation:
Closing for now -- this is stale but now that we are on AWS it could be worth revisiting

PR #473, opened 2026-01-22
Allow multiple state county boundaries and labels at the same time

nofurtherinformation:
In the upcoming Navajo Nation map module, we'll have multiple states present. We want to show the county boundaries and county labels for multiple state. The current schema uses a 2-digit STATEFP string (eg `03`, `17`, etc.) and an equality filter. The revised schema uses a `Set<string>` type and a `match` filter in maplibre.

Additionally, explicit _is_ better than implicit, so I've added an optional, explicit field to set the valid State FIPS code(s) for the map module. If not present, the frontend will default to inferring from the map tiles, so we don't need to backfill these. However, this allows us to have arbitrary geographies with IDs that are not numeric and not adherent to the Census FIPS code schema. In the case of the Navajo Nation chapters, the IDs provided are just text, and also some of the chapters span states.

Last, the parent-child-edges UDF was causing issues mysteriously with the new data for Navajo Nation. So, it has been refactored away into much easier to debug SQLAlchemy / python.
 
## Description
- Revise map controls state schema to use a set instead of string
- Revise map controls state update function to avoid excess state updates
- Revise county layer to expect multiple STATE Fips codes
- Adds State FIPS column to the districtr map module table
- Keeps GEOID inference as a fallback
- Refactors UDF away

## Reviewers
- Primary: @launaG
- Secondary:

## Checklist
- [x] Existing map module(s) checked
- [x] Map module(s) with multiple states checked

### DIFF
--- a/backend/app/utils.py
+++ b/backend/app/utils.py
@@ -10,7 +10,6 @@ from app.models import Document, DistrictUnionsResponse
 from fastapi import BackgroundTasks
 from app.thumbnails.main import generate_thumbnail, THUMBNAIL_BUCKET
 from app.core.config import settings
-import uuid
 
 metadata = MetaData()
 logger = logging.getLogger(__name__)
@@ -195,12 +194,6 @@ def create_parent_child_edges(
         session: The database session.
         districtr_map_uuid: The UUID of the districtr map.
     """
-    # validate uuid
-    try:
-        uuid.UUID(districtr_map_uuid)
-    except ValueError:
-        raise ValueError(f"Invalid UUID: {districtr_map_uuid}")
-
     stmt = select(DistrictrMap).where(DistrictrMap.uuid == districtr_map_uuid)
     map_row = session.exec(stmt).one_or_none()
 


backend/app/alembic/versions/5600a4bfe3c4_add_explicit_statefp_column_to_.py
```diff
@@ -0,0 +1,40 @@
...
+
+from alembic import op
+import sqlalchemy as sa
+
+from app.constants import SQL_DIR
+
+
+# revision identifiers, used by Alembic.
+revision: str = "5600a4bfe3c4"
+down_revision: Union[str, None] = "0db008690d60"
```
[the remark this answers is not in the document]
nofurtherinformation: Updated in c4dcb9765625d7bdf7e44aa6e5ca397675857dab

backend/app/core/dependencies.py
```diff
@@ -110,6 +110,7 @@ def get_document_public(
         DistrictrMap.child_geo_unit_type.label("child_geo_unit_type"),  # pyright: ignore
         DistrictrMap.data_source_name.label("data_source_name"),  # pyright: ignore
         DistrictrMap.comment.label("comment"),  # pyright: ignore
+        DistrictrMap.statefps.label("statefps"),  # pyright: ignore
         # get metadata as a json object
         Document.map_metadata.label("map_metadata"),  # pyright: ignore
```
peterrrock2: Why are there so many `# pyright: ignore`'s in this?

nofurtherinformation: Noted for future us to fix in #477

PR #476, opened 2026-01-27
Number of districts in a plan can be modified

nofurtherinformation:
Previously, the number of districts in a given plan have been locked to the map module. This means that the number of districts is always static once a plan is made. There has been a feature planned for a while to open this up and make it more flexible, and now we finally do need it for a specific context -- the Navajo Nation plans could have 24 or 25 districts.

## Description
- Moves number of districts to the document schema
- Adds FE controls
- Refactor some map metadata controls (eg colors) to follow on-save pattern

## Reviewers
- Primary:
- Secondary:

## Checklist
- [x] BE changes
- [x] FE changes
- [x] Tests

### DIFF
--- a/app/src/app/utils/api/apiHandlers/putUpdateAssignmentsAndVerify.ts
+++ b/app/src/app/utils/api/apiHandlers/putUpdateAssignmentsAndVerify.ts
@@ -48,7 +48,10 @@ export const putUpdateAssignmentsAndVerify = async ({
     // TODO: Have metadata confirmed after put and make it possible to update metadata without assignments
     metadata: {
       color_scheme: mapDocument.color_scheme ?? undefined,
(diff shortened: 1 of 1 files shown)


app/src/app/components/Toolbar/ZonePicker.tsx
```diff
@@ -1,22 +1,77 @@
...
+          <Button
+            variant="soft"
+            size="1"
+            onClick={handleDecreaseDistricts}
+            disabled={isReadOnly || numDistricts <= 2}
+          >
+            <MinusIcon />
+          </Button>
+          <Text size="2" weight="bold" style={{minWidth: '2ch', textAlign: 'center'}}>
+            {numDistricts}
```
peterrrock2: Handle text enter. Also limit to 538

peterrrock2 (commented):
Update to double buffer redraw on restore

PR #492, opened 2026-02-16
Separate out drawing responsibility of different layers so the zone layer doesn't handle everything

peterrrock2:
This refactor separates map rendering responsibilities by source/layer type, makes polygon draw order explicit/deterministic, and splits main vs demographic map composition to reduce branching and make layer behavior easier to reason about.

**What Changed**
- Replaced the single `MapComponent(isDemographicMap?)` pattern with:
  - `MainMapComponent`
  - `DemographicMapComponent`
  - Shared `MapShell` for common map plumbing.
- Added deterministic draw-order anchors via `MapLayerAnchors` and `MAP_LAYER_ANCHORS`.
- Introduced explicit polygon order contracts in `Map.tsx` using:
  - `MAIN_BLOCK_LAYER_ORDER`
  - `DEMOGRAPHIC_BLOCK_LAYER_ORDER`
- Split map code into single-idea folders based on purpose:
  - `GeoSources/`
  - `PolygonLayers/`
  - `PointLayers/`
- Replaced monolithic layer components:
  - Removed old `VtdBlockLayers`, old top-level `ZoneLayerGroup`, and `HighlightOverlayLayerGroup`.
  - Added focused polygon components:
- Added shared layer modules:
  - `layerConstants.ts` (outline IDs, unassigned background opacities)
  - `layerContracts.ts` (shared `BlockLayerOrder` contracts)
- Updated map page mounting to use explicit main/demographic components.
- Updated `getLayerFill` signature (removed `child` arg) and updated call sites.
- Updated rendering logic so assignment fill/background behavior is more explicit:
  - Assignment fill is only visible where `feature-state.zone` exists.
  - Geometry background fills unassigned geometry and is suppressed for assigned features.
- Updated point-data fetch source usage to `BLOCK_SOURCE_ID` in `usePointData`.


app/src/app/components/Map/PolygonLayers/layerConstants.ts
```diff
@@ -0,0 +1,9 @@
+export const GEOMETRY_OUTLINE_LAYER_IDS = {
```
nofurtherinformation: H: A lot of this kind of stuff lives in [constants/layers.ts](https://github.com/districtr/districtr-v2/blob/b5b484cf49686e89942433dbf8b91e5062c8f9e3/app/src/app/constants/layers.ts#L4) -- like we discussed, let's make a folder of map constants, and have a few files there like `constants/map/layerIds.ts`, `constants/map/layerStyling.ts` ...

app/src/app/components/Map/PolygonLayers/OverlayLayers.tsx
```diff
@@ -198,24 +197,21 @@ const OverlayLayer = ({
...
+        {selectedOutlineProps && <Layer {...selectedOutlineProps} />}
+        {/* Invisible click layer for line overlays */}
+        {clickLayerProps && <Layer {...clickLayerProps} />}
+      </Source>
     </>
   );
 };
 
-export const OverlayLayers = () => {
+export const OverlayLayers = ({layerBeforeId}: {layerBeforeId: string}) => {
```
nofurtherinformation, line 170: PP: Slight preference for `export const OverlayLayers: React.FC<Props> = ...`

app/src/app/components/Map/PolygonLayers/ChildBlockLayers.tsx
```diff
@@ -0,0 +1,108 @@
...
+import {useLayerFilter} from '@/app/hooks/useLayerFilter';
+import GeometryOutlineLayer from './GeometryOutlineLayer';
+import {GeometryBackgroundLayer} from './GeometryBackgroundLayer';
+import {ZoneLayerGroup} from './ZoneLayers/ZoneLayerGroup';
+import {HoverLayerGroup} from './HoverLayerGroup';
+import {GEOMETRY_OUTLINE_LAYER_IDS, UNASSIGNED_BACKGROUND_OPACITY} from './layerConstants';
+import {DemographicLayer} from './DemographicLayer';
+import type {BlockLayerOrder} from './layerContracts';
+
+export function ChildBlockLayers({layerOrder}: {layerOrder: BlockLayerOrder}) {
```
nofurtherinformation: H: From our chat today, let's try combining `ChildBlockLayers` and `ParentBlockLayers` into a single component, with the canonical layer IDs organized under `constants/map/layerIds.ts` (or similar) with a dict like
```ts
export const CANONICAL_LAYER_IDS = {
  CHILD: {
    BACKGROUND: ...
  }
} as const
```

Which the component for both can access directly `const layerIds = child ? CANONICAL_LAYER_IDS.CHILD : CANONICAL_LAYER_IDS.PARENT` or be passed in as a component property

app/src/app/components/Map/PolygonLayers/DemographyColorController.tsx
```diff
@@ -0,0 +1,69 @@
+'use client';
+import {BLOCK_SOURCE_ID} from '@/app/constants/layers';
+import {useAssignmentsStore} from '@/app/store/assignmentsStore';
+import {useDemographyStore} from '@/app/store/demography/demographyStore';
+import {useMapStore} from '@/app/store/mapStore';
+import {demographyCache} from '@/app/utils/demography/demographyCache';
+import {useLayoutEffect} from 'react';
+import {useMap} from 'react-map-gl/maplibre';
+
+export function DemographyColorController({enabled}: {enabled: boolean}) {
```
nofurtherinformation: H: Let's move this into `mapRenderSubs.ts` -- the logic here should be mostly the same, but we'll need a new render method like `renderDemographyColors` and a new subscription that matches the demography store state we need to listen to.

app/src/app/components/Map/PolygonLayers/GeometryBackgroundLayer.tsx
```diff
@@ -0,0 +1,47 @@
...
+  sourceLayerId: string;
+  filter: FilterSpecification;
+  beforeId: string;
+  style?: {
+    backgroundOpacity?: number;
+  };
+}) {
+  const backgroundOpacity = style?.backgroundOpacity ?? 0.2;
+
+  const fillOpacity = useMemo(
```
nofurtherinformation: Minor: This could be combined with the filter for slightly faster rendering:

```ts
  return (
    <Layer
      filter={[
        'case', 
          ['all', 
            ['!', ['==', ['feature-state', 'zone'], null]], // Hide if has feature-state
            filter // from props
          ]
      ]}
  )
  ...

If maplibre is smart, this _should_ return a falsy value fast on that first condition

File: app/src/app/components/Map/PolygonLayers/HoverLayerGroup.tsx [hunk 6]
```diff
@@ -0,0 +1,55 @@
...
+}: {
+  idBase: string;
+  sourceLayerId: string;
+  filter: FilterSpecification;
+  layerBeforeId: string;
+}) {
+  const isOverlay = useMapControlsStore(state => state.mapOptions.showDemographicMap) === 'overlay';
+  const fillOpacity = isOverlay ? 0.3 : 0.1;
+  const fillId = `${idBase}_demography_hover`;
+  const lineId = `${idBase}_line`;
```
nofurtherinformation: PP: If we are centralizing IDs in a new constants file, it may make sense to handle these as well!

app/src/app/components/Map/Map.tsx
```diff
@@ -4,105 +4,105 @@ import maplibregl from 'maplibre-gl';
...
+ * - Enforce common interaction constraints (no rotate/pitch) and shared interactive layers.
+ *
+ * Responsibilities intentionally not owned here:
+ * - Protocol registration (e.g. pmtiles), map syncing, or store-specific map-ref setup.
+ * - Variant-specific load behavior and side effects.
+ * - Layer selection/composition and ordering.
+ *
+ * Those variant concerns are injected by callers via `onMapLoad` and `children`.
+ */
+function MapShell({
```
nofurtherinformation: PP: `MapContainer` or `MapOuter` might be more descriptive

app/src/app/components/Map/Map.tsx
```diff
@@ -167,16 +157,135 @@ export const MapComponent: React.FC<{isDemographicMap?: boolean}> = ({isDemograp
...
+        <PointSelectionLayer />
+        <PointSelectionLayer child />
+        <MetaLayers isDemographicMap={false} />
+      </PointSource>
+      <NavigationControl showCompass={false} showZoom={true} position="bottom-right" />
+    </MapShell>
+  );
+}
+
+export function DemographicMapComponent() {
```
nofurtherinformation: Minor: For code-org, we should move the components in this file into separate files at some point

app/src/app/components/Map/Map.tsx
```diff
@@ -4,105 +4,105 @@ import maplibregl from 'maplibre-gl';
...
+import {ChildBlockLayers, DemographicChildBlockLayers} from './PolygonLayers/ChildBlockLayers';
+import {DemographyColorController} from './PolygonLayers/DemographyColorController';
+import type {ParentChildBlockLayerOrder} from './PolygonLayers/layerContracts';
 
+/**
+ * Global z-order from top -> bottom:
+ * labels -> hover -> overlays -> geometry-outline -> demography -> assignments -> counties.
+ * Anchors are mounted by <MapLayerAnchors /> and all map layers target these ids via beforeId.
+ */
+const MAIN_BLOCK_LAYER_ORDER: ParentChildBlockLayerOrder = {
```
nofurtherinformation: PP: This would be better in constants too imo

nofurtherinformation (changes requested):
Thank you for all your work on this!

Great stuff and a much needed refactor. I've added comments with the things we chatted about, mostly code org. 

Cheers!

app/src/app/components/Map/Map.tsx
peterrrock2: Done.

app/src/app/components/Map/PolygonLayers/ChildBlockLayers.tsx
peterrrock2: Okay, I did this, but I also tried to unify a bunch of stuff into a central dictionary. I then put some convenience exports into the file that map directly onto old code so that we don't have a bunch of constants changing elsewhere. However, the single source of truth for layer IDs should now be the `CANONICAL_LAYER_IDS` struct

app/src/app/components/Map/PolygonLayers/GeometryBackgroundLayer.tsx
peterrrock2: So I tried this, and I couldn't get it to work. It looks like 'feature-state' is not supported in geometry filters according to [https://maplibre.org/maplibre-style-spec/layers/#filter](https://maplibre.org/maplibre-style-spec/layers/#filter). 

I did manage to simplify the layer a bit with

```ts
    <Layer
      id={id}
      source={BLOCK_SOURCE_ID}
      source-layer={sourceLayerId}
      filter={filter}
      beforeId={beforeId}
      type="fill"
      layout={{visibility: 'visible'}}
      paint={{
        'fill-opacity': [
          'case',
          // Coalesce to -999 since none of our features will ever be assigned to that
          ['!=', ['coalesce', ['feature-state', 'zone'], -999], -999],
          0,
          backgroundOpacity,
        ],
        'fill-color': '#cecece',
      }}
    />
```

nofurtherinformation: Ah, bummer!

nofurtherinformation: This looks good though. 

Small thing - can we move `-999` to be an explicit sentenel value like used in `constants/map/layerStyle.ts`
```
// Sometimes, maplibre requires non-empty arrays for specifications
// This serves as a placeholder for otherwise empty arrays
export const SENTINEL_EMPTY_VALUE: string = '-999';
export const SENTINEL_EMPTY_ARRAY: string[] = [SENTINEL_EMPTY_VALUE];
```

peterrrock2: Can do!

peterrrock2: I'll resolve this when I fix that

nofurtherinformation: Ty!

nofurtherinformation (approved):
LGTM :)

nofurtherinformation (approved):
LGTMx2 :)

PR #497, opened 2026-02-19
Feat/community COI mode map page and styles

nofurtherinformation:
This PR contains the routing for COI maps (`.../coi/{public id}` and `.../coi/edit/{uuid}`). Additonally, it stubs out some district mode / coi mode disambiguations and features.

## Description
- Creates new route for COI page
- Stubs out brush controls
- Update demography panel to match UX
- Adds map styling controls
- Adds geocoding search
- Adds maptiler dependency for Satellite tiles and geocode search

## Reviewers
- Primary: @peterrrock2 
- Secondary:


app/src/app/store/mapControlsStore.tsx
```diff
@@ -40,8 +41,20 @@ export interface MapControlsStore {
...
 
+const initialMapMode = typeof window !== 'undefined' && typeof window.location !== 'undefined'
+  ? (
+      window.location.pathname.startsWith('/map')
+        ? 'districts'
+        : window.location.pathname.startsWith('/coi')
+          ? 'coi'
+          : 'districts'
+    )
+  : 'districts'
```
peterrrock2: I feel like this could be made simpler
```suggestion
const pathname = typeof window !== "undefined" ? window.location?.pathname ?? "" : "";
const initialMapMode = pathname.startsWith("/coi") ? "coi" : "districts";
```

app/src/app/utils/api/constants.ts
```diff
@@ -11,3 +11,9 @@ export const PARQUET_URL =
   process.env.NEXT_PUBLIC_S3_BUCKET_URL_MIRROR2 ?? process.env.NEXT_PUBLIC_S3_BUCKET_URL;
 
 export const RECAPTCHA_SITE_KEY = process.env.NEXT_PUBLIC_RECAPTCHA_SITE_KEY ?? '';
+
+/** MapTiler API key for basemaps (Streets/Satellite) and geocoding. */
+export const MAPTILER_API_KEY =
+  typeof window === 'undefined'
+    ? process.env.NEXT_PUBLIC_MAPTILER_API_KEY ?? ''
+    : process.env.NEXT_PUBLIC_MAPTILER_API_KEY ?? '';
```
peterrrock2: What is this doing? I'm not sure I understand why there is a ternary operator here
```suggestion
export const MAPTILER_API_KEY = process.env.NEXT_PUBLIC_MAPTILER_API_KEY ?? "";
```

app/src/app/components/MapPage/CoiMapPage.tsx
```diff
@@ -0,0 +1,114 @@
...
+import {SaveConflictModal} from '../SaveConflictModal';
+import {migrateUserMapsFromLocalStorage} from '@/app/utils/idb/migrateUserMaps';
+import { DemographicMap } from '../Map/DemographicMap';
+
+interface CoiMapPageProps {
+  isEditing: boolean;
+  documentId: string;
+}
+
+function ChildCoiMapPage({isEditing, documentId}: CoiMapPageProps) {
```
peterrrock2: React.FC styling?

app/src/app/components/Toolbar/ZonePicker.tsx
```diff
@@ -120,3 +120,23 @@ export function ZonePicker() {
...
+      <Text>TODO: Coi Zone Picker</Text>
+    </Box>
+  )
+}
+
+export const ZonePicker: React.FC = () => {
+  const mapMode = useMapControlsStore(state => state.mapMode);
+  switch (mapMode) {
+    case 'districts':
+      return <ZonePicker />;
```
peterrrock2:
```suggestion
      return <DistrictsZonePicker />;
```

app/src/app/components/Toolbar/ZonePicker.tsx
```diff
@@ -120,3 +120,23 @@ export function ZonePicker() {
     </Box>
   );
 }
+
+export const CoiZonePicker = () => {
```
peterrrock2:
```suggestion
export const CoiZonePicker: React.FC = () => {
```

peterrrock2 (changes requested):
Looking good so far! This review is mostly to familiarize me with what is going on and to mark the big style files as viewed before subsequent review. I did find a couple of things that are probably worth changing

app/src/app/utils/api/constants.ts
nofurtherinformation: In some cases we'll have different env variables based on server or client environments but in this case yeah it should always be the same

app/src/app/components/MapPage/CoiMapPage.tsx
nofurtherinformation: Good catch!

nofurtherinformation: Resolved in 1819edd2bdca45ca7c2c49375df68744ab92e5e2

peterrrock2 (commented):
Marking new changes as viewed. This just added in the fancy map router since I last saw it 😁

PR #501, opened 2026-03-04
Fix prettier command

nofurtherinformation:
Fix prettier

Review required: Someone else run the pre-commit in docker


peterrrock2 (approved):
Works on my end

PR #504, opened 2026-03-05
Scrollable Population Panel

nofurtherinformation:
## What this does

For maps with more than 10 districts, the Population panel's district rows become scrollable while everything informative stays fixed: the lock-all control and the "Ideal <population>" label sit in a fixed header strip above the scroll area, and the value axis sits in a fixed footer strip below it. Maps with 10 or fewer districts render exactly as before, with no scrollbar.

## Design decision: no fade-out, prominent scrollbar

Earlier iterations used a fade-out ("vanishing rows") effect at the bottom of the scroll area to hint that more content exists below. 
image

We removed it because fading interacts badly with a bottom axis: the gradient, the axis band, and the scrollbar all compete for the same edge; the fade extends beyond the bottom axis. 

Instead of using fade to indicate more rows are hidden below, we decide to make **scrollbar  always visible** (`type="always"` on the Radix ScrollArea, size 2) whenever scrolling is active — it is the sole affordance for "more rows below." To reinforce it, the scroll viewport is capped at exactly **10.6 rows**, so the half-visible 11th row makes it more obvious about the hidden contents below.

## Alignment: correct by construction

The previous draft aligned the left column with the bars via hand-derived magic constants. These are now derived from chart-owned formulas:

- `getChartHeight(count, rowHeight)` — the panel sizes the chart container with the same function the chart uses internally, so bar spacing always equals row height.
- `getBarCenterY(marginTop, rowHeight)` — the left column's top spacer is computed from the bars' actual geometry instead of a hardcoded offset.
- `PopulationChart` now takes `rowHeight` instead of `height`, making row height the single input everything derives from.

## Also fixed along the way

- Ideal-population dashed line and target-deviation band were double-rendered (band appeared ~2× darker when a target deviation was set)
- Background tokens unified on `var(--color-background)`
- `svg` elements set to `display: block` to remove phantom inline layout gaps


app/src/app/components/sidebar/PopulationPanel/PopulationPanel.tsx
```diff
@@ -131,16 +141,21 @@ export const PopulationPanel = () => {
...
-            {!isCommunityMode && (
+      <Flex
+        direction="row"
+        gapX="1"
+        align="center"
+        justify="between"
+        minHeight={isCommunityMode ? '12px' : '28px'}
+        style={{width: '5rem'}}
+      >
+        {!isCommunityMode && (
```
fangge518, line 156: Community mode does not display population data.

fangge518 (commented):
Note: A light vertical misalignment between the lock and the population bar.

app/src/app/components/sidebar/PopulationPanel/PopulationPanel.tsx
fangge518: Not displaying population data is intended here, as community drawing does not depend on population as redistricting.

peterrrock2:
## Scrollable Pop Panel update

@nofurtherinformation 

### Notes:

1. This is looking good from the FE!

nofurtherinformation:
LGTM!

PR #522, opened 2026-03-25
Coalition builder fixes

nofurtherinformation:
- Rename "Demographics Table" to just "Table" in the headers. Right now it repetitively says "Demography" as the top level header and then "Demography Table" underneath
- Change "Universe" to "Statewide" in the label for overall population numbers


peterrrock2 (approved):
Looks good!

PR #534, opened 2026-04-24
Fix backend dev container port 5432 conflicts.

fangge518:
## Description

A minimal change to fix a problem that when using backend dev container, the host machine cannot access the database through port 5432.

### The problem
Docker compose's db specification already publishes port 5432 to the host. Backend's forwarding of backend's port 5432 (which is not db:5432) actually creates a conflict that prevents localhost:5432 from connecting to the database.

### The fix
Removing backend's forwarding of 5432 fix the problem.

## Reviewers

- Primary:
- Secondary:


nofurtherinformation (approved):
LGTM :)

PR #540, opened 2026-05-01
API server memory fix

### LINKED ISSUE #535: Bugfix: API machine (FastAPI) has a memory leak
I suspect something with the network graphs for continuity, but I'm not sure

Image

fangge518:
## Context
The API server's memory usage keeps growing to about 7GB. This is because `backend/app/main.py`'s `_get_graph()` is using a memory cache that never evicts, so it will retain the dual graph for all states in the worst case, which occupies >1 GB disk space in total as pickle packages.

## Description

- Change to use an LRU cache of size 10 for the dual graph.
- Add a debug endpoint to show the cache stats (hits/misses) and memory usage.

## Reviewers

- Secondary: @peterrrock2


backend/app/main.py
```diff
@@ -1544,3 +1548,32 @@ async def get_group(
         "name": group.name,
         "slug": group.slug,
     }
+
+
+@app.get("/debug/cache")
```
peterrrock2: Nit: Maybe mangle this?
```suggestion
@app.get("/__DISTRICTRDEBUG__/cache")
```

peterrrock2 (approved):
LGTM!

backend/app/main.py
fangge518: Thanks for the suggestion. After some research on API naming conventions, I've decided on a middle road of "/_debug/cache".

PR #541, opened 2026-05-13
Implement metrics computation at backend

fangge518:
## Description

**New evaluation metrics**
- County splits (via `CountyContext`, replacing `IdealsForEguia`, which now also surfaces county population data)
- Block cut edges, Polsby-Popper, Reock, vote share
- Basic map-validity metrics: contiguity, assigned/unassigned units, unassigned population
- Note: demographic scores are out of scope by design

**Graph architecture overhaul**
- Replaced separate parent-level and child-level dual graphs with a single combined hybrid-level graph (block nodes + parent nodes + cross-level edges + weighted edge counts), simplifying storage, testing, and computation
- Moved graph generation entirely out of backend ingestion into the pipeline — graphs are now built from GeoPackage spatial joins (no DB access) and uploaded to S3; backend downloads on demand

**Testing**
- Fixture-based unit tests for all of the above
- FL integration test against the 2026 congressional plan (Moon Duchin, May 10, 2026)

## Reviewers

- Primary: @peterrrock2 

## Checklist

### 1. Backfill `parent_geo_unit_type` / `child_geo_unit_type` for existing maps

Run once against the production DB inside the backend container. This uses
`infer_geo_unit_type` (samples a `path` from `gerrydb.{layer}`) and
`update_districtrmap` to fill in NULL fields for maps created before this PR:

```python
from sqlmodel import select
from app.main import get_session
from app.models import DistrictrMap
from app.utils import infer_geo_unit_type, update_districtrmap

session = next(get_session())
maps = session.exec(select(DistrictrMap).where(DistrictrMap.parent_geo_unit_type == None)).all()
for m in maps:
    try:
        update_districtrmap(session, m.districtr_map_slug,
            gerrydb_table_name=m.gerrydb_table_name,
            parent_geo_unit_type=infer_geo_unit_type(session, m.parent_layer),
            child_geo_unit_type=infer_geo_unit_type(session, m.child_layer) if m.child_layer else None,
        )
        session.commit()
    except Exception as e:
        session.rollback()
        print(f"Failed {m.districtr_map_slug}: {e}")
```

### 2A. If S3 is not yet updated (will do so when PR is merged), build and upload graph `.pkl` files to S3

Graph files are now built by the **pipeline** (no DB access required) using spatial
joins on GeoPackage files. The pipeline produces one combined dual-level `.pkl` per
shatterable map (block nodes + parent nodes + cross-level edges + weighted edge counts)
and one plain adjacency `.pkl` per non-shatterable map.

Run from the `pipelines/` directory (requires GeoPackage files accessible from S3 or locally):

**Shatterable maps (v1 states):**
```bash
python cli.py transforms batch-create-graphs \
  --config-path transforms/configs/v1_graph_config.yml \
  --upload
```

**Shatterable maps (older non-v1 maps):**
```bash
python cli.py transforms batch-create-graphs \
  --config-path transforms/configs/vap_and_election.yml \
  --upload
```

GeoPackage files are read from S3 automatically if not present locally. Graphs are
uploaded to `s3://<S3_BUCKET>/graphs/<gerrydb_table_name>.pkl`.

The backend downloads graph files on demand from S3 if not present on the local volume.

Or... 
### 3. Update the data source column in districtrmap table
```
psql $DATABASE_URL -c "UPDATE districtrmap SET data_source_name = '2020 Decennial Census' WHERE data_source_name IS NULL OR data_source_name = '';"
```


backend/app/evaluation/context.py
peterrrock2, line 1: update to have the vtd-vtd block edge weights stored on the disk pkl files and update the ingestion logic

backend/app/evaluation/compactness.py
```diff
@@ -0,0 +1,150 @@
...
+logger = logging.getLogger(__name__)
+
+class CutEdgesResult(TypedDict):
+    cut_count: int
+    unit_type: str
+
+
+def _infer_unit_type(geo_id: str) -> str:
+    """Infer the geographic unit type from one geo_id."""
+    return "block" if ":" not in geo_id else geo_id.split(":", 1)[0]
```
peterrrock2: I think that we nee to check this for block groups for places like WV and CA which do not have VTDs. GerryDB does not add any prefixes to their identifiers since they are a part of the central spine. Probably better to do something like 
```suggestion
def _infer_unit_type(geo_id: str) -> str:
    """Infer the geographic unit type from one geo_id."""
    if len(geo_id) == 15 and geo_id.isdigit():
        return "block"
    elif len(geo_id) == 12 and geo_id.is_digit():
        return "bg"
    elif geo_id.startswith("vtd"):
        return "vtd"
    else:
        raise ValueError(f"geo_id {geo_id} is of unknown type")
```

backend/app/evaluation/compactness.py
```diff
@@ -0,0 +1,150 @@
...
+
+    unit_to_zone: dict[str, int] = {}
+    parent_unit_to_zone: dict[str, int] = {}
+    for geo_id, zone in rows:
+        # For shatterable maps, colon-prefixed geo_ids are parent-layer assignments
+        # (e.g. "vtd:xxxxx"); bare IDs are individually-assigned child units.
+        # For non-shatterable maps the colon check is skipped: all assignments are
+        # atomic units regardless of path format.
+        if is_shatterable and ":" in geo_id:
+            parent_unit_to_zone[geo_id] = zone
```
peterrrock2: This probably has the bg bug as well.

backend/app/evaluation/compactness.py
peterrrock2, line 1: It might be worth having a helper function that computes the actual cut edges so that we can pass that around to the county pieces and county splits.

backend/app/evaluation/graph.py
```diff
@@ -0,0 +1,246 @@
...
+
+    if url.scheme == "s3":
+        logger.info("Downloading file from S3")
+        s3 = settings.get_s3_client()
+        assert s3, "S3 client is not available"
+        gpkg_path = download_file_from_s3(s3=s3, url=url, replace=False)
+        logger.info("Path: %s", gpkg_path)
+
+    conn = sqlite3.connect(gpkg_path)
+    cursor = conn.execute(f"SELECT path_1, path_2 FROM {layer_name}")
```
peterrrock2: nit: unsanitized SQL

backend/app/evaluation/splits.py
```diff
@@ -0,0 +1,52 @@
...
+    split pieces are the minimum number of pieces required to accommodate the population
+    of the county, given the ideal district population.  The "actual" split pieces are
+    the number of split pieces in the submitted plan. For unfinished districting plans,
+    the actual split pieces does not treat the unassigned area as a zone, and completely
+    unassigned counties will thus have a piece-count of 0.
+
+    The number of counties split into two or more pieces can be easily derived from this
+    mapping by counting the number of counties where `actual_split_pieces` is 2 or more.
+    """
+    county_pops: dict[CountyGeoid, int] = COUNTY_CONTEXT.county_populations(context.gerrydb_table, context.session)
```
peterrrock2:
```suggestion
    county_pops: dict[CountyGeoid, int] = COUNTY_CONTEXT.county_populations(context.parent_table, context.session)
```
Okay, it took me a second to trace this, but I think that there is a bug here. The table can be a materialized view, and will trip the `relkind != '"r"` guard in `_populate_county_data()` if the population is not already cached.

The test suite looks like it is pre-populating the cache every time, so I don't think that it is exercising this path. Can you write a quick test to see if I'm seeing this correctly?

peterrrock2 (changes requested):
Notes on optimizing the computation of `block_cut_edges` and `county_splits` and `county_pieces`:

It looks like both metrics currently issue the same
 
```
SELECT geo_id, zone FROM assignments WHERE document_id = X AND zone IS NOT NULL query
``` 
and iterate the full result independently. We can probably use a cached property / shared helper to make this quicker. I would make it something like `cut_edge_pair` which does the graph traversal and returns `(cut_edges, assigned_counties)` as a tuple.
- `cut_edges` would be a `frozenset[tuple[frozenset[tuple[str, int]], int]]`, so each element is (frozenset({(geo_id_a, zone_a), (geo_id_b, zone_b)}), weight).
- `assigned_counties` would be a `frozenset[str]` 

Then you can compute block_cut_edges by just summing the weights of the cut edges, and county pieces would be computed in the following way
1. Loop through all cut edges. As you loop
    - count (county, district) pairs (call the set `pairs`) ,
    - unique counties in (county, district) pairs (call it `counties_in_pairs`),
    - and counties with > 1 (county, district) pair in the cut edges  (this is county_splits)
2. Compute `county_pieces = total_counties - len(unassigned_counties) + len(pairs) - len(counties_in_pairs)` 

<strike>To keep memory usage low, we can then add a `context.cleanup()` method that would pop `cut_edge_pair` from `__dict__` when you finish computing metrics.</strike>

^ I'm a silly goose and the context does not live long enough for this to matter


---
## Other notes:

- We should probably add some integration tests for the metrics

Otherwise, this is looking good!

backend/app/evaluation/context.py
fangge518: This is done.

All the graph related logic has been centralized to graph.py

backend/app/evaluation/compactness.py
fangge518: Added. Moved this part to the deployment operation in utils.py to fill the parent_geo_unit_type or child_geo_unit_type column.

backend/app/evaluation/compactness.py
fangge518: Fixed. Also changed the way the two passes is carried out. Non-shatterable map will be dealt with in the first pass.

backend/app/evaluation/compactness.py
fangge518: As we discussed. county_pieces do not need significant time cost besides fetching the assignment array (if not cached). I'll leave the code as it is.

backend/app/evaluation/graph.py
fangge518: Added assert_safe_ident.

backend/app/evaluation/splits.py
fangge518: Great catch! I've added a unit test for this mistaken behavior.

backend/app/evaluation/context.py
```diff
@@ -91,48 +130,164 @@ def rep_wins(self) -> dict[Election, pd.Series]:
...
+    @cached_property
+    def is_shatterable(self) -> bool:
+        """Whether this map has a child (block) layer."""
+        return self.child_layer is not None
+
+    @cached_property
+    def parent_geo_unit_type(self) -> GeoUnitTypeName:
+        """Parent unit type (e.g. 'vtd', 'block'). Raises ValueError if unset in districtrmap"""
+        if not self._districtr_map.parent_geo_unit_type:
+            raise ValueError(f"Unknown parent_geo_unit_type '{self.parent_geo_unit_type}' for document '{self.document_id}'.")
```
nofurtherinformation: H: The error message accesses `self.parent_geo_unit_type`, which re-enters this same property, causing a `RecursionError`. Could be:

```
    raise ValueError(
        f"DistrictrMap for document '{self.document_id}' has no parent_geo_unit_type set."
    )
```

backend/app/contiguity/main.py
```diff
@@ -39,281 +24,204 @@ def subgraph_number_connected_components(
...
+    ymax: float
+
+
+def get_assigned_nodes_bboxes(
+    session: Session,
+    document_id: str,
+    districtr_map: DistrictrMap,
+    zone: int,
+    G: Graph | None = None,
+) -> list[NodeWithBBoxes]:
```
nofurtherinformation: I: This function can also return `None` - I'd update the return type or return an empty list for the `not rows` case

backend/app/evaluation/validity.py
nofurtherinformation, line 1: PP: I think this file/module could be named something more descriptive than `basic`. Maybe `summary` or `districts` or ??

backend/app/evaluation/partisans.py
```diff
@@ -71,11 +70,41 @@ def seats(context: DocumentEvaluationContext) -> dict[Election, dict[str, int]]:
...
+    """Per-election, statewide vote shares: `{election: {"dem": share, "rep": share}}`.
 
+    Formula:
+        V_dem(e) = sum(dem_votes(d, e)) / sum(dem_votes(d, e) + rep_votes(d, e))
+        V_rep(e) = sum(rep_votes(d, e)) / sum(dem_votes(d, e) + rep_votes(d, e))
+        where d ranges over districts and e over elections.
+    """
+    return {
+        election: {
+            "dem": context.dem_state_votes[election] / context.total_state_votes[election],
```
nofurtherinformation: Q: Is there ever a data case here where `context.total_state_votes[election]` ends up being `0` and should we put in a guard?

backend/app/evaluation/basic.py
```diff
@@ -0,0 +1,102 @@
...
+    assigned_count: int
+    partially_assigned_count: int
+    total_count: int
+    unit_type: GeoUnitTypeName
+
+
+def assigned_units(context: DocumentEvaluationContext) -> AssignedUnitsResult:
+    """Returns assigned and total parent-unit counts for the document's districting plan."""
+    unit_to_zone, parent_unit_to_zone = context.split_zone_assignments
+    G = get_graph(context.gerrydb_table)
```
nofurtherinformation: I: This graph is first used after two potential return paths, meaning `get_graph` could be too eager -- should we bump this down to ~`41`?

backend/app/evaluation/context.py
```diff
@@ -151,10 +306,81 @@ class IdealsForEguia:
     # the DB indefinitely for a permanently malformed gerrydb table.
     MAX_LOAD_ATTEMPTS: ClassVar[int] = 3
 
+    _DATA_DIR: ClassVar[Path] = Path(__file__).parent.parent / "data"
+    _COUNTY_NAMES_FILE: ClassVar[Path] = _DATA_DIR / "county_names.csv"
+    _COUNTY_NAMES_URL: ClassVar[str] = (
```
nofurtherinformation: H: I have a moderate preference to have this on our CDN / S3 bucket to avoid external data dependency. WDYT?

backend/app/evaluation/splits.py
```diff
@@ -0,0 +1,56 @@
...
+
+    The number of counties split into two or more pieces can be easily derived from this
+    mapping by counting the number of counties where `actual_split_pieces` is 2 or more.
+    """
+    county_pops: dict[CountyGeoid, int] = COUNTY_CONTEXT.county_populations(context.parent_layer, context.session)
+
+    county_zones: dict[CountyGeoid, set[int]] = {}
+    for geo_id, zone in context.zone_assignments:
+        bare_id = geo_id.split(":", 1)[1] if ":" in geo_id else geo_id
+        geoid = CountyGeoid(bare_id[:5])
```
nofurtherinformation: I: This geoid extraction happens here, at `52-53` and in `context.py` -- worth abstracting to it's own utility?

backend/app/evaluation/partisans.py
```diff
@@ -246,26 +262,32 @@ def competitive_metrics(context: DocumentEvaluationContext) -> CompetitiveMetric
...
+        return {
+            "n_dem_districts": 0,
+            "n_rep_districts": 0,
+            "n_swing_districts": 0,
+            "n_competitive_districts": 0,
+            "n_districts": n_districts,
+            "n_elections": 0,
+        }
     dem_districts = [1] * n_districts
     rep_districts = [1] * n_districts
```
nofurtherinformation: PP: Not touched in this PR, but `np.ones` might be more elegant

nofurtherinformation (changes requested):
Great work, Ge! This is a really nicely principled and structured refactor / enhancement. I've got a few minor optimizations / clean ups suggested, but no correctness issues so far as I can tell.

Excited to see this on dev and hoping that our Fly instances can handle it :D

backend/app/evaluation/context.py
peterrrock2: I agree!

backend/app/evaluation/graph.py
```diff
@@ -219,28 +222,83 @@ def annotate_graph_with_parents(
         if child_path in G.nodes:
             G.nodes[child_path]["parent"] = parent_path
     logger.info("Annotated %d parent edges for map %r", len(rows), map_uuid)
-    return G
 
 
-def build_parent_adjacency(G: Graph) -> Graph:
-    """Build a weighted parent-unit adjacency graph from an annotated block graph.
+def _build_combined_graph(G: Graph) -> None:
```
peterrrock2: passes child graph

backend/app/evaluation/graph.py
```diff
@@ -219,28 +222,83 @@ def annotate_graph_with_parents(
...
-    for u, v in G.edges():
-        parent_u = G.nodes[u].get("parent")
-        parent_v = G.nodes[v].get("parent")
-        if parent_u is None or parent_v is None or parent_u == parent_v:
+    G.graph["weighted_edges"] = {}
+
+    for u, v in list(G.edges()):
+        p_u = G.nodes[u].get("parent")
+        p_v = G.nodes[v].get("parent")
+        if p_u is None or p_v is None:
```
peterrrock2: Error for children with no parents

backend/app/evaluation/partisans.py
peterrrock2: Nope!

backend/app/evaluation/splits.py
peterrrock2: Yeah, probably worth it

backend/app/contiguity/main.py
peterrrock2, line 1: I: Change to bind param whenever possible

backend/app/evaluation/basic.py
fangge518: Good call. Bumping it down.

backend/app/evaluation/validity.py
fangge518: Agreed. I've renamed it as "validity", as they all function to show whether a map is finished in a valid way or not.

backend/app/evaluation/context.py
fangge518: Indeed. This is very problematic. Thanks for catching it.

backend/app/evaluation/context.py
fangge518: Agreed. Changed to S3.

backend/app/contiguity/main.py
fangge518: Agreed!

backend/app/evaluation/splits.py
fangge518: Agreed and refactored.

backend/app/evaluation/partisans.py
fangge518: Done. Thanks for making this more elegant.

backend/app/evaluation/graph.py
fangge518: Done.

backend/app/evaluation/graph.py
fangge518: use ["parent"] to let it raise naturally.

backend/app/contiguity/main.py
fangge518: I've double checked, and we are already using bind params when possible.

zone_filter, geo_source are short hands for constructing variable commands for different conditional paths.
map_uuid and safe_table is used to construct a table name, not a value. And it is internally provided.

I'll leave the code as it is.

pipelines/transforms/graph.py
```diff
@@ -0,0 +1,289 @@
...
+        ``non_contiguous_parents``: set of parent-unit paths whose child blocks form
+        more than one connected component (e.g. island precincts). These are expanded
+        to their block children during contiguity evaluation so that geographic
+        disconnection is not hidden by the single-node representation.
+    """
+    G.graph["weighted_edges"] = {}
+
+    for u, v in list(G.edges()):
+        p_u = G.nodes[u]["parent"]
+        p_v = G.nodes[v]["parent"]
```
peterrrock2, line 173: I: This will crash if the parent is never annotated. The function `_annotate_graph_with_parents_from_gpkg` can leave blocks unannotated, which will show up in the logging on 131. Probably good to crash here since we expect blocks to nest perfectly, but we should probably handle the failure state.

pipelines/transforms/graph.py
peterrrock2, line 1: I think that we need to make some tests for these so that we can see what happens if the geopackages contain misaligned units. Theoretically, all of the data that I hand you will be aligned, but we shouldn't trust that.

peterrrock2: Add an issue to document the need to add these in another PR (priority)

peterrrock2 (approved):
Note: It looks like we have an R2 bucket setting in backend/app/core/config.py that reads our graphs and then another in pipeline/core/settings.py that writes to an S3 bucket. Just checking that this is intentional?

TODO: Unify these

pipelines/transforms/graph.py
fangge518: An issue is created to track this effort.

pipelines/transforms/graph.py
fangge518: I've added a raise that will dump all the unannotated children.

fangge518:
All the deployments steps have been finished against the dev's db and S3.

nofurtherinformation (approved):
LGTM!

PR #542, opened 2026-05-14
Fix the eguia computation to use parent layer instead

fangge518:
## Description

Currently, eguia computation use the `gerrydb_table_name` from the `districtrmap` table for aggregating county demographics. For shatterable map which integrates the parent and the child layers, it will result in double the population counts. This fix switch to used the parent layer for county demographic instead.

## Reviewers

- Primary: @peterrrock2


peterrrock2 (commented):
This fix looks good to me! Can we add in a simple regression test that would've caught the bug? I think that the KS test fixture will probably work for this.

PR #543, opened 2026-05-28
Add timeouts for backend db requests. Log long requests.

fangge518:
## Description
Occationally the API server has long hanging requests, blocking all the other requests from creating new db connection. This fix introduces db timeouts so that these requests won't run indefinitely. Also logs the long running requests for future diagnosis.

## Reviewers

- Secondary:


nofurtherinformation (approved):
LGTM!

PR #545, opened 2026-06-01
Fix/bg task connection leak

nofurtherinformation:
## Problem

After a couple days of normal activity the API degrades badly — query times climb from ~hundreds of ms to minutes, then requests start failing with:

```
sqlalchemy.exc.TimeoutError: QueuePool limit of size 5 overflow 10 reached, connection timed out, timeout 30.00
```

The error surfaces on reads like `GET /api/gerrydb/views`, but that endpoint is just the victim — whatever happens to be waiting when the pool is already starved. Not reproducible locally (needs days of sustained traffic; local restarts reset the state).

## Potential causes

Two independent issues, both in how recently-added code uses the DB connection. **(1) is the primary suspect for the production symptom; (2) is a real but likely minor contributor.**

### 1. Per-request temp tables are never dropped (primary)

The assignment-write paths (`PUT /api/assignments` → `update_assignments`, and `batch_insert_assignments`) bulk-load via `COPY` into a per-request temp table (`CREATE TEMP TABLE temp_assignments_{uuid} ...`) and never drop it. `CREATE TEMP TABLE` defaults to `ON COMMIT PRESERVE ROWS`, so after the request commits the table lives for the life of the **pooled connection** — pooled connections aren't closed between requests, and SQLAlchemy's reset-on-return is a `ROLLBACK`, which doesn't drop a *committed* temp table. On the hot save path this churns the system catalogs and bloats `pg_class`/`pg_attribute` over days, slowing the **planning** phase of every query until connections are held long enough to exhaust the pool.

Fits the symptoms: gradual over days (bloat builds up), "ms → minutes" (slow planning, not one slow query), pool exhaustion as a downstream effect, invisible locally.

### 2. Background tasks reuse the request-scoped session (secondary)

Thumbnail generation and comment moderation were scheduled with `add_task(..., session=session)`, handing the task the request session that `get_session()` closes at teardown. The task then checks out a fresh pooled connection; the thumbnail path opens a transaction via `session.connection()` and never commits/closes it, leaking that connection. These endpoints are rarely hit (thumbnails especially), so it's a slow, minor leak — but a real bug regardless (the codebase already had `moderate_comment_by_id` as the correct self-owning pattern).

## Changes

**Temp tables → `ON COMMIT DROP`** — each temp table now drops when its transaction ends (covers commit *and* error paths), with no accumulation:
- `backend/app/main.py` — `update_assignments`: `temp_assignments_{id}`
- `backend/app/assignments/assignments.py` — `batch_insert_assignments`: `temp_assignments_{id}`, `uniform_vtds_{id}` (the temp index drops with its table)

Safe because each CREATE → COPY → INSERT runs in a single transaction with no intermediate commit, so the table is present for the INSERT and only drops at the end.

**Background tasks own their session** — `generate_thumbnail`/`generate_blank` and the `moderate_*` functions open their own `with Session(engine)` when run as a background task (mirroring `moderate_comment_by_id`); call sites no longer pass the request session.
- `backend/app/thumbnails/main.py`, `backend/app/comments/moderation.py`, `backend/app/comments/main.py`, `backend/app/utils.py`

**Backstop** — `backend/app/core/db.py` sets `idle_in_transaction_session_timeout = '60s'` so Postgres reclaims any future connection left idle in an open transaction instead of leaking it.

### DIFF
--- a/backend/app/comments/moderation.py
+++ b/backend/app/comments/moderation.py
@@ -110,37 +110,49 @@ def update_moderation_score(
         raise
 
 
-def moderate_comment(comment: Comment, session: Session):
+def _moderate(cls: Table, key: int, text: str, session: Session | None) -> None:
+    """Score ``text`` and persist the result for ``cls``/``key``.
+
+    When ``session`` is None (the background-task case) a dedicated session is
+    opened and closed here. Background tasks must not reuse the request-scoped
+    session: it is already closed by the time they run, so any connection they then
+    check out would never be returned to the pool (a leak).
+    """
+    if session is not None:
+        moderate_text(cls=cls, key=key, text=text, session=session)
+    else:
+        with Session(engine) as owned_session:
+            moderate_text(cls=cls, key=key, text=text, session=owned_session)
+
+
+def moderate_comment(comment: Comment, session: Session | None = None) -> None:
     comment_text = f"{comment.title} {comment.comment}"
-    moderate_text(cls=Comment, key=comment.id, text=comment_text, session=session)
+    _moderate(Comment, comment.id, comment_text, session)
 
 
-def moderate_comment_by_id(comment_id: int, comment_text: str):
+def moderate_comment_by_id(comment_id: int, comment_text: str) -> None:
     """
     Moderate a comment by ID. Use when the Comment object may be detached
     (e.g. in background tasks after request session is closed).
     """
-    with Session(engine) as session:
-        moderate_text(cls=Comment, key=comment_id, text=comment_text, session=session)
+    _moderate(Comment, comment_id, comment_text, None)
 
 
-def moderate_commenter(commenter: Commenter, session: Session):
-    commenter_data = str(commenter)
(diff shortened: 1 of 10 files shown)


backend/app/comments/moderation.py
```diff
@@ -110,37 +110,49 @@ def update_moderation_score(
...
 
 
-def moderate_comment(comment: Comment, session: Session):
+def _moderate(cls: Table, key: int, text: str, session: Session | None) -> None:
+    """Score ``text`` and persist the result for ``cls``/``key``.
+
+    When ``session`` is None (the background-task case) a dedicated session is
+    opened and closed here. Background tasks must not reuse the request-scoped
+    session: it is already closed by the time they run, so any connection they then
+    check out would never be returned to the pool (a leak).
```
fangge518, line 119: This is a really good catch and very subtle!

fangge518 (approved):
LGTM.

PR #547, opened 2026-06-03
Frontend Eval page

fangge518:
## Description
Creates the evaluation page for a public map on the frontend.

### Backfill statefps to the prod database
This is needed for the county boundary toggle to work properly.

## Reviewers
- Primary:


app/public/al_district1.png
nofurtherinformation, line 1: PP: We could give these PNGs a transparent background for more flexible layouts in the future. 

Also, a micro optimization we could do: use tinypng.com or similar to compress the images further, likely ~75% smaller file sizes

app/src/app/components/EvalPanel/BasicsSection.tsx
```diff
@@ -0,0 +1,183 @@
...
+import {InfoCircledIcon, TriangleRightIcon} from '@radix-ui/react-icons';
+import {useMapStore} from '@store/mapStore';
+import {DocumentEvaluation} from '@utils/api/apiHandlers/getEvaluation';
+import {SubsectionHeading, useDistrictHover} from './shared';
+
+interface Props {
+  evaluation: DocumentEvaluation;
+}
+
+const GEO_UNIT_LABELS: Record<string, string> = {
```
nofurtherinformation: I: Should we move these to constants or keep them colocated here?

app/src/app/components/EvalPanel/BasicsSection.tsx
```diff
@@ -0,0 +1,183 @@
+'use client';
+import * as Accordion from '@radix-ui/react-accordion';
+import {Flex, Text, Heading, Callout} from '@radix-ui/themes';
+import {InfoCircledIcon, TriangleRightIcon} from '@radix-ui/react-icons';
+import {useMapStore} from '@store/mapStore';
+import {DocumentEvaluation} from '@utils/api/apiHandlers/getEvaluation';
+import {SubsectionHeading, useDistrictHover} from './shared';
+
+interface Props {
```
nofurtherinformation: O: We could name this more descriptively even though only used in this file

app/src/app/components/EvalPanel/BasicsSection.tsx
```diff
@@ -0,0 +1,183 @@
...
+  block: 'blocks',
+};
+
+const GEO_UNIT_DESCRIPTIONS: Record<string, string> = {
+  vtd: 'VTDs, also called "voting tabulation districts" or "voting districts," are the closest approximation of electoral precincts in Census geography.',
+  bg: 'Block groups are Census geographic units that nest within counties and tracts, typically containing 600–3,000 people.',
+  block: 'Census blocks are the smallest Census geographic unit, corresponding roughly to city blocks.',
+};
+
+export function BasicsSection({evaluation}: Props) {
```
nofurtherinformation: O: I prefer `export const BasicsSection: React.FC<BasicsSectionProps>` for more explicit/cleaner typing, and this would be internally consistent in the codebase. 

My preference for this comes from subtle differences in function declarations and const functions, and generally being an old man yelling at clouds apropos react conventions.

However, this is non-blocking :)

app/src/app/components/EvalPanel/CompactnessSection.tsx
```diff
@@ -0,0 +1,227 @@
...
+  );
+}
+
+function PerDistrictTable({
+  polsby_popper,
+  reock,
+}: {
+  polsby_popper: Record<string, number>;
+  reock: Record<string, number>;
+}) {
```
nofurtherinformation: PP: I generally lean towards breaking out components into separate files for better modularity. 

Also, I'd prefer separating out props into their own interfaces or types and again lean towards `React.FC` syntax

app/src/app/components/Map/PolygonLayers/CountyLayers.tsx
```diff
@@ -68,6 +69,21 @@ export const CountyLayers = ({layerBeforeId}: {layerBeforeId: string}) => {
...
+          source-layer="tl_2023_us_county"
+          paint={{
+            'line-color': HIGHLIGHT_LINE_COLOR,
+            'line-width': HIGHLIGHT_LINE_WIDTH,
+            'line-opacity': 0.2,
+          }}
+          filter={
+            hoveredCountyGeoid
+              ? ['==', ['get', 'GEOID'], hoveredCountyGeoid]
+              : ['==', ['get', 'GEOID'], '']
```
nofurtherinformation: I: We'd be more explicit and consistent using `SENTINEL_EMPTY_VALUE` from `.../constants/map/layerStyle.ts:11`

app/src/app/components/EvalPanel/CompactnessSection.tsx
```diff
@@ -0,0 +1,227 @@
...
+          {/* Polsby-Popper */}
+          <SubsectionHeading>Polsby-Popper Scores</SubsectionHeading>
+          <Text size="2" as="p" mb="2">
+            The <strong>Polsby-Popper score</strong> compares a district&apos;s area to its
+            perimeter. Scores range from 0 to 1; higher scores indicate more compact districts.
+            Unlike cut edges, this measure depends on map projection and boundary resolution rather
+            than the choice of geographic units.
+          </Text>
+          <ExemplarTable scoreKey="polsby_popper" />
+          {polsby_popper && (() => {
```
nofurtherinformation: H: We should really move this somewhere else, a few options being:
- Outside the return (template) of the function component, maybe in a `useMemo`
- A custom hook

The reason for this is to avoid excess logic in the component, and avoid extra work on re-render. Also IIFE's are not my preferred syntax if we can avoid them :)

app/src/app/components/EvalPanel/BasicsSection.tsx
```diff
@@ -0,0 +1,183 @@
...
+            equally into the specified number of districts. Population deviation of a plan is
+            measured as the largest amount that any district differs from ideal size. Legislative
+            plans should typically have individual deviations under ±5%, which ensures a
+            top-to-bottom deviation of under 10%. Congressional plans are typically more tightly
+            balanced.
+          </Text>
+          {population_deviation ? (
+            <Text size="2" as="p">
+              Your plan&apos;s most populous district is{' '}
+              <strong
```
nofurtherinformation: H: `strong` tags aren't traditional interactive elements, and we should use something more a11y friendly and semantic like a styled `button`

app/src/app/components/EvalPanel/CompactnessSection.tsx
```diff
@@ -0,0 +1,227 @@
...
+  polsby_popper: 'A Polsby-Popper score of',
+  reock: 'A Reock score of',
+};
+
+function ExemplarTable({scoreKey}: {scoreKey: 'polsby_popper' | 'reock'}) {
+  return (
+    <Table.Root size="1" mb="3">
+      <Table.Body>
+        <Table.Row>
+          <Table.RowHeaderCell>
```
nofurtherinformation: PP: We should probably make these regular cells, not headers

app/src/app/components/EvalPanel/shared.tsx
```diff
@@ -0,0 +1,77 @@
...
+}
+
+const TYPE_LABELS: Record<string, string> = {
+  pres: 'Pres',
+  gov: 'Gov',
+  sen: 'Sen',
+  ag: 'AG',
+};
+
+export function formatElectionKey(key: string): string {
```
nofurtherinformation: H: We should de-dupe these utils with the functions in `utils/number.ts` and make `formatElectionKey` somewhere in utils, too

app/src/app/components/EvalPanel/shared.tsx
nofurtherinformation, line 1: H: On code org, this kinda of miscellany file is something we should avoid. I'd break this up along the lines of:
- app/src/app/hooks/useDistrictHover.ts — move the hook; it's reusable beyond EvalPanel.
- app/src/app/utils/{numbers|formatters}.ts — formatElectionKey, formatPct, formatDecimal, TYPE_LABELS. Pure functions, .ts (not .tsx).
  - Inline SubsectionHeading and MetricRow into a small components.tsx if they're reused, or just inline at call sites if each is used only 2–3 times.

app/src/app/components/Map/PolygonLayers/CountyLayers.tsx
```diff
@@ -68,6 +69,21 @@ export const CountyLayers = ({layerBeforeId}: {layerBeforeId: string}) => {
           }}
           filter={countyFilter}
         />
+        <Layer
+          id={CANONICAL_LAYER_IDS.COUNTIES.HIGHLIGHT}
+          type="line"
```
nofurtherinformation, line 75: H: We can use `beforeId` to handle the layer order issue

app/src/app/components/MapPage/MapPage.tsx
```diff
@@ -94,10 +107,10 @@ function ChildMapPage({isEditing, mapId}: MapPageProps) {
   }
 
   return (
-    <div className="h-screen w-screen overflow-hidden flex justify-between p flex-col-reverse lg:flex-row-reverse landscape:flex-row-reverse">
-      <SidebarComponent />
-      <div className={`h-full relative w-full flex-1 flex flex-col lg:h-screen landscape:h-screen`}>
-        <Topbar />
+    <div className={`h-screen w-screen overflow-hidden flex justify-between p flex-col-reverse lg:flex-row-reverse landscape:flex-row-reverse${isEval ? ' eval-page-root' : ''}`}>
```
nofurtherinformation, line 111: PP: We could make this a cleaner wrapper component and simplify the child template logic -- potentially worth a later revisit but maybe add a TODO to clean up this return at some point

app/src/app/globals.css
```diff
@@ -75,3 +75,56 @@ body {
     url(/break.png) 12 12,
     pointer;
 }
+
+.county-detail-scroll {
+  max-height: 400px;
+  overflow-y: auto;
+}
+
+@media print {
```
nofurtherinformation, line 79: Q: Is this required to be vanilla CSS or could we use `styled` from `stitches`?

nofurtherinformation (changes requested):
Awesome work! Overall looks great and works well!

I picked up on a few mostly code-org and some minor runtime enhancements, but overall looks pretty close to good to go :)

app/public/al_district1.png
fangge518: These are really helpful tips. Thank you!

The pictures have been changed.

app/src/app/components/EvalPanel/BasicsSection.tsx
fangge518: I've moved GEO_UNIT_LABELS, as well redefined the string literals in constants/document/geoUnits.ts. The Descriptions are left in this document.

app/src/app/components/EvalPanel/BasicsSection.tsx
fangge518: Thanks for reminding me about this. I recall reading about this earlier, so I'm all for this style. Plus it's more consistent with the other code in the repo.

app/src/app/components/EvalPanel/CompactnessSection.tsx
fangge518: Absolutely. These are all wrapped with useMemo now.

app/src/app/components/EvalPanel/CompactnessSection.tsx
fangge518: The syntax is updated.

I feel PerDistrictTable is too small to be its own file. I've inlined it, which also makes the presentation flow clearer.

app/src/app/components/EvalPanel/shared.tsx
fangge518: All suggestions adopted. Now the code seem much better organized.

app/src/app/components/EvalPanel/shared.tsx
fangge518: Done.

app/src/app/components/Map/PolygonLayers/CountyLayers.tsx
fangge518: Done. Now the highlights are shown properly. Also removed previous hacks on highlight colors.

app/src/app/components/Map/PolygonLayers/CountyLayers.tsx
fangge518: Good point! Done.

app/src/app/components/EvalPanel/BasicsSection.tsx
fangge518: Done.

app/src/app/components/EvalPanel/CompactnessSection.tsx
fangge518: Agreed and Done.

app/src/app/components/EvalPanel/BasicsSection.tsx
fangge518: Agreed and Done.

app/src/app/components/MapPage/MapPage.tsx
fangge518: I've added a TODO.

app/src/app/globals.css
fangge518: The answer is yes. 

`.county-detail-scroll` was scoped to a single `<div>` in `CountySplitsSection` and removed the class from globals.css entirely.

Claude prefers to keep the `@media print` block in globals.css and I agree: "Those rules span multiple components (MapPage, EvalPanel) and are best read as a single cross-cutting print layout rather than scattered across files."

nofurtherinformation (approved):
LGTM :)

PR #548, opened 2026-06-04
Change the timeout setting's trigger mechanism.

fangge518:
## Description
DB operation timeout is not triggered during db shortage. Fix the trigger mechanism on api server.

### DIFF
--- a/backend/app/core/db.py
+++ b/backend/app/core/db.py
@@ -11,7 +11,7 @@ engine = create_engine(
 )
 
 
-@event.listens_for(engine, "connect")
+@event.listens_for(engine, "checkout")
 def set_db_timeouts(dbapi_conn, _):
     # Prevent runaway queries from holding pool connections indefinitely.
     # lock_timeout: fail fast if waiting for a row lock (e.g. concurrent saves on same document).
--- a/backend/app/core/db.py
+++ b/backend/app/core/db.py
@@ -12,19 +12,18 @@ engine = create_engine(
 
 
 @event.listens_for(engine, "checkout")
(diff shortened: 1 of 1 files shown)


nofurtherinformation (approved):
Nice catch. LGTM!

backend/app/core/db.py
```diff
@@ -11,7 +11,7 @@
 )
 
 
-@event.listens_for(engine, "connect")
+@event.listens_for(engine, "checkout")
 def set_db_timeouts(dbapi_conn, _):
```
nofurtherinformation: H: From the tests run via GH action `TypeError: set_db_timeouts() takes 2 positional arguments but 3 were given`

`set_db_timeouts` is registered as a SQLAlchemy checkout pool
  event listener, but that event passes three arguments: `dbapi_connection, connection_record, and
  connection_proxy`. Our function only accepts two (`dbapi_conn, _`).



```suggestion
def set_db_timeouts(dbapi_conn, _connection_record, _connection_proxy):
```

Reference: SQLAlchemy's PoolEvents.checkout
  (https://docs.sqlalchemy.org/en/20/core/events.html#sqlalchemy.events.PoolEvents.checkout)

nofurtherinformation (changes requested):
Minor syntax/signature fix needed!

backend/app/core/db.py
fangge518: Thanks for the references. The interface is fixed.

nofurtherinformation (approved):
LGTM :)

PR #550, opened 2026-06-09
Perf: Msgpack and unassigned optimization

nofurtherinformation:
## Description

Performance pass on the assignment-heavy endpoints, plus a UX tweak for the "zoom to unassigned" flow. All changes are against `dev`.

**Binary wire format (msgpack) for assignment-heavy endpoints.** Dropped Pydantic response validation and `JSON.parse`/`stringify` in favor of msgpack on both ends. New shared client helpers `getMsgpack` / `putMsgpack` in `app/src/app/utils/api/msgpack.ts` so each handler stays a few lines.
- `GET /api/get_assignments/{id}` — rows now serialized as 3-tuples `[geo_id, zone, parent_path]`, dropping the repeated dict keys.
- `PUT /api/assignments` — accepts msgpack body, still validates against `AssignmentsCreate` after decode.
- `GET /document/{id}/contiguity/{zone}/connected_component_bboxes` — msgpack response.
- Logging: added `query_ms` for the `get_protected_document` dependency to make auth-step latency visible.

**Rewrote `GET /document/{id}/unassigned` to skip PostGIS.** Old endpoint called a `get_unassigned_bboxes` SQL UDF that did `ST_Union(ST_Envelope(...))` + `ST_Transform` across every unassigned geometry — slow on large maps. New behavior:
- SQL just enumerates unassigned `geo_id`s (no geometry, no transforms).
- NetworkX `connected_components` (over the already-cached parent-layer graph) groups them.
- Shattered children are resolved up to their parents via `parentchildedges` so adjacency works on one graph.
- Response is `{components: [[geo_id, ...], ...]}` as msgpack.
- Client (`GeometryWorker.getUnassignedGeometries`) computes per-component bboxes from centroid points it already has, instead of running Turf `dissolve` over polygons.
- The UDF migration is left in place; no other callers, can be dropped in a follow-up.

**Geometry worker — shattered-child centroid coverage.** Added `childPointData` + `setChildPointData` on the worker. `usePointData(isChild=true)` now pushes the child centroid set whenever `exposedChildIds` changes (and resets it when no parents are shattered), so unassigned bboxes that include shattered children still find centroids.

**Zoom UX.** Bumped per-component and overall unassigned-area `fitBounds` padding to 240px so users see surrounding context instead of a tight crop. `ZoomToFeature` gained an optional `padding` prop; the contiguity flow keeps its existing tight zoom by default.

## Reviewers

- Primary: @fangge518 
- Secondary:

## Checklist

- [ ] Added/Updated related documentation (if applicable). — _none touched; binary protocol noted in code comments and via the new helper module._
- [ ] Added/Updated related unit tests (if applicable). — _not added; would benefit from coverage on the unassigned components resolver (parent/child resolution, orphan fallback) and the msgpack PUT body parser._

## Review required

- **Wire format change is breaking** between the listed endpoints. Backend and frontend must deploy together; old clients hitting new server (or vice versa) will fail to parse responses / bodies. Confirm deploy ordering.
- **`/unassigned` correctness on edge cases**:
  - Documents with `child_layer` set but nothing shattered (this caused a regression earlier in the work — now fixed by always using the parent graph and resolving children up).
  - Documents with shattered parents whose children straddle multiple parent-graph neighborhoods.
  - Documents where some unassigned ids are missing from `parentchildedges` (orphan path returns them as singletons).
- **`PUT /api/assignments`** loses FastAPI's automatic OpenAPI request schema. If anything downstream consumes `/openapi.json` for this endpoint (codegen, docs), it'll now show as untyped.
- **Centroid-based bbox is approximate.** It's the bbox of the centroid points, not the polygon edges — tighter than the previous polygon-based output. The 240px `fitBounds` padding compensates, but worth a visual check on a few maps.
- **Shattered-children coverage in `pointData`.** Verify that scrolling between unassigned areas in a heavily shattered document still produces sensible zoom targets for child-level components.


nofurtherinformation:
Expected speed ups:
- GET `get_assignments` 2-3x improvement
- GET `unassigned` 3x to 15x ?!?!
- PUT `assignments` modest improvement, 0.2x

backend/app/main.py
```diff
@@ -649,6 +651,22 @@ async def update_assignments(
         HTTPException: 400 if no assignments provided
         HTTPException: 409 if document was updated by another client and overwrite=False
     """
+    body_bytes = await request.body()
+    try:
+        raw = msgpack.unpackb(body_bytes, raw=False)
+    except Exception as e:
```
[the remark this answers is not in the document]
nofurtherinformation: Stale comment - tests are fixed

app/src/app/utils/GeometryWorker/geometryWorker.ts
```diff
@@ -317,57 +310,96 @@ const GeometryWorker: GeometryWorkerClass = {
...
-      return {dissolved: {type: 'FeatureCollection', features: []}, overall: null};
+      if (minX < oMinX) oMinX = minX;
+      if (maxX > oMaxX) oMaxX = maxX;
+      if (minY < oMinY) oMinY = minY;
+      if (maxY > oMaxY) oMaxY = maxY;
+
+      componentFeatures.push({
+        type: 'Feature',
+        properties: {
+          bbox: [minX, minY, maxX, maxY],
```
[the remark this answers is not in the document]
nofurtherinformation: This is true but we don't actually load the geo into the worker anymore so we don't have it accessible. OK for now to test UX, if we get strong pushback we can revise

backend/app/main.py
```diff
@@ -649,6 +654,22 @@ async def update_assignments(
...
+        raise HTTPException(
+            status_code=status.HTTP_400_BAD_REQUEST,
+            detail=f"Could not decode msgpack body: {e}",
+        )
+    try:
+        data = AssignmentsCreate.model_validate(raw)
```
fangge518, line 675: I: You mentioned that model validation is the largest time sink of the API request. At this line, I assume we are still going through this slow procedure.

Here are some options to speed up the validation:
* AssignmentsCreate.model_construct(...) — bypasses validation entirely. Can there be malicious users?
* Restructuring the model to use a single flat list instead of nested objects (fewer validator invocations)
* Validating only the envelope/metadata and trusting the inner rows (partial validation)
* Pushing row validation into a SQL constraint or doing it set-based in the DB instead

backend/app/main.py
```diff
@@ -1119,7 +1140,7 @@ async def update_num_districts(
 
 
 # called by getAssignments in apiHandlers.ts
-@app.get("/api/get_assignments/{document_id}", response_model=list[AssignmentsResponse])
+@app.get("/api/get_assignments/{document_id}")
```
fangge518, line 1151: PP: As we discussed, operationally it might be useful to add a debug wrapper endpoint that shows human-readable assignments.

fangge518, line 1151: I: With the response changed to payload, the backend contract is no longer immediately visible for a new developer. Perhaps add a comment to document the shape of the response, or provide a reference for the relevant Model? Similarly for other changed endpoints.

backend/app/main.py
```diff
@@ -1262,32 +1280,98 @@ async def get_document_list(
...
     ).bindparams(
         bindparam(key="doc_uuid", type_=UUIDType),
         bindparam(key="exclude_ids", type_=ARRAY(String)),
     )
     try:
-        results = (
-            session.connection()
+        unassigned_ids = [
+            row[0]
+            for row in session.connection()
```
fangge518: I: What is iterated here is `session.connection().execute(..).fetchall()`, but the intentation has made this syntax obscure. Also we can just `session.execute(..)` here.

So perhaps:
```
result = session.execute(
    stmt, {"doc_uuid": document.document_id, "exclude_ids": exclude_ids}
)
unassigned_ids = [row[0] for row in result.fetchall()]
```

backend/app/main.py
```diff
@@ -1262,32 +1280,98 @@ async def get_document_list(
...
+    # Enumerate geo_ids that could be unassigned: the union of any rows already
+    # tracked in document.assignments (covers shattered children) and every parent
+    # path. Then keep only those whose effective assignment is NULL.
     stmt = text(
-        "SELECT bbox from get_unassigned_bboxes(:doc_uuid, :exclude_ids)"
+        f"""
+        WITH possible_ids AS (
+            SELECT DISTINCT geo_id FROM document.assignments WHERE document_id = :doc_uuid
+            UNION
+            SELECT path AS geo_id FROM gerrydb.{parent_layer}
```
fangge518, line 1364: H: If a parent node is shattered and its children fully assigned, this node will still show up in the results. I assume that's unintended behavior?

backend/app/main.py
```diff
@@ -1262,32 +1280,98 @@ async def get_document_list(
...
-        # TODO: When is this happening? Should investigate
         logger.warning("No results found for unassigned geoids")
-        results = []
+        unassigned_ids = []
 
-    return {"features": [row[0] for row in results if row[0] is not None]}
+    components: list[list[str]] = []
+    if unassigned_ids:
+        try:
+            G = await _get_graph(parent_layer)
```
fangge518: H: After 541 is merged, we should assume that the backend only have access to the hybrid-level dual-graphs for shatterable map, which have both parent-layer nodes and children-layer nodes. This block of code can be updated (and perhaps simplified) by directly checking whether the unassigned_ids (parent and children) are connected in that graph without resolving across levels (aside from dealing with non-contiguous parents, which we may choose to skip over here)

fangge518 (changes requested):
Awesome changes. I really look forward to seeing the 10+x improvement on the unassigned check. This batch of comments are targeting backend change only. I will review the frontend in another batch of comments.

backend/app/main.py
nofurtherinformation: Added in 6ca86418cdf0ea8293cfe3d6942ceb66daa619a0 -- this allows a query parameter to the route(`format={msgpack|json|csv}`) that returns those. We can enumerate addition future formats if needed!

backend/app/main.py
nofurtherinformation: Good catch -- cleaned up in ea15772035a0a5df112129a20ac51f6be0452845

backend/app/main.py
nofurtherinformation: Good catch, fortunately this is handled though! 

A while back we optimized this endpoint to rely on the client's knowledge of shattered parents. The client always knows this, and it saves us from having to complicate the query and bring in the parent/child edges table. The `exclude_ids` query param allows the client to explicit tell the query which parents are broken and can be ignore. 

This was (and remains) a reasonable optimization -- if the FE is out of sync on which parents are shattered, the application state is in bigger trouble, and there is no value in a malicious user trying to trick the backend into sending back extra unassigned units.

backend/app/main.py
nofurtherinformation: Definitely -- I was anticipating some conflicts once we get that merged in!

backend/app/main.py
nofurtherinformation: Clarified in 9cee31f122ac456698be147d2e535cab3099e0de

backend/app/main.py
nofurtherinformation: These are great ideas -- the main perf issue on model validation was the response validation from `get_assignments` although it's not zero cost here either. 

I did a quick sanity check locally to see the perf cost of the model validation and I think its not massive -- this is for Florida, so not as substantial as CA, but I think suggests the validation cost is not huge (and worth it, I think).

| total | body_read | msgpack_decode | model_validate | precheck_queries | delete | copy_load | insert | metadata | comments | commit | body_bytes |
|---|---|---|---|---|---|---|---|---|---|---|---|
| 139.1 | 0.4 | 1.3 | 2.3 | 6.0 | 10.8 | 21.3 | 25.3 | 8.7 | 53.8 | 9.3 | 134645 |
| 43.9 | 0.5 | 1.3 | 2.1 | 2.1 | 2.7 | 7.1 | 21.5 | 1.9 | 2.3 | 2.4 | 134651 |
| 38.7 | 0.2 | 0.6 | 1.0 | 2.4 | 2.5 | 7.3 | 19.4 | 2.0 | 1.3 | 2.0 | 134651 |
| 40.9 | 0.5 | 1.0 | 2.0 | 2.0 | 2.7 | 7.1 | 20.0 | 1.8 | 1.5 | 2.4 | 134651 |
| 50.4 | 0.5 | 1.1 | 1.9 | 3.1 | 2.9 | 7.9 | 22.7 | 1.9 | 2.1 | 6.3 | 134651 |
| 52.2 | 0.3 | 0.8 | 1.2 | 2.0 | 2.8 | 13.7 | 25.3 | 2.0 | 1.5 | 2.6 | 134651 |

backend/app/main.py
nofurtherinformation: More clarifications in 40ff57c1c1424ab7d376a6b8563779d24e596e87

backend/app/main.py
nofurtherinformation: Updated in 66ae59d866a9b8e356443e492482fea7493720bd

fangge518: This looks good to me now.

nofurtherinformation: TY!

backend/app/main.py
fangge518: Thank you for the explanation. I had always wondered what the exclude_ids is for, and now I have my answers. This contract is quite important for future work on importing maps. With that, the code looks good to me now.

backend/app/main.py
fangge518: Thank you for the perf data. With this knowledge, I'm happy with the current code and the choice to validate the UPDATE endpoint. Plus, it makes more sense to validate the inputs to the database.

fangge518 (approved):
The FE change looks good to me!

PR #551, opened 2026-06-09
publish /stats to S3, per-zone cache eviction, native GeoJSON

nofurtherinformation:
## Description

Follow-up to the run3 stress test (92% → 42% failure rate). Three root causes remained: RDS CPU saturation from `/stats` recomputing on every viewer request, a cold-cache unique-violation race under concurrent load, and several correctness gaps in the district_unions cache lifecycle.

- **Per-zone district_unions cache** — lazy rebuild: only recomputes zones whose membership changed on save (dirty-zone tracking via temp-table diff); evicts only affected rows rather than the whole document. Unassigned row recomputed only when a zone is rebuilt.
- **Stats CDN offload** — `/stats` redirects public reads to S3 when fresh (`stats_published_at ≥ assignments_updated_at`); background task publishes on first viewer miss. Added `assignments_updated_at` and `stats_published_at` timestamps to `document.document` to track staleness independently of `updated_at`.
- **Publish on save** — `PUT /assignments` and `POST /api/create_document` now enqueue `publish_district_stats_to_s3` as a background task, so the CDN is warm before the first viewer arrives post-save.
- **`ST_CoverageUnion` with `ST_UnaryUnion` fallback** — tries the faster linear-time union first (valid coverage topology); falls back via savepoint rollback on topology error or old PostGIS.
- **Cold-cache race fix** — both INSERTs use `ON CONFLICT DO NOTHING` + re-select for zones that lost the race; prevents unique-violation 500s under concurrent `/stats` load.
- **Thundering-herd guard** — `publish_district_stats_to_s3` reads freshness at entry and returns immediately if already fresh; collapses N concurrent publish tasks to one actual PUT.
- **Publish freshness race fix** — stamps `stats_published_at` with the `assignments_updated_at` snapshot taken *before* rebuilding, not `NOW()`, so a mid-publish save doesn't mark a stale object as fresh.
- **`reset_map` cache eviction** — deletes `district_unions` rows and bumps `assignments_updated_at` on reset; previously left the cache and CDN permanently stale.
- **Thumbnail regeneration fix** — `publish_district_stats_to_s3` now calls `generate_thumbnail` directly instead of losing it in a throwaway `BackgroundTasks()` bag.
- **`geometry: dict`** — `DistrictUnionsResponse.geometry` is now a native dict (from `ST_AsGeoJSON(...)::json`) rather than a JSON string; fixed three consumers: `shapely.geometry.shape()` in evaluation and shapefile export, `json.dumps()` in GeoJSON export f-string.
- **`GZipMiddleware`** — added at `minimum_size=1000`; covers `/evaluation`, document list, and other JSON endpoints (`/stats` is already gzip'd at the S3 layer).
- **Duplicate import removed** — `update_or_select_district_stats` was imported twice in `main.py`.

## Reviewers

- Primary:
- Secondary:

## Checklist

N/A — backend-only changes.

## Review required

- **`update_or_select_district_stats`** (`utils.py`) — the savepoint try/except for `ST_CoverageUnion` fallback, the `ON CONFLICT DO NOTHING` + re-select pattern for both INSERT paths, and the unassigned-row race recovery. Core of the cache correctness fix.
- **`publish_district_stats_to_s3`** (`utils.py`) — thundering-herd guard, `snap_ts` freshness stamp, and direct `generate_thumbnail` call. Verify the stamp logic: `stats_published_at = assignments_updated_at_at_publish_start` is the intended invariant.
- **`reset_map`** (`main.py`) — confirm the combined UPDATE (`color_scheme = NULL, assignments_updated_at = NOW()`) plus the `district_unions` DELETE covers all reset scenarios including community maps (community maps don't write to `district_unions`, so the DELETE is a no-op there — correct).
- **Migration** (`a30db9686b7c`) — `assignments_updated_at` and `stats_published_at` columns + partial unique indexes on `district_unions`. Confirm `alembic upgrade head` has run on prod (`\d+ document.district_unions` should show the indexes; `\d+ document.document` should show both new columns).
- **Three infra items not in this PR** (tracking separately): raise `backendMinCount` to 4, confirm departition migration on prod, move RDS to non-burstable class.


backend/app/main.py
```diff
@@ -264,12 +275,51 @@ async def db_is_alive(session: Session = Depends(get_session)):
...
+        document.stats_published_at is not None
+        and document.stats_published_at >= document.assignments_updated_at
+    )
+
+    if is_public_read and cdn_fresh and public_id is not None:
+        cdn = stats_cdn_url(
+            public_id, cache_buster=str(int(document.stats_published_at.timestamp()))
+        )
+        if cdn:
+            return RedirectResponse(
```
fangge518, line 303: O: Does the `?v=` cache-buster actually do anything on `tilesets1.cdn.districtr.org`? CloudFront ignores query strings for cache-key purposes unless the distribution's cache policy is configured to forward them, and the CDN's existing usage (tiles, parquet — stable, versioned paths baked into the bundle) never needed that, so there's no existing evidence it's turned on for this domain. If it's inert, a republished object could still be served from a stale CloudFront cache entry — bounded, though: the upload sets `max-age=60, must-revalidate`, so as long as CloudFront respects origin cache headers, staleness caps at ~60s post-republish rather than persisting indefinitely. Worth a staging check; not blocking.

backend/app/utils.py
```diff
@@ -640,59 +666,125 @@ def update_or_select_district_stats(
...
+                RETURNING
+                    zone,
+                    ST_AsGeoJSON(geometry)::json AS geometry,
+                    demographic_data,
+                    updated_at
+            """
+            # ST_CoverageUnion is faster (linear) but requires valid coverage
+            # topology (no overlaps, shared edges identical). Fall back to
+            # ST_UnaryUnion if it errors (topology violation or old PostGIS).
+            coverage_sql = insert_sql.replace(
```
fangge518, line 759: PP: Recommendation, not a blocker: consider a `pg_advisory_xact_lock` per (document_id, zone) around the check-cache → compute → insert sequence, in addition to keeping the unique index + ON CONFLICT as a correctness backstop. Reason: the aggregate union has to fully evaluate before Postgres can even check for a conflict, so under the current design every racer for a given zone pays the full compute cost regardless of who wins the write — ON CONFLICT dedupes the *write*, not the *work*. A loser also holds its connection for that whole computation, then possibly waits again on the winner's commit. An advisory-lock loser blocks immediately (no computation), and with a re-check-after-acquiring-lock, often skips computing at all — less connection-hold time under contention, which matters given the small pool. Worth it specifically for big states/popular documents, where an expensive union plus a concurrent viewership spike could otherwise reproduce a smaller version of the original CPU-saturation problem.

backend/app/main.py
```diff
@@ -754,6 +813,30 @@ async def update_assignments(
...
+    # below to drop only the affected rows from document.district_unions
+    # rather than wiping the cache for the whole document. Community maps
+    # don't feed into district_unions, so we skip the snapshot there.
+    diff_load_id: str | None = None
+    if not is_community_map:
+        diff_load_id, _ = str(uuid4()).split("-", maxsplit=1)
+        old_snapshot_table = f"old_assignments_{diff_load_id}"
+        session.connection().execute(
+            text(
+                f"CREATE TEMP TABLE {old_snapshot_table} "
```
fangge518, line 827: I: Since the frontend already sends the whole document's assignments on every save (not a delta), the new snapshot + post-write diff adds roughly 2x the DB-side set-based work of the pre-PR endpoint to every district-mode save — a real, unconditional per-save cost, not a small addition. Likely still worth it in aggregate if reads outnumber writes (the whole premise of this PR), but worth naming explicitly.

A cheaper design: `district_unions` already records which zones are currently clean (cached) vs. dirty (evicted, awaiting recompute) — an already-dirty zone can't get "more dirty," so there's nothing to protect by including it. Scope the snapshot to only geo_ids currently in a clean zone (small once anything's already gone dirty this session), plus a post-write check of geo_ids whose *new* zone lands in a clean zone (to catch transfers in from an already-dirty zone). Degrades to today's behavior in the worst case (nothing dirty yet) but gets cheaper across a burst of edits in one session — which seems like the common case this endpoint actually sees.

backend/app/utils.py
```diff
@@ -640,59 +666,125 @@ def update_or_select_district_stats(
...
+                "ST_UnaryUnion(ST_Collect(geos.geometry))",
+                "ST_CoverageUnion(ST_Collect(geos.geometry))",
+            )
+            try:
+                sp = session.begin_nested()
```
fangge518: I: Tiny one: `sp = session.begin_nested()` sits inside the `try`, so if `begin_nested()` itself raises, the `except` hits an unbound `sp` (NameError masking the original error). Hoist the assignment above the `try`.

backend/app/utils.py
```diff
@@ -640,59 +666,125 @@ def update_or_select_district_stats(
...
+                result = session.execute(
+                    text(coverage_sql),
+                    {"document_id": document_id, "missing_zones": list(missing_zones)},
+                )
+                rebuilt_rows = list(result.mappings().all())
+                sp.commit()
+            except Exception:
```
fangge518, line 771: I: The fallback is silent, so if prod's PostGIS/GEOS doesn't support `ST_CoverageUnion` (or our layers routinely violate coverage topology — split `-datadem-N` parts share edges by construction, worth confirming they don't overlap), production would run the try-fail-retry path on every rebuild with nothing in the logs. Suggest a `logger.info`/`warning` on fallback so we can see which union prod actually uses.

fangge518 (commented):
I've checked all Review-required items, and they are working as intended. A closer read of the code yields three I, one PP, one O — nothing here blocks merge.

backend/app/utils.py
nofurtherinformation: Nice catch -- handled in c27c076e8d37f1f1bf88243df2eda90e8f6fa4bd!

backend/app/utils.py
nofurtherinformation: Logged in e9403dda4d9ef3fa0b85ebbf4e02b0063b3840bd

backend/app/main.py
nofurtherinformation: I've updated the CDN config to respect query params, so we should be good here!

backend/app/utils.py
nofurtherinformation: Good catch! Added in a40468fa46e86b6d973d05c8592042ed1ad1abc0

backend/app/main.py
nofurtherinformation: This is a good catch, but we are pretty CPU rich right now, and it may be worth distributing the load over time. Worth keeping an eye on though!

PR #556, opened 2026-06-10
API Blocking work async cleanup

nofurtherinformation:
## Description

Every endpoint was declared `async def` while doing synchronous work (sync SQLAlchemy/psycopg, boto3, bcrypt, networkx), so any slow request blocked the worker's event loop — and uvicorn ran as a single process on a shared-cpu-4x
machine, so one slow export or stats request stalled every concurrent user.

- Convert DB-only handlers to plain `def` (34 endpoints across main, cms,  comments, exports, save_share, thumbnails) so FastAPI runs them in its  threadpool instead of on the event loop. A convention comment in `app/main.py` documents the rule for future endpoints.
- Comment-creation endpoints stay `async` for the recaptcha `await`; their DB   work now runs via `run_in_threadpool`.
- `_get_graph` is now sync with a load lock so concurrent cache misses can't unpickle the same multi-hundred-MB block graph twice.
- `VerifyToken.verify` is now sync, moving the blocking Auth0 JWKS fetch and `jwt.decode` off the event loop.
- Remove debug `print`s in the share endpoint that logged plaintext passwords and ran a redundant bcrypt hash (also crashed on passwordless shares).
- Deployment: `WEB_CONCURRENCY=4` in fly.toml (uvicorn worker count, tunable without rebuild); `PROMETHEUS_MULTIPROC_DIR` so `/metrics` aggregates across workers; explicit `[http_service.concurrency]` block (Fly's ~25-connection default throttles at the proxy otherwise); uvicorn 0.29 → 0.35 (multiprocess manager with worker respawn); add httptools; document per-worker DB pool math in `app/core/db.py`.

## Reviewers

- Primary:
- Secondary:

## Checklist

  - Endpoint sync/async convention documented in `app/main.py`; worker/pool sizing rationale documented in `fly.toml`, `Dockerfile`, and `db.py`.
  - No new tests: endpoint behavior is unchanged (sync/async conversion is transparent to the TestClient-based suite). Full suite passes (225).

N/A — backend/deployment only.

## Review required

- Deploy-time validation, not code: after release, watch per-worker RSS via `/_debug/cache` (graph LRU now duplicates per worker — drop WEB_CONCURRENCY to 2–3 or shrink the cache if memory gets tight), confirm Postgres  max_connections has headroom for up to 60 pooled connections (4 workers × 15), and spot-check `/metrics` aggregates correctly in multiprocess mode.
- Sanity-check the Fly concurrency limits (soft 100 / hard 150 requests) against real traffic patterns.


fangge518:
Superseded by #729, which covers this handler conversion plus `VerifyToken.verify` and the Turnstile comment threading. Closing this one.

PR #557, opened 2026-06-10
Version check component and api route

nofurtherinformation:
## Description

After #550 switched `PUT /api/assignments` from JSON to msgpack, users with tabs open from before the deploy hit `Could not decode msgpack body: unpack(b) received extra data` on every save — the stale bundle sends JSON, whose first byte (`{` = 0x7b) decodes as the msgpack integer 123, leaving the rest as "extra data". This PR adds version-skew detection so future deploys can't strand stale tabs against a changed API contract:

- **Build stamping**: `fly-deploy-app.yml` writes `NEXT_PUBLIC_BUILD_TAG=${{ github.sha }}` into `.env.production`. The same SHA is inlined into the client bundle at build time and loaded by `next start` at runtime (the file survives into the runtime image — `.dockerignore` only excludes `.env*.local`), so client and server versions come from one source of truth and are stable across server restarts.
- **`GET /api/version`** (Next.js route handler, `no-store`): reports the running server's build tag. Returns `{version: null}` in local dev, which disables the check entirely.
- **`<VersionCheck />`** (mounted in the root layout): compares the bundle's inlined tag against the endpoint on mount, when the tab becomes visible (via the existing `useVisibilityState` hook), on window focus, and every 5 minutes (polling pauses while hidden). On mismatch it shows a non-dismissible dialog — ESC/backdrop are no-ops — so reloading is the only way forward.

Limitations worth knowing:
- This protects **future** deploys only; tabs that are stale today don't have this code. A JSON fallback on the assignments endpoint would be the complementary fix for those. However, since we are in beta, it would be reasonable to simply move forward.
- The tag only bumps when `app/**` changes (that's what triggers the frontend deploy). API contract changes must touch both `app/` and `backend/` in one merge for stale tabs to get prompted — as #550 did.

## Reviewers

- Primary: @fangge518 
- Secondary:

## Checklist
  - Behavior documented in code comments on `VersionCheck.tsx` and `api/version/route.ts`; no user-facing docs affected.
  - No unit tests: the interesting behavior (build-time env inlining vs. runtime env, deploy stamping) only exists in a production build/deploy; the client logic is a thin fetch-and-compare. Manual validation steps below.

## Review required
- **Deploy plumbing** is the riskiest part and can't be tested locally: confirm on the dev deploy that `GET /api/version` returns the commit SHA (not `null`) and that no reload prompt appears immediately after deploy (which would indicate client/server tag mismatch from the build — e.g., `.env.production` not reaching the build or runtime image).
- **Skew simulation**: after this lands on dev, keep a tab open, push any trivial `app/**` change, and verify the dialog appears within ~5 minutes or on refocusing the tab — and that reloading clears it.
- **UX call**: the dialog is intentionally non-dismissible. If reviewers prefer a softer "remind me later" affordance, that's a one-line change in `VersionCheck.tsx`.

### DIFF
--- a/.github/workflows/fly-deploy-app.yml
+++ b/.github/workflows/fly-deploy-app.yml
@@ -33,6 +33,7 @@ jobs:
             NEXT_PUBLIC_S3_BUCKET_URL_MIRROR1=https://tilesets2.cdn.districtr.org
             NEXT_PUBLIC_S3_BUCKET_URL_MIRROR2=https://tilesets3.cdn.districtr.org
             NEXT_PUBLIC_API_URL=${{ steps.deploy-env.outputs.api_url }}
+            NEXT_PUBLIC_BUILD_TAG=${{ github.sha }}
             EOF
         working-directory: app
       - uses: superfly/flyctl-actions/setup-flyctl@1.5
--- a/app/src/app/api/version/route.ts
+++ b/app/src/app/api/version/route.ts
@@ -1,0 +1,24 @@
+import {unstable_noStore as noStore} from 'next/cache';
+
+/**
+ * Reports the build tag of the currently deployed frontend.
+ *
+ * NEXT_PUBLIC_BUILD_TAG is the git SHA written to .env.production by the
+ * deploy workflow. The same value is inlined into the client bundle at build
+ * time, so a client whose inlined tag differs from this response is running a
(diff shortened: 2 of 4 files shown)


fangge518 (approved):
LGTM. :)

PR #561, opened 2026-06-12
AWS migration: Pulumi infra, deploy workflows, and code prep

nofurtherinformation:
# AWS migration (phase 1–3): Pulumi infra, deploy workflows, and code prep

## Description

First three phases of the Fly.io → AWS migration (ECS Fargate + ALB, RDS PostgreSQL/PostGIS, Pulumi TypeScript with an S3 state backend, GitHub OIDC deploys). Nothing in this PR changes the running Fly deployment — the Fly workflows and apps stay live until the cutover runbook completes.

**Code prep**
- `backend/app/core/config.py`: `get_s3_client()` falls back to the default boto3 credential chain behind a new opt-in `AWS_USE_DEFAULT_CREDENTIALS` flag, so the ECS task role replaces static AWS keys. With the flag unset, local behavior is unchanged (`None` still means "S3 not configured", preserving the thumbnail-skip path in `app/utils.py`).
- `backend/app/evaluation/graph.py`: graph pickles stream from S3 directly into memory (`pickle.loads`) instead of being downloaded to the `/data` volume. The in-RAM `lru_cache` remains the only cache — re-downloads are free in-region via an S3 gateway endpoint, so no data volume is needed on Fargate. The local-path check stays for docker-compose dev with mounted `./data`.
- `app/Dockerfile`: `SENTRY_AUTH_TOKEN` enters the build as a BuildKit secret mount (never lands in an image layer); builds still work without it.

**Infrastructure (`infra/`)** — Pulumi TS project with `dev`/`prod` stacks:
- VPC with public subnets + strict security-group chaining (ALB → tasks → DB), no NAT gateway; free S3 gateway endpoint for graph/thumbnail traffic
- ECR repos (immutable tags, keep-last-20), RDS PostgreSQL + PostGIS (Multi-AZ prod, `dbPubliclyAccessible`/`operatorCidr` toggle for one-off data loads)
- One ALB per env: host-based routing (`api.*` → backend), ACM DNS validation, HTTP→HTTPS redirect, fixed 403 for `/metrics` and `/_debug/*`
- ECS Fargate services with target-tracking autoscaling, deployment circuit breakers with rollback, and a RunTask-only `districtr-{env}-migrate` task definition (Fly `release_command` equivalent)
- Secrets: Pulumi config (KMS-encrypted) → SSM SecureStrings → task-definition `secrets`
- CloudWatch log groups + alarm set (ALB 5xx, unhealthy targets, service memory, RDS CPU/storage) with optional SNS email
- `scripts/bootstrap.sh`: one-time state bucket / KMS key / GitHub OIDC provider + deploy role / SSM seed params
- `README.md`: bootstrap steps, day-to-day ops, and the Fly→RDS data-migration runbook

**CI/CD (`.github/workflows/`)**
- `infra.yml`: `pulumi preview` on PRs touching `infra/**`, `pulumi up` on push; `dev`→dev, `main`→prod (same branch convention as today)
- `deploy-api.yml`: build → ECR → run Alembic migrations as a one-off ECS task (deploy aborts with logs if it fails, service untouched) → `pulumi up`
- `deploy-app.yml`: writes `.env.production` exactly like the Fly workflow → build with Sentry secret → ECR → `pulumi up`
- Image tags flow through SSM params (`/districtr/{stack}/meta/*-image-tag`) written by the deploy workflows, so infra-only `pulumi up` runs never roll services back to a stale image; `backendImageTag`/`frontendImageTag` config values remain as manual rollback overrides

**Not in this PR** (needs AWS admin credentials; see `infra/README.md`): running bootstrap, filling stack-config TODOs with real Fly secret values, ACM/DNS records, data migration, and cutover.

## Reviewers

- Primary:
- Secondary:

## Checklist
- [x] Added/Updated related documentation (if applicable). — `infra/README.md` (bootstrap, ops, data-migration runbook)
- [x] Added/Updated related unit tests (if applicable). — `tests/test_config.py` (S3 client credential modes), `tests/test_graph.py` (S3 streaming, no disk writes)

N/A — infra and CI only; no UI changes.

## Review required
- **S3 client behavior change** (`config.py`): confirm the `AWS_USE_DEFAULT_CREDENTIALS` opt-in can't surprise any local/pipeline workflow that relies on `get_s3_client()` returning `None`.
- **Graph disk-cache removal** (`graph.py`): confirm no other consumer depended on pickles persisting under `$VOLUME_PATH/graphs/`.
- **Security posture decisions**: tasks in public subnets with public IPs (SG-locked to the ALB) instead of private subnets + NAT; deploy role starts at `AdministratorAccess` (scoping down is a follow-up); RDS subnet group is public to allow the temporary data-load toggle.
- **Sizing/cost assumptions**: backend 2 vCPU/8GB ×2 prod, RDS `db.t4g.large` Multi-AZ (~$500/mo prod, ~$140/mo dev) — sanity-check against budget before bootstrap.
- **Placeholders**: domain names and Auth0 values in `Pulumi.{dev,prod}.yaml` are `TODO(fill-in)` — real values must come from Fly secrets, not `fly.toml`.
- **Migration choreography** in `deploy-api.yml`: the describe→jq→register→run-task→wait→exit-code sequence deserves a careful read; it can't be exercised until the dev stack exists.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


.github/workflows/deploy-api.yml
```diff
@@ -0,0 +1,167 @@
+name: AWS Deploy API
+on:
+  push:
```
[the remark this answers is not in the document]
nofurtherinformation: Handled with repo environment vars `AWS_DEPLOY_DEV` and `AWS_DEPLOY_PROD`

infra/config.ts
```diff
@@ -0,0 +1,76 @@
...
+  frontendCpu: cfg.getNumber("frontendCpu") ?? (isProd ? 1024 : 512),
+  frontendMemory: cfg.getNumber("frontendMemory") ?? 2048,
+  frontendMinCount: cfg.getNumber("frontendMinCount") ?? (isProd ? 2 : 1),
+  frontendMaxCount: cfg.getNumber("frontendMaxCount") ?? (isProd ? 4 : 2),
+
+  // Database
+  dbInstanceClass: cfg.get("dbInstanceClass") ?? (isProd ? "db.t4g.large" : "db.t4g.small"),
+  dbAllocatedStorage: cfg.getNumber("dbAllocatedStorage") ?? (isProd ? 100 : 20),
+  dbMultiAz: cfg.getBoolean("dbMultiAz") ?? isProd,
+  dbEngineVersion: cfg.get("dbEngineVersion") ?? "16",
```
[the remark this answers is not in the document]
nofurtherinformation: Postgres migrations should be forward compatible. It's more important to be on the latest PG version

.github/workflows/deploy-app.yml
```diff
@@ -0,0 +1,132 @@
...
+        run: |
+          cat <<'EOF' > .env.production
+          NEXT_PUBLIC_RECAPTCHA_SITE_KEY=${{ secrets.RECAPTCHA_SITE_KEY }}
+          NEXT_PUBLIC_MAPTILER_API_KEY=${{ secrets.NEXT_PUBLIC_MAPTILER_API_KEY }}
+          NEXT_PUBLIC_S3_BUCKET_URL=https://tilesets1.cdn.districtr.org
+          NEXT_PUBLIC_S3_BUCKET_URL_MIRROR1=https://tilesets2.cdn.districtr.org
+          NEXT_PUBLIC_S3_BUCKET_URL_MIRROR2=https://tilesets3.cdn.districtr.org
+          NEXT_PUBLIC_API_URL=${{ steps.cfg.outputs.api_url }}
+          NEXT_PUBLIC_BUILD_TAG=${{ github.sha }}
+          EOF
```
fangge518, line 61: PP: The heredoc content is indented 10 spaces inside the YAML block, so every line in `.env.production` will start with 10 leading spaces:

```
          NEXT_PUBLIC_RECAPTCHA_SITE_KEY=abc123
```

This is the same as the fly deployment file, but since we are touching the file we might consider
writing the file with `printf` / `tee` so the indentation is unambiguous.
```
         printf '%s\n' \
            "NEXT_PUBLIC_RECAPTCHA_SITE_KEY=${{ secrets.RECAPTCHA_SITE_KEY }}" \
            "NEXT_PUBLIC_MAPTILER_API_KEY=${{ secrets.NEXT_PUBLIC_MAPTILER_API_KEY }}" \
            > .env.production
```

.github/workflows/deploy-app.yml
```diff
@@ -0,0 +1,132 @@
...
+            echo "Image already pushed for this sha; skipping build"
+          else
+            docker build --secret id=SENTRY_AUTH_TOKEN,env=SENTRY_AUTH_TOKEN -t "$IMAGE" app
+            docker push "$IMAGE"
+          fi
+          echo "image=$IMAGE" >> "$GITHUB_OUTPUT"
+      - name: Install Pulumi
+        run: |
+          curl -fsSL https://get.pulumi.com | sh
+          echo "$HOME/.pulumi/bin" >> "$GITHUB_PATH"
```
fangge518: I: All three deploy workflows install Pulumi with `curl -fsSL https://get.pulumi.com | sh`, which always fetches the latest version. A breaking Pulumi release could silently affect deploys. Consider using the official action instead, which supports version pinning:

```yaml
- uses: pulumi/actions@v5
  with:
    pulumi-version: "3.x.x"
```

.github/workflows/deploy-api.yml
```diff
@@ -0,0 +1,173 @@
...
+          # Own polling loop: `aws ecs wait tasks-stopped` gives up at 10
+          # minutes, which a long PostGIS migration can legitimately exceed.
+          DEADLINE=$((SECONDS + 2700))
+          STATUS=""
+          while [ "$SECONDS" -lt "$DEADLINE" ]; do
+            STATUS=$(aws ecs describe-tasks --cluster "$CLUSTER" --tasks "$TASK_ARN" \
+              --query 'tasks[0].lastStatus' --output text)
+            [ "$STATUS" = "STOPPED" ] && break
+            sleep 15
+          done
```
fangge518, line 122: I: If Fargate fails to place the migration task (transient capacity issue, IAM error, network misconfiguration), `run-task` returns a `failures` array instead of a `tasks` array and `tasks[0].taskArn` resolves to the literal string `"None"`. The polling loop then calls `describe-tasks --tasks "None"`, which returns an empty tasks array — `lastStatus` is never `"STOPPED"`, so the loop polls every 15 seconds for the full 45-minute deadline before timing out. The deploy is blocked for 45 minutes before anyone knows the task never started.

Perhaps add a guard immediately after capturing `TASK_ARN`:

```bash
if [ -z "$TASK_ARN" ] || [ "$TASK_ARN" = "None" ]; then
  echo "::error::Failed to place migration task — check ECS capacity and IAM permissions"
  exit 1
fi
```

infra/network.ts
```diff
@@ -0,0 +1,130 @@
...
+  // resources on one group makes the provider fight itself on refresh.
+  const dbIngress: aws.types.input.ec2.SecurityGroupIngress[] = [
+    {protocol: "tcp", fromPort: 5432, toPort: 5432, securityGroups: [backendSecurityGroup.id]},
+  ];
+  if (config.dbPubliclyAccessible && config.operatorCidr) {
+    // Temporary operator access for pg_dump/pg_restore during migration.
+    dbIngress.push({
+      protocol: "tcp",
+      fromPort: 5432,
+      toPort: 5432,
```
fangge518: I: Data-loading and backfills are an ongoing operational need, not just a one-time migration, so I worry that constantly flipping the toggle (which requires `pulumi up` and temporary public exposure) may not be practical as the standard approach.

Perhaps a better default: enable SSM Session Manager on the backend service, which lets you shell into a running ECS task and reach RDS from inside the VPC.

infra/database.ts
```diff
@@ -0,0 +1,58 @@
...
+    vpcSecurityGroupIds: [network.dbSecurityGroup.id],
+    publiclyAccessible: config.dbPubliclyAccessible,
+    multiAz: config.dbMultiAz,
+    backupRetentionPeriod: config.isProd ? 7 : 1,
+    copyTagsToSnapshot: true,
+    deletionProtection: config.isProd,
+    skipFinalSnapshot: !config.isProd,
+    finalSnapshotIdentifier: `${name}-db-final`,
+    performanceInsightsEnabled: config.isProd,
+    applyImmediately: true,
```
fangge518: O: `applyImmediately: true` applies RDS parameter and instance class changes outside the maintenance window. Fine for the initial bootstrap, but after launch, surprise parameter changes during business hours could cause brief unavailability. Consider switching to `false` post-cutover and relying on the maintenance window for non-urgent changes.

infra/README.md
fangge518, line 1: I: The RDS instance is plain PostgreSQL 16. RDS makes PostGIS available but does not install it automatically. `CREATE EXTENSION postgis;` must be run against the new database before Alembic migrations run for the first time. This step is not mentioned anywhere in `infra/README.md`. If it's missing from the bootstrap procedure, the migrate task will fail on first deploy with an error about missing spatial types.

backend/app/evaluation/graph.py
fangge518, line 1: PP: Similar to graph's disk cache, `context.py` downloads `county_names.csv` from S3 to `VOLUME_PATH/county_names.csv` on
first use and reads it from disk on subsequent calls. This is inconsistent with the
graph change in this PR (which now streams directly into memory). The
CSV is small so this has no meaningful performance impact, but it could be simplified to
match the graph pattern — stream into memory, skip the disk write entirely. Perhaps adding a TODO would suffice here.

infra/README.md
fangge518, line 1: PP: I think it might be worthwhile to explain how to set up a local development environment for working on the Pulumi code itself. Perhaps a section covering:
- Installing the Pulumi CLI (with a pinned version)
- Running `npm ci` inside `infra/` to install TypeScript dependencies and get editor type-checking working
- How to run `pulumi preview` locally before pushing

This is certainly non-blocking.

fangge518 (approved):
Thank you Dylan for setting up this Pulumi framework. I've learnt a lot just by looking at it.

I have written several comments below, but none of them is blocking. I think we should go ahead and try to deploy it on aws.

infra/database.ts
nofurtherinformation: Turned off for prod in 0c65f915bd47760b18e9210722f6f9a38d97c411

.github/workflows/deploy-api.yml
nofurtherinformation: Added in 5f7560a2874bb94324655203e26b1c283917ac5d

.github/workflows/deploy-app.yml
nofurtherinformation: Swapped in 7e138f77dca0080eae1ad196bd39a2367853e4bf

infra/README.md
nofurtherinformation: `966d8d72887e_add_gerrydb_schema.py:26` has this covered!

infra/network.ts
nofurtherinformation: Good call - swapped in 54fd3786633c5b463de6bf2e8ef8fb6920915ae7

backend/app/evaluation/graph.py
nofurtherinformation: Noted in 14eb7d3d668a8c10c74c5df7866d633455ca67bd!

.github/workflows/deploy-app.yml
nofurtherinformation: I believe the YAML `|` block strips the spaces -- we should be good here, but noting this as a follow up if we hit issues!

infra/README.md
nofurtherinformation: Added in 6032f119156f20a93d62adfbabec02d0f9136910!

PR #568, opened 2026-06-12
Preserve original zone IDs on CSV upload and optimization

### LINKED ISSUE #366: Block import zone number inference
In #260, unique zone values are remapped in order of appearance to `1..N` zone ids. However, in many cases the district number matters, as folks are used to talking about "District 5."

Ideally, zone IDs in block imports should be preserved as best as possible. The logic in block imports should attempt to use the user's zone numbers while still allowing for other distinct values to be mapped to zone ids.

fangge518:
Closes #366

## Summary

- **Remove zone remapping** in `batch_insert_assignments`: uploaded zone values are now inserted as-is (original integers from the CSV), instead of being remapped to 1, 2, 3… via a sequential defaultdict.
- **Graph-based backend logic rewrite**: `_heal_with_graph` replaces the previous SQL-heavy healing path.
- **Frontend full-row validation**: `validateRows` in `upload.ts` now checks every CSV row (not just the first 200 sample rows) for out-of-range zone values and errors early, before any network request is made.

## Performance

Profiled against a Tennessee census block upload (~9,241 rows) on a local instance with hot graph cache (LRU hit):

| Path | HTTP latency |
|---|---|
| Before (old code) | ~3s |
| After, cold graph (first load) | ~0.7s |
| After, hot graph (LRU hit) | ~0.3s |

## Backend behavioral changes

- Non-integer zone strings (e.g. `"My zone 1"`) now return HTTP 500 instead of a silent skip. The frontend already prevents these from being submitted (full-row validation), so this is a misuse path that should surface loudly.
- Duplicate geo_ids in upload input return HTTP 400 via Python-level detection (dict membership check before insert), replacing the old PostgreSQL UniqueViolation path which is bypassed by the dict deduplication in the new code.
- Out-of-range zones (exceeding `num_districts`) (also guarded by FE validation, should only occur in mischievous user behavior) are silently skipped like invalid geo_ids, rather than raising an error.

## Test plan

- [x] `pytest tests/test_main.py -k "block_assignment"` — 10 tests, all pass
- [x] `pytest tests/test_main.py` — 63 tests, all pass
- [x] Pre-commit (ruff + ruff-format + prettier) — all pass
- [x] Upload a real CSV to a local instance and verify zone values are preserved, no remapping


backend/app/assignments/assignments.py
```diff
@@ -122,140 +160,59 @@ def batch_insert_assignments(
...
-                        districtr_map.num_districts is not None
-                        and zone_val > districtr_map.num_districts
-                    ):
-                        raise ValueError("Too many unique zones provided")
-                    copy.write_row([record[0], zone_val])
-                else:
-                    null_count += 1
-            except ValueError:
-                import_errors += 1
+    G = get_graph(districtr_map.gerrydb_table_name)
```
nofurtherinformation, line 169: H: We should check if the selected module has a child layer before getting the graph, otherwise, we should just use the block assignments.

Also, we should update the `get_graph` error message: `Graph unavailable. This map does not support contiguity checks.` is the `except` message for s3 `ClientError` (file not found found essentially). This should probable be something like `Graph unavailable. Unable to complete this operation.` to be generic, or handle `HTTPException`s outside of `get_graph` -- eg `return [ok: bool, G: graph]; if ok: ... else: ...`

backend/app/assignments/assignments.py
```diff
@@ -122,140 +160,59 @@ def batch_insert_assignments(
...
+
+    null_count = 0
+    invalid_count = 0
+    zone_by_geo: dict[str, int] = {}
+    num_districts = districtr_map.num_districts
+    for record in assignments:
+        if record[1] and record[1] != "":
+            geo_id = record[0]
+            zone = int(record[1])  # ValueError propagates for non-integer zones
```
nofurtherinformation: H: Do we want this behavior or coerce to ints? Let's chat at today's meeting with moon. Here are some cases this may be not the correct behavior:
- Excel outputs 1.0, 2.0, _n.0_ instead of 1, 2, _n_
- Entries are 'District 1', 'District 2', 'District n'

backend/app/main.py
```diff
@@ -510,12 +510,16 @@ async def create_document(
...
-                    detail="Duplicate geoids found in input data. Ensure all geoids are unique",
-                )
+        except DuplicateGeoIdError:
+            raise HTTPException(
+                status_code=status.HTTP_400_BAD_REQUEST,
+                detail="Duplicate geoids found in input data. Ensure all geoids are unique",
+            )
+        except ValueError:
+            raise HTTPException(
+                status_code=status.HTTP_500_INTERNAL_SERVER_ERROR,
```
nofurtherinformation: I: This is still a client error, so we should probably use `400`(bad request) or `422` (unprocessable content) error status codes

backend/app/main.py
```diff
@@ -510,12 +510,16 @@ async def create_document(
...
-                raise HTTPException(
-                    status_code=status.HTTP_400_BAD_REQUEST,
-                    detail="Duplicate geoids found in input data. Ensure all geoids are unique",
-                )
+        except DuplicateGeoIdError:
+            raise HTTPException(
+                status_code=status.HTTP_400_BAD_REQUEST,
+                detail="Duplicate geoids found in input data. Ensure all geoids are unique",
+            )
+        except ValueError:
```
nofurtherinformation, line 520: PP: Elsewhere we have used explicit `session.rollback()` in except cases. This happens implicitly when the session closes but we could add the rollback here for clarity.

backend/app/assignments/assignments.py
```diff
@@ -102,6 +106,40 @@ def duplicate_document_community_assignments(
     return inserted_assignments
 
 
+def _heal_with_graph(zone_by_geo: dict[str, int], G: Graph) -> dict[str, int]:
```
nofurtherinformation, line 109: I: We should add a direct test for this

nofurtherinformation (changes requested):
This looks great overall. I've suggested we do some minor/moderate clean up changes, but big pictures looks good. 

One follow up not for this PR -- we're entering a dev ops space where it is **critical** that all the graphs always be present for each module. If we always do things right, that shouldn't be a problem, but that _can_ be a high bar. Can you add a tech debt / dev ops issue to make a CRON job that runs daily to snag all the current map modules and checks the existence of the expected graphs in S3 and sends out an alert if something is missing? This will be easier to execute once we are on AWS

app/src/app/utils/api/upload.ts
nofurtherinformation, line 1: P.s. While we are touching this file -- move the constants on rows `9:11` to constants?

backend/app/main.py
fangge518: Agreed. I've changed it to 422.

backend/app/main.py
fangge518: Done. Thanks for pointing it out.

backend/app/assignments/assignments.py
fangge518: Agreed and done.

backend/app/assignments/assignments.py
fangge518: Changed to the generic message.

I still retain the position for getting G here because it is also used for validating geo-id for non-shatterable maps. This step makes sure that all the path provided by the user are in fact geoids of the current map. The second step of healing reuse the G that is accessed in the validation step, but this second step occurs only for shatterable maps.

backend/app/assignments/assignments.py
nofurtherinformation: Won't do strings

fangge518: The floats are supported now.

Since Dave also does not support "Districts X" format, I will leave it for now.

nofurtherinformation (approved):
LGTM!

backend/app/assignments/assignments.py
fangge518: A unittest is added.

PR #572, opened 2026-06-16
Navigation and UX enhancements

nofurtherinformation:
## Description

Overhauls the in-map navigation so map **creation** and **recents** live on the static site, and switching between **Edit / Display / Evaluate** is a single, always-present control. Also locks the toolbar to the sidebar and consolidates every loading/locking overlay into one component.

**Navigation (4 goals)**
- Removed map creation from the in-map "Districtr" dropdown; the hamburger is now a focused tool menu (Home, New map, Upload, Export, View recent maps, Reset).
- "New map" now links to the static `/places` flow; "View recent maps" links to `/my-maps` (the old in-dropdown creator and dead `RecentMapsModal` are gone).
- New **`ViewSwitcher`** (Google Docs–style dropdown) switches Edit ⟷ Display ⟷ Evaluate on the current map. COI maps show only Edit/Display (no eval route).
- Editors switching to Display/Evaluate on an unshared draft get a **read-only `public_id` minted silently** (via `patchSharePlan`), persisted to IndexedDB, then navigate — no prompt. `editableDocId` is stashed so "back to Edit" works after a Display/Eval round-trip; Edit is disabled (with tooltip) when no owner UUID is known.
- Password-protected maps that are unlockable (`?pw=true`) or in local maps can enter Edit; otherwise Edit shows a lock + "Unlock to draw and paint districts" and opens the password prompt.

**Loading / transitions**
- New **`LoadingOverlay`** replaces the legacy `MapLockShade` and covers all cases: the ~3s fake evaluation pre-loader (Loading data → Cutting edges → Finalizing metrics), a real-time Display pre-loader, document load, save, and map-lock states.
- Overlays stay up until data is actually on the map (no blank-map flash), driven by a `loadingStates` dictionary (`publicSourceLoaded`, `metricsLoaded`, `documentLoading`) set by `PublicSource`/`EvalPanel`/`useDocumentWithSync` and read by the new `useViewTransition` hook.
- View switches keep the current camera (`lastViewBounds`) and `fitBounds(..., {animate: false})` — instant, no fly animation.

**Toolbar**
- Toolbar is now fixed to the sidebar and can no longer be dragged onto the map. Removed `DraggableToolbar` and the entire drag machinery from `toolbarStore` (location/x-y/rotation/dimensions + 9 unused setters), simplifying `ToolButtons`/`ToolControls`/`Toolbar`. Persist version bumped to drop stale layout state.

**Fixes / cleanup**
- `fetchDocument` falls back to the local IndexedDB copy on a remote 404 (fixes "document not found" for locally-cached maps).
- Fixed the "An image named 'lock' already exists" race in `MetaLayers`.
- Removed dead code: `MapLockShade.tsx`, `RecentMapsModal.tsx`, `ViewTransitionOverlay`, toolbar drag fields, and a stale `globals.css` print selector.

_32 files, +874/−714. Added 6 files, deleted 2._

## Reviewers

- Primary:
- Secondary:

## Checklist
- [x] Added/Updated related unit tests (if applicable). — Playwright e2e: `e2e/tests/map/topbar-navigation.spec.ts`, `e2e/tests/tools/toolbar-location.spec.ts`.
- [ ] Added/Updated related documentation (if applicable). — No user-facing docs; in-code comments updated (toolbar/overlay rationale).

## Review required

Most of this is UI/timing behavior that needs a human at a real browser — automated checks here were limited to typecheck, lint, and data-layer/DOM structure (the test browser couldn't reliably run paint/timer/navigation).

- **Silent auto-publish**: switching an unshared draft to Display/Evaluate mints a **read-only `public_id`**, making the map reachable by that numeric id with no prompt. Confirm this is the desired privacy trade-off and that it persists correctly to My-Maps / on reload.
- **Camera preservation**: verify the map keeps its view across Edit↔Display↔Evaluate and never flashes a blank/zoomed-out map.
- **Pre-loader timing**: the eval pre-loader should hold ~3s *and* until metrics load; the display pre-loader until the source loads; neither should linger after data is up (there's a 15s safety cap).
- **Password/unlock path**: `?pw=true` and local-maps cases into Edit, plus the locked-Edit copy for maps you can't edit.
- **Back-to-Edit**: from a Display/Eval view of an owned map, Edit returns to the UUID route; for someone else's shared link Edit is correctly disabled.
- **Toolbar**: confirm it renders/behaves identically in the sidebar (the layout was hardcoded to what previously rendered) and that persisted toolbar size survives the store version bump.


fangge518 (changes requested):
### Notes:
- Thanks for this work. Now the transition between different parts of the site becomes super snappy. :)
- The importer from the static site is not working at the moment when first loaded. This IS expected because our current importer depends on MapStore's mapViews which is only filled after the first loading the map. This will be fixed by the importer PR [#585](https://github.com/districtr/districtr-v2/pull/585)

### Nits:
**N1**. The first level menu items "Draw/Learn/Catelog" font looks a bit small; perhaps we can enlarge it to the same level as [district.org](http://district.org/)'s top menu font.
**N2**. In each state's landing page, the top level's "Draw" grays out. Is this intended? I thought the landing pages belong as a sub page of "Draw" category.
**N3**. Assuming I have the power to edit but temporarily switch to the view mode, the share link at the view mode still asks to create a new copy of the map. This should be the case for users with only view access, but feels unintuitive for someone with edit access. 

### Bugs:
**B1**. In the draw mode's topbar menu, clicking on "catalog" sometimes does not take us back to catalog, especially when we was first redirect to this mode from the "Draw" of the static site.
**B2**. Evaluate Mode's click-to-zoom feature stops working.

### Questions:
**O1**. Should we allow users without edit access to a map to still able to export the map? Currently they aren't allowed because our BE's exporting API requires the uuid, which they don't have. But we can enable them by changing our exporting API into supporting public id.

(@peterrrock2: "As for export, this should definitely be supported for public users. Dave's uses a login system to gate the export function, but making an account is almost zero effort so any member of the public can access it.") 

The following two lists are action items, each preferably with a numbered observation/question spawning the need for the action.
### Immediate work:
- Fix B1.
- Understanding the the reasons why click-to-zoom features stops working. This shouldn't be affected by the navigational revamp. Actual fix can be applied latter if difficult.

### Future Work:
- Fix B2 if fixing it is difficult right now.

app/src/app/components/Topbar/ModeSwitcher.tsx
fangge518, line 1: I: Since "View" is now one mode out of many, the name for this file might be confusing. Perhaps "ModeSwitcher"?

app/src/app/components/Topbar/ModeSwitcher.tsx
```diff
@@ -0,0 +1,249 @@
...
+      target = await mintAndResolve(mode);
+    }
+    if (!target) return;
+    // Reset the destination's load flags and show the overlay, then navigate
+    // immediately so the overlay covers the real load and only clears once the
+    // destination view's data has loaded (see LoadingOverlay / useViewTransition).
+    setLoadingState('publicSourceLoaded', false);
+    setLoadingState('metricsLoaded', false);
+    setViewTransition(mode);
+    router.push(target);
```
fangge518, line 203: O: Here is one UX question that might worth discussing further: should we force users to save local changes to the cloud before going into View Mode? I think users might expect the View Mode to show whatever that is shown in the Edit Mode and be surprised when these mismatch.

app/src/app/components/Topbar/ViewSwitcher.tsx
```diff
@@ -0,0 +1,249 @@
...
+  const prefix = routeForMode(mapMode);
+
+  // Public/eval pages load the doc read-only and surface document_id as "anonymous",
+  // so fall back to the UUID retained for this session (set while editing).
+  const liveDocId = mapDocument.document_id;
+  // A map can be edited if we hold its UUID directly (edit route), retained it this
+  // session, or have a local copy (owner). Password-protected maps without a local
+  // copy are editable by unlocking instead.
+  const editDocId =
+    (liveDocId && isUUID(liveDocId) ? liveDocId : null) ?? editableDocId ?? localEditDocId;
```
fangge518: H: When the map owner opens their own password-protected share link in a new tab, they see the password prompt (triggered by ?pw=true / password_required), but localEditDocId resolves from IDB and lets them switch to Edit after dismissing it without entering a password. To the owner, this may look like the password isn't protecting anything — they might not realize they bypassed it only because the browser recognized their local copy. Consider suppressing the prompt entirely when a local copy is found, or adding a note in the UI ("You own this map — no password needed").

nofurtherinformation:
Thanks Ge!

> **N1**. The first level menu items "Draw/Learn/Catelog" font looks a bit small; perhaps we can enlarge it to the same level as [district.org](http://district.org/)'s top menu font. 

Resolved in 6c6605eac44c86d7bef59c230923df201cccb152

>**N2**. In each state's landing page, the top level's "Draw" grays out. Is this intended? I thought the landing pages belong as a sub page of "Draw" category. 

Resolved in 6c6605eac44c86d7bef59c230923df201cccb152

> **B1**. In the draw mode's topbar menu, clicking on "catalog" sometimes does not take us back to catalog, especially when we was first redirect to this mode from the "Draw" of the static site. 

Resolved in 6c6605eac44c86d7bef59c230923df201cccb152 to make the target bigger.

> **B2**. Evaluate Mode's click-to-zoom feature stops working.

Fixed in d83f205cc6c9b5ce30df15fdbb3310b548901d94. This related to a change in the query key of the public view to support more rapid mode switching

app/src/app/components/Topbar/useViewTransition.ts
```diff
@@ -0,0 +1,58 @@
...
+    };
+  }, [viewTransition, setViewTransition]);
+
+  // Clear once the destination view's data is loaded and the minimum has elapsed.
+  React.useEffect(() => {
+    if (!viewTransition || !minElapsed) return;
+    const dataReady =
+      viewTransition === 'evaluate' ? publicSourceLoaded && metricsLoaded : publicSourceLoaded;
+    if (dataReady) setViewTransition(null);
+  }, [viewTransition, minElapsed, publicSourceLoaded, metricsLoaded, setViewTransition]);
```
fangge518: O: This is a super nit:

The second effect does not need to depend on viewTransition/setViewTransition. It can use a ref to viewTransition. 

```
  const viewTransitionRef = React.useRef(viewTransition);
  viewTransitionRef.current = viewTransition;

...

  React.useEffect(() => {
    if (!minElapsed || !viewTransitionRef.current) return;
    const isEval = viewTransitionRef.current === 'evaluate';
    const dataReady = isEval ? publicSourceLoaded && metricsLoaded : publicSourceLoaded;
    if (dataReady) setViewTransition(null);
  }, [minElapsed, publicSourceLoaded, metricsLoaded, setViewTransition]);

```

I think it might make its function clearer. But again, choosing one way over another isn't very important. The current code's logic is valid as well.

app/src/app/utils/idb/idb.ts
```diff
@@ -75,6 +75,17 @@ export class DocumentsDB extends Dexie {
       .then(documents => documents.map(document => document.document_metadata));
   }
 
+  /**
+   * Find the editable document_id (UUID) of a locally-stored map by its public_id,
+   * or null if the user has no local copy. Lets read-only/public views offer "Edit"
+   * for maps the user owns locally.
+   */
+  async getEditableIdByPublicId(publicId: number): Promise<string | null> {
+    const docs = await this.documents.toArray();
```
fangge518, line 87: PP: This path is called a lot, so we might want to avoid operations that are too slow. If a user has a large number of maps stored, it might be worthwhile to find the publicID through some index. But this is non-blocking. Perhaps a TODO would suffice here.

app/src/app/store/mapControlsStore.tsx
```diff
@@ -101,6 +124,16 @@ export const useMapControlsStore = create<MapControlsStore>()(
     setIsEditing: isEditing => set({isEditing}),
     isEval: false,
     setIsEval: isEval => set({isEval}),
+    evalTablesOnly: false,
+    setEvalTablesOnly: evalTablesOnly => set({evalTablesOnly}),
```
fangge518: I: This toggle can be removed as we haven't decided what a table-only report would look like.

fangge518 (changes requested):
Some thoughts on the FE logic, not including the static pages.

peterrrock2:
## Navigation and UX (release 2.2.5 prep)

@nofurtherinformation 

### Notes:

1.  I like the two layer nav bar, but I think that just changing the color of the text is a bit too subtle.

Verified working

- [x] **Silent auto-publish** 
- [x] **Camera preservation** 
- [x] **Pre-loader timing**
- [x] **Toolbar**

???

- [] Password/unlock path: ?pw=true and local-maps cases into Edit, plus the locked-Edit copy for maps you can't edit.

^ I think that this is working, but I'm a little confused by the wording


### Nits:

1. Post (6c)[https://github.com/districtr/districtr-v2/commit/6c6605eac44c86d7bef59c230923df201cccb152] the text is still a bit small / squished to my eye
3. Filter in Catalog: "Filter by map name, **slug**, or module". I think that the words "slug" and "module" are confusing to users.
4. Moving from "Draw" -> "View" with unsaved changes does not warn the user. This can be confusing.
5. "Cutting Edges" on the user spin page is a bit confusing

### Bugs:

1. There is a sequence of events using the new `ViewSwitcher` dropdown that can cause you have an out of sync eval page which requires drawing to correct. Link to recordings in slack.
2. (very minor) You can pull the sidebar past left of the window which then requires refreshing to fix.
3. If I make a password protected map, and then make a new password protected map, it looks like my old password is cached

### Immediate Work:

1. Remove password caching (Bug 3)
2. Decide if Bug 1 is this sprint or next

### Future Work:

1.  Add other ways of exporting Evaluation metrics beyond JSON. (Not in this PR but I noticed it)

### Questions:

1. Should we add a "Main Map" option into the hamburger in top-left of main drawing interface?

fangge518:
A new bug regarding the catalog's maps: when going back to the catalog (browser go back or menu dropdown go back) from edit mode, the map currently being edited is not clickable in the catalog list, while all other maps are normal. Refreshing the catalog would make it normal again.

app/src/app/utils/idb/idb.ts
nofurtherinformation: Added in f04783c96b8943794fbacb33149ad0ca8003d80a -- I think this is a true micro-optimization but we could refine this sometime!

app/src/app/store/mapControlsStore.tsx
nofurtherinformation: Removed in 52e519cc816c5626435b447ad1932a2192f55fef

app/src/app/components/Topbar/useViewTransition.ts
nofurtherinformation: Cleaned up in 1e7da1f330502b647c9bb8af82a176d0b018cab8

app/src/app/components/Topbar/ModeSwitcher.tsx
nofurtherinformation: Auto-save now occurs on switch!

app/src/app/components/Topbar/ModeSwitcher.tsx
nofurtherinformation: Renamed in 5730ebe010fd09415d4ecea3f1213f967e1b578a

nofurtherinformation:
There was an underlying bug when rapidly switching views where the wrong timestamp could get saved to IDB, causing the FE state to believe it was up to date with the backend. This was replicable by rapidly going back and forth between the hydrated pages (eg `catalog` and `map/edit`). This is resolved in 3df7bcf1b64c46cf0ddc50b0750a65163dfa31f8

nofurtherinformation:
> 1. Post (6c)[https://github.com/districtr/districtr-v2/commit/6c6605eac44c86d7bef59c230923df201cccb152] the text is still a bit small / squished to my eye

Increased more in b27e093c2385775e768180becfa558d685f5f4d3
Screenshot 2026-06-27 at 6 28 21 PM


> 2. Filter in Catalog: "Filter by map name, **slug**, or module". I think that the words "slug" and "module" are confusing to users.

Simplified Screenshot 2026-06-27 at 6 28 37 PM


> 3. Moving from "Draw" -> "View" with unsaved changes does not warn the user. This can be confusing.

Auto-save added

> 4. "Cutting Edges" on the user spin page is a bit confusing

Renamed to be more vernacular in bce19c7fbc3d001b238f65c85eb371081de2e052

> ### Bugs:
> 1. There is a sequence of events using the new `ViewSwitcher` dropdow
n that can cause you have an out of sync eval page which requires drawing to correct. Link to recordings in slack.

Handled!

> 2. (very minor) You can pull the sidebar past left of the window which then requires refreshing to fix.

Added some footgun guardrails in a15249befa4c9134954d0ef8e1d125b4b8a25b57

> 3. If I make a password protected map, and then make a new password protected map, it looks like my old password is cached

Cleaned up in f3574dd3d185faa707d7f86905926751842a76ad -- needs a double check

> ### Questions:
> 1. Should we add a "Main Map" option into the hamburger in top-left of main drawing interface?

Added

PR #578, opened 2026-06-18
Export overhaul

fangge518:
# Export overhaul

## What changed

**Block assignment CSV**: switched from a SQL-level join to graph-based expansion — assignments are fetched at the parent level then expanded to child GEOIDs using the in-memory GerryDB graph.

**District boundary exports (GeoJSON + Shapefile), Evaluation JSON**: one-line wrappers around the already-cached results; no second SQL query needed.

## Test plan

- [x] `docker-compose exec backend pytest backend/tests/test_exports.py -v`
- [x] Export block assignments CSV and verify child GEOIDs are correct
- [x] Export GeoJSON and Shapefile and verify geometry in QGIS
- [x] Export evaluation JSON and confirm metrics are present
- [x] Confirm all four export types appear in the frontend Export menu


app/src/app/components/Topbar/Topbar.tsx
```diff
@@ -149,47 +150,51 @@ export const Topbar: React.FC = () => {
...
+                    disabled={
+                      !mapDocument?.document_id || mapDocument.document_id === ANONYMOUS_DOCUMENT_ID
+                    }
+                  >
                     Export assignments
                   </DropdownMenu.SubTrigger>
                   <DropdownMenu.SubContent>
-                    <DropdownMenu.Item>
-                      <Tooltip content="Download a CSV of Census GEOIDs and zone IDs">
+                    <DropdownMenu.Item disabled={!mapDocument?.child_layer}>
```
nofurtherinformation: H: Should we disable only if no child layer and the parent unit is not blocks?

backend/app/exports/main.py
```diff
@@ -1,124 +1,109 @@
...
+        writer.writerow(["geo_id", "zone"])
+        for geo_id, zone in rows:
+            if "children" in G.nodes[geo_id]:
+                for child in G.nodes[geo_id]["children"]:
+                    writer.writerow([child, zone])
+            else:
+                writer.writerow([geo_id, zone])
+
+
+def build_evaluation_json(
```
nofurtherinformation, line 69: I: Docstring pls

nofurtherinformation (approved):
One small (but important!) UI tweak and an ask for docstrings on the new functions, but otherwise LGTM!

backend/app/exports/main.py
```diff
@@ -1,124 +1,109 @@
...
+        .where(Document.document_id == document_id)
+    ).first()
+    if doc_row is None:
+        raise ValueError(f"No map found for document_id: {document_id}")
+    gerrydb_table_name, child_layer = doc_row
+
+    rows = session.exec(
+        select(Assignments.geo_id, Assignments.zone)
+        .where(col(Assignments.document_id) == document_id)
+        .where(col(Assignments.zone).is_not(None))
```
nofurtherinformation, line 47: H: One more thing - we have a subtle edge case here:
- A user breaks a parent unit
- User leaves some blocks unpainted or erases some
- The blocks would be in the assignments as unassigned (zone null), but retained as entries so we have the full list of blocks

Given that, i think we need to leave in null/None assignments on this query

app/src/app/components/Topbar/Topbar.tsx
fangge518: Good catch. This condition made sense previously when we allowed for the exporting of parent-layer units, which caught all non-shatterable maps. For this version, we should let this option catch those, too. So here I have removed the guard completely.

backend/app/exports/main.py
fangge518: Done.

backend/app/exports/main.py
fangge518: As we discussed in slack, we decided not to export unassigned rows in the block assignment CSV. Compatibility with Dave's Redistricting was the deciding factor: Dave's treats any sentinel value as a literal zone label (-1 becomes a zone named "-1", "None" becomes "None", and empty string becomes zone "1"), so no sentinel is safe. This also matches Dave's own convention of simply omitting unassigned units.

To maintain the frontend/backend data contract — which requires every child block of a shattered parent to have an entry in document.assignments (with zone = NULL for unassigned blocks) so the map can render and expose them as interactive — we instead enforce correctness on import: _heal_or_fill now detects any parent with at least one assigned child in the uploaded CSV and inserts the remaining siblings as zone = NULL rows. This approach is more robust to varied user inputs and does not require users to know our internal data contract.

Issue #580, opened 2026-06-23
Evaluate page polish


fangge518:
This is a persisting thread that collects the UX, UI, and editorial feedback for the evaluation page. Here is the template for a batch of review:

## TITLE

"@" for assigned dev

The following are numbered lists of observations/questions broken down into four categories.

### Notes:
Simple observations that don't fit into other categories.

### Nits:
Non-blocking critiques of work. Should be resolvable in this thread.

### Bugs:
Can spawn immediate/blocking work or future/non-blocking work. An issue should probably be opened for these.

### Questions:
Open questions that the reviewer would like input on

The following two lists are action items, each preferably with a numbered observation/question spawning the need for the action.
### Immediate work:
Blocking. Things that must be changed before merge / release.

### Future Work:
Other pieces that were noticed during review that fall outside the scope of the PR, but need to be addressed.

fangge518:
I've pasted (and slightly edited) what @peterrrock2 said in the slack channel:

## Release 2.2.4 prep

[@Ge Fang](https://mggg-space.slack.com/team/U0ANE0MLZ28)

### Notes:
**O1**. It's nice that hovering on the district numbers and clicking outlines and zooms on a district 
**O2**. Top-to-bottom deviation is reported for plans with only 1 district, which is a bit strange. 

### Nits:
**N1**.  A completely unassigned plan reports some statistics (Basics, Compactness), but not others (Election Results and Partisanship, Splits). We should either have placeholder text for the missing categories or not report anything at all and have a template message.
We can get messages like "Your plan's most populous district is District 6 and least populous district is District 1, for a top-to-bottom deviation of 38370.323%." We might want to clamp this to "over 100%".
**N2**.  When a map is redrawn on the editable side and then you refresh the eval page, there is not progress indicator telling the user what is going on
**N3**. Toggling POV for Election results mildly changes the width of the columns on my machine
**N4**. Boundary overlays are covered by dissolved districts, unlike the old districtr (reference image attached).

<img width="878" height="556" alt="Image" src="https://github.com/user-attachments/assets/7359cce6-18f7-45df-896a-502ea7d2664c" />

### Bugs:
[...]

PR #582, opened 2026-06-24
Eval page more polish

fangge518:
## Problems

Described in https://github.com/districtr/districtr-v2/issues/580#issuecomment-4782840570

## Changes

- **Empty plan**: show a placeholder message instead of partial metrics when the plan has no assignments
- **Deviation clamping**: cap top-to-bottom deviation display at "over 100%" for plans with extreme imbalance
- **Table width stability**: fix column width jitter when toggling Point of View in the partisanship tables (`width` on toggle-affected headers; `table-layout: fixed` for the metrics table)
- **Table alignment**: left-align the "Votes vs. Seats" title and both POV toggles
- **Counties table**: rename "Overly Split Counties" to "Counties"; add a Show dropdown ("unnecessarily split counties" / "all counties", defaulting to the former); highlight the "Pieces in This Plan" cell in red for unnecessarily split counties in both views

### DIFF
--- a/app/src/app/components/EvalPanel/BasicsSection.tsx
+++ b/app/src/app/components/EvalPanel/BasicsSection.tsx
@@ -12,6 +12,7 @@ import {useZoomToDistrict} from '@/app/hooks/useZoomToDistrict';
 type DeviationView = 'top_to_bottom' | 'max_absolute' | 'both';
 
 function formatDeviation(value: number): string {
+  if (value > 1) return 'over 100%';
   const pct = value * 100;
   const formatted = pct.toFixed(3);
   return formatted === '0.000' ? 'under 0.001%' : `${formatted}%`;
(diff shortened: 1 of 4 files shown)


nofurtherinformation (commented):
Code LGTM.

Two UX questions:
1. Is this how we want the tables to be, with a bottom and right border but no others?
Image

2. Did we get direction on which districts to use for Polsby popper / reock examples?
Image

fangge518:
> Code LGTM.
> 
> Two UX questions:
> 
> 1. Is this how we want the tables to be, with a bottom and right border but no others?
> 
> Image

This does look strange. I have added borders and use a consistent style for tables.

> 2. Did we get direction on which districts to use for Polsby popper / reock examples?
> 
> Image

Not yet. This should be easy to change in a minimal PR in the future.

fangge518:
Some changes intended for this PR are not pushed to the remote before this PR is merged. Those are merged through PR #584

Issue #462, opened 2025-11-04
Uploader workflow


nofurtherinformation:
Improve the workflow to make it more obvious how to upload a CSV

Fix the list

Issue #109, opened 2024-09-25
Load from block assignment file


mariogiampieri:
Need to spec out, but presumably a user would be able to load a table with geoid:zone_id and have that pre-populate a map.

nofurtherinformation:
In progress in #241

cdonnay:
Importing and exporting to CSVs are a highly desired feature

- challenge for imports is weird headers
- has to be at block level, but backend could merge children blocks into parent precincts
- Dave's can do this without the user specifying anything about their csv input

originally tagged as @mapmeld and @raphaellaude 

currently in a PR as #254 and commit 85a10d4

cdonnay:
We are deprioritizing geojson, shapefile, and thumbnails

PR #585, opened 2026-06-25
Simplify CSV import: auto-infer map, remap string zone labels

fangge518:
## Context

This PR address issues #109 and #462.

## Summary

- Remove map selector UI from the uploader; the congressional map is now inferred from the state FIPS code in the uploaded GEOIDs
- First column is assumed to be GEOID, second column zone — no column detection heuristics
- String zone labels (e.g. `"District 1"`, `"01"`) are silently remapped to unused integer slots; whole-number floats like `"2.0"` are accepted as-is
- Errors surface for: mixed states in CSV, no matching congressional map, too many distinct zones
- Report the remapped district labels both during import and in district comments.
- Made some modifications to the backend preview specification to make fullstack preview work

## Review Needed
- The frontend has some preliminary validation check. We should double check the efficiency and the correctness of such checks. 
- The decision to inter the map from the fips code, and use the congressional version of the state map.
- The backend's algorithm for building the zone numbers from strings that can be numbers or string labels.


peterrrock2:
## CSV import feedback

@fangge518 

### Notes:

1. Worked with a couple of good plans
2. Partial assignments seems to work. 

### Nits:

1. When uploading something with more districts than expected"Too many zones" -> "Too many districts"
    - We will want to get rid of this eventually since we will let people choose the number of districts. Will need a cap
2. When uploading something with invalid geoids error message is: "Graph unavailable. Unable to complete this operation." This doesn't really tell me what went wrong.
3. Somewhat unexpectedly, whole vtds work.
[OK_small_with_vtds.csv](https://github.com/user-attachments/files/29523237/OK_small_with_vtds.csv)



### Bugs:

1. Mixed state detection does not error properly. See attached csv for mix of CO, AL, and CT where I have removed parts of CT and added things for CO and AL that do not get caught.
[AL_CO_CT_mix.csv](https://github.com/user-attachments/files/29522616/AL_CO_CT_mix.csv)
    - Should probably construct a fips set by looping through the whole file quickly on the FE and then warn users. Could eventually upgrade this to prompt users to select what state they intended this to be for.
2. Invalid blocks are silently ignored. See 
[OH_with_invalid_block.csv](https://github.com/user-attachments/files/29522762/OH_with_invalid_block.csv)
3. KS is not working?
[KS-block-assignments.csv](https://github.com/user-attachments/files/29522910/KS-block-assignments.csv)


### Immediate Work:

1. Fix Bugs 1, 2, and 3.

### Future Work:

1.  Possibly allowing submissions with mixed states to select a state

fangge518:
> ## Response to CSV import feedback by @peterrrock2 
> 
> ### Nits:
> 1. When uploading something with more districts than expected"Too many zones" -> "Too many districts"
>    
>    * We will want to get rid of this eventually since we will let people choose the number of districts. Will need a cap

Agreed. Text is changed.
> 2. When uploading something with invalid geoids error message is: "Graph unavailable. Unable to complete this operation." This doesn't really tell me what went wrong.

This is a side effect of Bug 3. Now that Bug 3 is fixed, users should not see this. If the user does see this, the problem really is the missing of the dual-graph for the actively maintained map modules.

> 3. Somewhat unexpectedly, whole vtds work.
>    [OK_small_with_vtds.csv](https://github.com/user-attachments/files/29523237/OK_small_with_vtds.csv)
> 
This is unexpected. This csv does not work now: will emit a geoid error.

> ### Bugs:
> 1. Mixed state detection does not error properly. See attached csv for mix of CO, AL, and CT where I have removed parts of CT and added things for CO and AL that do not get caught.
>    [AL_CO_CT_mix.csv](https://github.com/user-attachments/files/29522616/AL_CO_CT_mix.csv)
>    
>    * Should probably construct a fips set by looping through the whole file quickly on the FE and then warn users. Could eventually upgrade this to prompt users to select what state they intended this to be for.

Fixed. The user will see an error now. I've added a TODO for future extension of this function

> 2. Invalid blocks are silently ignored. See
>    [OH_with_invalid_block.csv](https://github.com/user-attachments/files/29522762/OH_with_invalid_block.csv)

Invalid blocks are returned from the BE and listed before the user clicks a button to redirect to the imported map.

> 3. KS is not working?
>    [KS-block-assignments.csv](https://github.com/user-attachments/files/29522910/KS-block-assignments.csv)
> 

This is caused by the not-so-robust logic for districtrmap inference. Fixed now.

peterrrock2:
## Follow Up on CSV import

@fangge518 

### Notes:

1. Old bugs appear to be addressed

### Nits:

1. "Go to imported map" -> "Show map anyway" for when you want to proceed to the map
2. There is an inconsistency in the handling of bad GEOIDs. If they look like blocks, they just get dropped. If they look like VTDs, then loading completely fails even if there are valid blocks. Probably best to default to warning + "Show map anyway"

.github/workflows/fly-preview.yml
```diff
@@ -113,7 +113,7 @@ jobs:
           AUTH0_DOMAIN: ${{ secrets.AUTH0_DOMAIN }}
           AUTH0_API_AUDIENCE: ${{ secrets.AUTH0_API_AUDIENCE }}
           AUTH0_ISSUER: ${{ secrets.AUTH0_ISSUER }}
-          AUTH0_ALGORITHMS: ${{ secrets.AUTH0_ALGORITHMS }}
+          AUTH0_ALGORITHMS: RS256
```
peterrrock2, line 116: Why make this change?

fangge518:
## Response to @peterrrock2 's follow-up comments
> ### Nits:
> 1. "Go to imported map" -> "Show map anyway" for when you want to proceed to the map
> 2. There is an inconsistency in the handling of bad GEOIDs. If they look like blocks, they just get dropped. If they look like VTDs, then loading completely fails even if there are valid blocks. Probably best to default to warning + "Show map anyway"

Agreed and all fixed in [3364d8c](https://github.com/districtr/districtr-v2/pull/585/commits/3364d8cb33b18135705a900648ef860a4c249746)

Note: At this stage, we only skip these vtds/bgs. We can also support importing those, but this would further increase BE logic complexity because we need to check whether vtds/bgs overlap with blocks in the assignments. We can revisit whether it is worthwhile to implement this in the future.

app/src/app/utils/api/upload.ts
```diff
@@ -5,194 +5,210 @@ import {DistrictrMap} from '@/app/utils/api/apiHandlers/types';
...
+  const skippedGeoIds: string[] = [];
+  const stateFipsSet = new Set<string>();
+  for (const row of rows) {
+    const raw = `${row[0] ?? ''}`.trim();
+    const parsed = parseGeoid(raw);
+    if (parsed?.type === 'block') {
+      blockRows.push(row);
+      stateFipsSet.add(parsed.stateFips);
+    } else {
+      skippedGeoIds.push(raw || '[empty]');
```
peterrrock2, line 82: H: Bug. GEOIDs are parsed and trimmed here

app/src/app/utils/api/upload.ts
```diff
@@ -5,194 +5,210 @@ import {DistrictrMap} from '@/app/utils/api/apiHandlers/types';
...
-          detail: {message: `Cannot upload more than ${MAX_ROWS} rows at once`},
+          detail: {message: `No congressional map found for ${stateName}`},
         });
+        return;
       }
 
+      const assignments: [string, string][] = blockRows.map(r => [
+        (r[0] ?? '').padStart(15, '0'),
+        (r[1] ?? '').trim(),
+      ]);
```
peterrrock2: H: Bug. c.f. [https://github.com/districtr/districtr-v2/pull/585/changes#r3514584995](https://github.com/districtr/districtr-v2/pull/585/changes#r3514584995) untrimmed version is assigned here.

app/src/app/utils/api/upload.ts
peterrrock2: Continued [https://github.com/districtr/districtr-v2/pull/585/changes#r3514598453](https://github.com/districtr/districtr-v2/pull/585/changes#r3514598453)

backend/app/assignments/assignments.py
```diff
@@ -181,28 +237,32 @@ def batch_insert_assignments(
...
-                raise DuplicateGeoIdError(geo_id)
-            zone_by_geo_int[geo_id] = zone
-        else:
-            null_count += 1
+        raw_zones.add(record[1])
 
-    logger.info(
-        f"{null_count} unassigned rows skipped, {invalid_count} invalid geo_ids/zones skipped"
-    )
+    zone_mapping = _build_zone_mapping(raw_zones, num_districts)
```
peterrrock2: I: Bug. This builds the mapping before the invalid GEOIDs are skipped. So invalid ids still affect the district numbering.

peterrrock2 (changes requested):
This is looking good! A couple of small edge cases that I noticed after reading through this.

### High / hold-up (H)

1. We check for GEOID membership after trimming but then pass the untrimmed version to the BE. See "spaced_geoid.csv".

### Important / Improvement (I)

  1. Numbers beyond the expected number of districts are silently converted to strings, which is probably unexpected. See "district_over_co_count.csv".
        - We should probably add a pop up telling the user that we remapped some of their districts so that they know about it, with a button "Continue to Map" at the bottom of the popup.
2. Invalid GEOIDs still contribute to district count. See  "skipped_geoid_counts_towards_limit.csv"

peterrrock2:
Files for latest review
[spaced_geoid.csv](https://github.com/user-attachments/files/29606936/spaced_geoid.csv)
[skipped_geoid_counts_toward_limit.csv](https://github.com/user-attachments/files/29606937/skipped_geoid_counts_toward_limit.csv)
[district_over_co_count.csv](https://github.com/user-attachments/files/29606938/district_over_co_count.csv)

.github/workflows/fly-preview.yml
fangge518: The previous reference to a non-existent secret caused preview deployments to fail at startup. Since every other environment (docker, CI tests) hardcodes RS256 and the value is a public algorithm name rather than a credential, we hardcode it here to match.

backend/app/assignments/assignments.py
fangge518: This is an intentional decision. Putting it before the loop allows us to do the entire processing and zone assignment in one pass, making the code run faster.

The UX cost for this added efficiency is extremely small. The valid assignments do not see any difference. For csvs containing invalid geoids, the cost I see (that is not captured by the examples below) is that users may see number gaps. I think this cost is acceptable, because if the user uses string labels, these numbers are remapped numbers, which are not semantically meaningful. If they use actual numbers, then the meaningful zone number is not supposed to change. If it pushes the total number past the allowed number of districts, (as the skipped_geoid_counts_toward_limit.csv below shows), import will fail and force the user to prune the csv. One may argue that a failure could useful at forcing the user to prune the assignment and check implicit assumption.

app/src/app/utils/api/upload.ts
fangge518: Good catch. We computed the normalized (padded) value inside parseGeoid, but we discarded it by pushing the original row into blockRows. Although this does not lead to processing errors, this is indeed quite redundant, inefficient, and confusing. I've constructed the assignment array directly in the validation loop. The code is much more elegant now.

app/src/app/utils/api/upload.ts
fangge518: Agreed. The assignment array is now directly constructed in this function instead of going through another mapping operation downstream.

peterrrock2 (approved):
LGTM!

Issue #569, opened 2026-06-15
DevOps: daily cron to audit graph pkl presence in S3 for all active map modules


fangge518:
## Background

`get_graph` is now critical for both contiguity checks and CSV upload geo_id validation (`batch_insert_assignments`). A missing graph pkl silently breaks uploads for that module with a 404.

## Task

Implement a daily cron job that:
1. Queries the DB for all active `DistrictrMap` records and their `gerrydb_table_name` values
2. Checks S3 for the existence of `graphs/{gerrydb_table_name}.pkl` for each
3. Sends an alert (email or Slack) listing any missing graphs

## Notes

- Will be straightforward to wire up on AWS (Lambda + EventBridge cron + SNS/SES for alerting)
- Deferred until after AWS migration

PR #595, opened 2026-06-29
Daily S3 graph comprehensiveness check

fangge518:
## Context
This PR addresses issue #569 by setting up the cron job for monitoring the graph pickle file.

## Summary

- Adds a `check-missing-graphs` CLI command that queries all `DistrictrMap` records with a `gerrydb_table_name` and does an `s3.head_object` check for each `graphs/{name}.pkl` in S3. Any missing files trigger a single SNS alert email.
- Adds `ALARM_SNS_TOPIC_ARN` to backend `Settings` (no-op when unset, safe for local dev).
- Adds Pulumi resources in `infra/graphcheck.ts`: a dedicated ECS Fargate task definition (reusing the backend image), a least-privilege task role (`s3:HeadObject/GetObject/ListBucket` + `sns:Publish` scoped to the alarm topic), and an EventBridge Scheduler rule that runs the task daily at 06:00 UTC.
- Refactor `createBackend` to add a utility function that exposes `executionRole`, `image`, `environment`, and `secrets` for reuse by the graph-check task.

## Test plan before merge

- [x] Run `python cli.py check-missing-graphs` locally against a DB with known maps and verify correct S3 key checks and log output

## Test plan after merge
- [ ] After deploy, manually trigger the ECS task via `aws ecs run-task` and confirm SNS email fires for any missing graphs


infra/config.ts
```diff
@@ -68,6 +68,6 @@ export const config = {
   vpcCidr: cfg.get("vpcCidr") ?? (isProd ? "10.0.0.0/16" : "10.1.0.0/16"),
 
   // Monitoring
-  alarmEmail: cfg.get("alarmEmail"),
+  alarmEmails: cfg.getObject<string[]>("alarmEmails") ?? [],
```
nofurtherinformation, line 71: H: We need to update `Pulumi.dev.yaml` with this too

backend/cli.py
```diff
@@ -822,5 +826,63 @@ def delete_overlay(session: Session, overlay_id: str):
...
+            logger.warning("Graph %s: s3://%s/%s", status, settings.R2_BUCKET_NAME, key)
+            problems.append((m.gerrydb_table_name, status))
+
+    if not problems:
+        logger.info("All %d graph(s) present in S3.", len(maps))
+        return
+
+    topic_arn = settings.ALARM_SNS_TOPIC_ARN
+    if not topic_arn:
+        logger.warning("ALARM_SNS_TOPIC_ARN not set — skipping SNS alert.")
```
nofurtherinformation: I: Should we throw an error with this, not just a warning?

backend/cli.py
```diff
@@ -822,5 +826,63 @@ def delete_overlay(session: Session, overlay_id: str):
         logger.error(f"Overlay with ID {overlay_id} not found")
 
 
+@cli.command("check-missing-graphs")
+@with_session
+def check_missing_graphs(session: Session):
+    """Query all DistrictrMap records and alert via SNS if any graph pkl is absent from S3."""
+    maps = session.scalars(
+        select(DistrictrMap).where(col(DistrictrMap.gerrydb_table_name).isnot(None))
```
nofurtherinformation: I: Do we want to query only maps that are `visible=True` so we can exclude legacy or deprecated maps?

infra/graphcheck.ts
```diff
@@ -0,0 +1,133 @@
+import * as aws from "@pulumi/aws";
+import * as pulumi from "@pulumi/pulumi";
+import {config} from "./config";
+import {ClusterResources} from "./cluster";
+import {Network} from "./network";
+import {BackendTaskConfig} from "./backend";
+
+const ECS_TASKS_TRUST = aws.iam.assumeRolePolicyForPrincipal({
```
nofurtherinformation: H: I think this is the same as `backend.ts:8-10` -- can we export from there (or a common `ecs_trust.ts` file?) and import?

backend/cli.py
```diff
@@ -822,5 +826,63 @@ def delete_overlay(session: Session, overlay_id: str):
...
+
+    # (gerrydb_table_name, status) where status is "missing" or an S3 error code
+    problems = []
+    for m in maps:
+        assert m.gerrydb_table_name is not None
+        key = f"{S3_GRAPH_PREFIX}/{m.gerrydb_table_name}.pkl"
+        try:
+            s3.head_object(Bucket=settings.R2_BUCKET_NAME, Key=key)
+            logger.info("Graph present: s3://%s/%s", settings.R2_BUCKET_NAME, key)
+        except botocore.exceptions.ClientError as e:
```
nofurtherinformation, line 868: H: There are some other errors that could (unlikely, but possible) that could come up here, use `botocore.exceptions.BotoCoreError` or good ol' `except Exception` ?

infra/graphcheck.ts
```diff
@@ -0,0 +1,133 @@
...
+  clusterResources: ClusterResources,
+  network: Network,
+  taskConfig: BackendTaskConfig,
+  alarmTopicArn: pulumi.Input<string>
+) {
+  const name = config.name;
+  const {cluster} = clusterResources;
+  const {executionRole, image, environment, secrets, logConfiguration} = taskConfig;
+
+  const logGroup = new aws.cloudwatch.LogGroup(`${name}-graph-check-logs`, {
```
nofurtherinformation: PP: We could move this to `cluster.ts:23` where other log groups are centralized

backend/cli.py
```diff
@@ -822,5 +826,63 @@ def delete_overlay(session: Session, overlay_id: str):
...
+        return
+
+    s3 = settings.get_s3_client()
+    if not s3:
+        logger.error("S3 client not available — check AWS credentials configuration.")
+        raise SystemExit(1)
+
+    # (gerrydb_table_name, status) where status is "missing" or an S3 error code
+    problems = []
+    for m in maps:
```
nofurtherinformation, line 862: I: Getting each object could be slow, could we do something like this?
```python
paginator = s3.get_paginator("list_objects_v2")
existing = {
    obj["Key"]
    for page in paginator.paginate(Bucket=settings.R2_BUCKET_NAME, Prefix=f"{S3_GRAPH_PREFIX}/")
    for obj in page.get("Contents", [])
}
missing = [n for n in names if f"{S3_GRAPH_PREFIX}/{n}.pkl" not in existing]
```

nofurtherinformation (changes requested):
Solid work! A few things here that are important but should be straightforward :)

infra/config.ts
fangge518: Good catch. Fixed in [22969db](https://github.com/districtr/districtr-v2/pull/595/commits/22969db916a002fde75a3cb4608ce90634d8b264)

backend/cli.py
fangge518: This is to allow local, manual usage of this command without sending any email notifications. Perhaps using a null setting is not the best way to indicate this.

I've added a bool to allow local/manual usage in [22969db](https://github.com/districtr/districtr-v2/pull/595/commits/22969db916a002fde75a3cb4608ce90634d8b264).

backend/cli.py
fangge518: Yes. Thanks for informing me about this flag. I was worried about being flooded by warning messages about deprecated maps. 

Fixed in [af7a31b](https://github.com/districtr/districtr-v2/pull/595/commits/af7a31ba740660ea90f3b06706f911f84501069c)

infra/graphcheck.ts
fangge518: Refactored into a common file backendTask.ts, along with the backendTaskConfig, in [08d3a34](https://github.com/districtr/districtr-v2/pull/595/commits/08d3a346bcb31fef29cad9824781d55828e55437).

backend/cli.py
fangge518: Good catch. New errors are handle in [b075b28](https://github.com/districtr/districtr-v2/pull/595/commits/b075b283be30e731cfd09fc15345f5d72b736c1b)

infra/graphcheck.ts
fangge518: Agreed. This is way more elegant. Changed in [ead0473](https://github.com/districtr/districtr-v2/pull/595/commits/ead0473823b038d5ba5b5d721af516f6a4f901a8)

backend/cli.py
fangge518: A single list_objects_v2 call would definitely be faster than N HEADs. I still decided to stay with the head_object loop for two reasons:

1. The savings don't surface anywhere. With several dozen maps the sequential HEADs take ~2–4 seconds, and the job runs once a day inside a Fargate task whose runtime is dominated by provisioning and image pull (30–60s+). The listing would shave a few seconds off a ~90-second daily job.

2. Per-object HEADs preserve the missing-vs-unreachable distinction. Each key failure carries its own error code, so a 403 or a transient S3 error is reported as "unreachable" rather than "missing." With a single LIST, any key absent from the listing looks missing, which is a more coarse-grained report.

nofurtherinformation: Sounds reasonable!

backend/cli.py
nofurtherinformation: Gotcha, thanks for the context!

nofurtherinformation (approved):
LGTM!

Issue #517, opened 2026-03-20
If a user's last action is an undo, IDB does not properly cache updates


nofurtherinformation:
This manifests in undone changes re-appearing on refresh for both community and district modes

nofurtherinformation:
https://github.com/user-attachments/assets/ec0c342f-18c2-4f9e-a9fd-7639539e1657

nofurtherinformation:
There is also have weird behavior if you delete all of the communities and then try to call undo/redo

fangge518:
Fixed with #597

Issue #493, opened 2026-02-16
After break/shatter, users should stay on the same tool


nofurtherinformation:

fangge518:
Fixed in #597

PR #597, opened 2026-06-30
County boundary visibility, undo/redo interaction with sync, and shatter behavior

### LINKED ISSUE #579: County boundary layer styling
## Description:
The county boundary layer was too light. This was lightly adjusted in #577 but a more comprehensive pass is needed to see if we want different boundary colors in different modes.

fangge518:
Closes #579, #517, #493

This PR bundles several bug fixes for easier deployment preview.

---

## #579 — County boundary layers not visible above district fills

**Files:** `layerIds.ts`, `layerRenderConfig.ts`, `CountyLayers.tsx`, 5 map shell components (`MainMap`, `CoiMap`, `DemographicMap`, `PublicMap`, `PublicDemographicMap`)

Added a `reference` anchor slot in the map layer stack, positioned above all fill/geometry layers but below hover feedback. County boundary lines (`COUNTIES.BOUNDARY`) now render at this anchor so they stay visible through district fill colors. Labels, hit-test fill, and hover highlight stay in their existing positions.

## #517 — Undo does not persist to IDB; cloud save corrupts undo/redo history

**Files:** `temporalStore.ts`, `putUpdateAssignmentsAndVerify.ts`

Two related bugs with the same root cause (zundo temporal tracking + IDB not updated):

**Bug 1 (IDB not written after undo):** `undo()`/`redo()` in zundo restore a snapshot in-memory but never write IDB. On refresh, `fetchDocument` finds `clientLastUpdated === updated_at`, treats there as "no local edits", and loads the server copy — discarding the undo. Fix: wrap `undo`/`redo` in `useTemporalStore` to immediately write IDB with `new Date().toISOString()`. A fresh timestamp (not the snapshot's) ensures IDB is always treated as having local edits newer than server.

**Bug 2 (Cloud save corrupts undo stack):** After a successful save, `setClientLastUpdated(T_save)` fires `temporalDiff`, which pushes a spurious snapshot onto `pastStates` and clears `futureStates`. Fix: pause both temporal stores around `setClientLastUpdated` in `putUpdateAssignmentsAndVerify.ts`.

## #493 — After break/shatter, clicking "return to districts" resets tool to Shatter

**File:** `mapStore.ts` (`exitBlockView`)

When exiting break mode, `exitBlockView` was unconditionally switching `activeTool` to SHATTER, discarding whatever tool the user had been using inside the block view (brush, eraser). Removed that override so the active tool is preserved on exit.

## Test plan

- [x] Enable county boundaries in draw mode — lines visible above district fills
- [x] Undo a paint stroke, refresh page — undo is preserved
- [x] Paint → undo → save — redo remains available after save
- [x] Enter break mode, switch to pan or eraser, click "return to districts" — tool stays on pan/eraser


peterrrock2 (approved):

nofurtherinformation:
LGTM!

PR #601, opened 2026-07-06
Feat/nav ux feedback

nofurtherinformation:
## Description

Consolidates map editor actions into a single dropdown, adds hover previews for the site nav, and introduces autosave.

**Map actions dropdown (editor topbar)**
- New `MapActionsDropdown` consolidates Save, Share, Export assignments, Visual map settings, Revert, and Reset into one menu, replacing the individual Share/Save/Revert/Settings topbar buttons (`SavePopover`, `SharePopoverAndModal`, `RevertPopover`, `SettingsPopoverAndModal` deleted; their dialogs lifted into the dropdown).
- The top-left "Districtr" hamburger is now navigation-only (Home / Main Map / Catalog) — Export and Reset moved to Map actions.
- All gating preserved: Save/Revert only while editing and disabled with no pending changes; Share/Export/Reset disabled without a document; Reset disabled for read-only access; dropdown hidden on eval pages.
- Trigger and Mode switcher restyled (`surface` variant + hover shadow) to read as interactive; Map actions trigger shows a MixIcon that turns red with a corner badge when there are unsaved changes, and the Save item shows the red cloud icon.
- Dropdown menus are at least as wide as their trigger buttons (`--radix-dropdown-menu-trigger-width`).

**Header hover subnav (static site)**
- Hovering **Learn** or **Catalog** in the site header shows the full-width `SecondaryNav` bar flush under the header (new `floating` prop), so the mouse can travel straight into it. Suppressed on pages already showing that subnav statically; Draw has no subnav.

**Autosave**
- New `useAutoSave` hook (mounted in Topbar): saves pending changes on tab close/hide, on window unfocus, and after 30s of non-action from the user. Single-flight so simultaneous blur + visibilitychange triggers can't race a stale `last_updated_at`. Constants in `constants/document/sync.ts`.
- Close-time saves are best-effort (no `keepalive`) — an interrupted final save fails silently and remains pending locally.

**Fixes**
- API error messages: caught fetch errors were serialized with `JSON.stringify(error)` (always `"{}"` for `Error` objects); now surfaces `error.message`.

## Reviewers

- Primary:
- Secondary:

## Checklist

  - No docs cover the topbar/nav UI.
  - Updated `e2e/tests/save-share/share-map.spec.ts` to open the Map actions menu before clicking Share/Reset.

## Review required

- Autosave behavior is the highest-risk piece: verify no unexpected saves fire while brushing, on alt-tab, or in read-only/eval views, and that the "Saving plan" lock flash on blur feels acceptable.
- Visual/UX judgment: dropdown item ordering and the red icon+badge treatment on the trigger.
- Hover subnav feel on `/draw` and content pages (flush positioning when scrolled, no flicker moving mouse into the bar).
- e2e specs were restructured around the new menu — worth a CI run plus a quick manual pass of share and reset flows.


fangge518:
In general it looks pretty good!

## Nits:
N1. The secondary-level menu on hover of "Draw" and "Learn" looks good. Given this design, I think it would be even more intuitive if these labels are aligned to the right or to the horizontal position of the first level label.

N2. The label "Visual map settings" can be shortened to "Visual settings." Since it is already under the "Map actions" menu, I think users will unambiguously understand it as visual settings of the map.

## Personal Opinion
With unsaved changes, the red dot at the map actions is making me, a slightly compulsive individual, much more anxious while drawing, especially in comparison to the previous design, where the unsaved icon has no connection to preconditioned compulsions to clean all the red dots. An even more compulsive individual, I imagine, would want to save for every small change they make. This is perhaps not worth further thoughts now, but worth more discussion in the future.

peterrrock2:
## Feat/nave UX tweeks (prep for 2.2.5)

@nofurtherinformation 

### Notes:

1. I like the new dropdown menus from Learn / Draw / Catalog.
2. Autosave appears to be working fine.
3. Adding in the map actions button was a good choice here. Topbar components now feel much more cohesive.

### Nits:

1. This might just be me. It's a bit hard to parse what I am selecting in the subnav when I am hovering. The text is light and the underline is thin. Increasing the width of the underline or making the visual signal otherwise stronger would help, I think.
2. I find the red dot helpful. But we might want to add some text next to save map that tells the user that autosave exists. Something like "Save map <sub>(autosave on)</sub>".
3. When autosave hits, the save can happen so fast that it looks like the app glitched unexpectedly. We should add a 1-2 second speedbump to the display with a "Auto saving ..." popup to tell the user that  we are saving their map for an auto save coming from a timeout. The autosaves could probably be spaced from the last interaction a little bit more as well (maybe 30s - 1m of idle).
    - It looks like the idle timeout applies to things like map actions but not comments. Also there is a button called "Save" when I submit a comment that I expect to save the whole map. Probably wouldn't be the worst thing to have that be the case since the delta will likely just be the comment.

fangge518:
The static site's main-secondary menu interaction looks good to me, as well as the topbar. The last nitpick is the autosave feature, which is PP and non-blocking.

## Nits:
N3. When drawing, autosave only kicks in, and the associated "Map Saved" message shown, when the user has modified something. When switching modes, "Map Saved" message will always show when switching from or to the draw mode. This signal might mislead user after being conditioned on the draw mode on the trigger for autosave. The user might think there is some unsaved change saved when there is none. Perhaps we should only show the "Map Saved" message when something change is being saved?

nofurtherinformation:
> The static site's main-secondary menu interaction looks good to me, as well as the topbar. The last nitpick is the autosave feature, which is PP and non-blocking.

Thank you! 

> ## Nits:
> N3. When drawing, autosave only kicks in, and the associated "Map Saved" message shown, when the user has modified something. When switching modes, "Map Saved" message will always show when switching from or to the draw mode. This signal might mislead user after being conditioned on the draw mode on the trigger for autosave. The user might think there is some unsaved change saved when there is none. Perhaps we should only show the "Map Saved" message when something change is being saved?

Good catch, yes -- 429f0c267b75c046f1a195452a382658e59f4007 does just this.

nofurtherinformation:
@peterrrock2 ae6b3c1485a31ecf3822bc6e5e875f51530c65b7 simplifies the autosave as we discussed

peterrrock2:
## Feat/nav ux feedback

@nofurtherinformation 

### Notes:

1. We talked about simplifying the autosave model to "if inactive for 30 s or if switches active window/tab"
    - Confirmed working!
3. Map actions is looking good.
4. Save on comment is working
5. Auto save on alt-tab seems to be working
6. Hamburger seems to be fine
7. I like the link coloring on hover in the subnav. 

### Nits:

1. Pulling the shade too far to the left causes an overflow on the map actions button.
image


### Bugs:

1. If you export the assignments from an unsaved map with changes, you will export the last save, and then the map will save itself. So if you hit export twice without making any changes, you will get two different exports.

### Future Work:

1.  Bug 1 is an edge case and can be a different PR

nofurtherinformation:
> ### Bugs:
> 1. If you export the assignments from an unsaved map with changes, you will export the last save, and then the map will save itself. So if you hit export twice without making any changes, you will get two different exports.

Wrapped better in 28c5acb54171273fcf20aea5d2d7e8a6032d5d29

> ### Future Work:
> 1. Bug 1 is an edge case and can be a different PR

Noted in #613 !

PR #603, opened 2026-07-06
Feat/under construction page cms

nofurtherinformation:
## Description
- Adds a small site settings table and toggle for under construction

## Reviewers
- Primary: @fangge518 
- Secondary:

Screenshot 2026-07-06 at 3 47 11 PM
Screenshot 2026-07-06 at 3 47 17 PM

## Review required
Spot check of mission critical code pieces

### DIFF
--- a/app/src/app/admin/page.tsx
+++ b/app/src/app/admin/page.tsx
@@ -34,7 +34,7 @@ function UnderConstructionToggle() {
       <div className="px-4 py-5 sm:p-6">
         <h3 className="text-lg font-medium text-gray-900">Under Construction Mode</h3>
         <p className="mt-2 text-sm text-gray-500">
-          When on, all public traffic is redirected to an under construction page. Admin pages stay
+          When on, all page views are redirected to an under construction page. Admin pages stay
           accessible. Takes up to a minute to apply.
         </p>
         <label className="mt-4 inline-flex items-center gap-2">
--- a/app/src/app/admin/page.tsx
+++ b/app/src/app/admin/page.tsx
@@ -1,5 +1,51 @@
 'use client';
+import {useEffect, useState} from 'react';
+import {Switch, Text} from '@radix-ui/themes';
 import {pages} from './config';
+import {useCmsFormStore} from '@/app/store/cmsFormStore';
+import {getSiteSettings, updateSiteSettings} from '@/app/utils/api/cms';
+
+function UnderConstructionToggle() {
+  const {session} = useCmsFormStore();
+  const [enabled, setEnabled] = useState<boolean | null>(null);
+  const [error, setError] = useState('');
+
+  useEffect(() => {
+    getSiteSettings({}).then(r => r.ok && setEnabled(r.response.under_construction));
+  }, []);
+
+  const handleToggle = async (checked: boolean) => {
+    setError('');
+    setEnabled(checked);
+    const r = await updateSiteSettings({
+      body: {under_construction: checked},
+      session: session ?? undefined,
+    });
+    if (r.ok) {
+      setEnabled(r.response.under_construction);
+    } else {
+      setEnabled(!checked);
+      setError(`Could not update: ${r.error.detail}`);
+    }
+  };
+
+  return (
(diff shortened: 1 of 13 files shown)


app/src/proxy.ts
```diff
@@ -1,5 +1,28 @@
...
+    const res = await fetch(`${apiUrl}/api/cms/site_settings`);
+    const settings = await res.json();
+    underConstructionCache = {
+      value: settings.under_construction === true,
+      fetchedAt: Date.now(),
+    };
+  } catch {
+    // fail open: keep the site up if the API is unreachable
+    underConstructionCache.fetchedAt = Date.now();
+  }
```
fangge518, line 27: H: Fail open is right when the backend is unexpectedly down, but during a planned migration we expect the backend to be down. During the database down time, the site permanently fails open, which is probably not what we want here. Perhaps add an env override.

```
async function isUnderConstruction(): Promise<boolean> {
  // Env override for planned downtime: forces maintenance mode even when the
  // backend (where the DB flag lives) is itself down. Read at request time,
  // so it applies to every instance from its first request.
  if (process.env.UNDER_CONSTRUCTION === 'true') {
    return true;
  }
```

During the migration, after toggle the db switch on and before the data manipulation. We should

```
fly secrets set UNDER_CONSTRUCTION=true -a districtr-v2-frontend
```
Then start the aws with the env var set as true, unset it after database is ready, and flip the toggle to allow traffic.

app/src/app/admin/page.tsx
```diff
@@ -1,5 +1,51 @@
...
+      setError(`Could not update: ${r.error.detail}`);
+    }
+  };
+
+  return (
+    <div className="bg-white overflow-hidden shadow rounded-lg">
+      <div className="px-4 py-5 sm:p-6">
+        <h3 className="text-lg font-medium text-gray-900">Under Construction Mode</h3>
+        <p className="mt-2 text-sm text-gray-500">
+          When on, all public traffic is redirected to an under construction page. Admin pages stay
```
fangge518: O: Nit. With this toggle on, people can still publicly access the API directly, or through scripts. Perhaps change "all public traffic" to "all page views" to be more precise?

app/src/proxy.ts
```diff
@@ -1,5 +1,28 @@
 import {auth0} from '@/app/lib/auth0';
 
+// ponytail: per-instance 60s cache; a toggle takes up to a minute to reach every visitor
```
fangge518: I: Strip the "ponytail" marker from the comment  (and elsewhere in this PR).

app/src/app/under-construction/page.tsx
```diff
@@ -0,0 +1,22 @@
+import type {Metadata} from 'next';
+import {Flex, Heading, Text} from '@radix-ui/themes';
+
+export const metadata: Metadata = {
+  title: 'Under Construction | Districtr',
+  description: 'Districtr is temporarily unavailable while we make improvements.',
+};
```
fangge518, line 8: I: Nit. We can add `robots: {index: false}` to prevent search engine indexing here.

fangge518 (changes requested):
Thanks for adding this page. Overall, the code looks good. With several lines of minor changes I think this would be good to go.

app/src/proxy.ts
nofurtherinformation: Done in 6d65ae6316144820fac50e65fa7af13bfa8d1bff

app/src/proxy.ts
nofurtherinformation: Handled in 1b56919dd8c310cbaba7a990e599aa60385ea50e

app/src/app/admin/page.tsx
nofurtherinformation: Updated in 1b56919dd8c310cbaba7a990e599aa60385ea50e

app/src/app/under-construction/page.tsx
nofurtherinformation: Done in 1b56919dd8c310cbaba7a990e599aa60385ea50e

fangge518 (approved):
LGTM!

PR #607, opened 2026-07-07
Production stress-test harness (Locust, seed/cleanup CLI, observability, runner) + fix concurrent cold-evaluation race 

nofurtherinformation:
## Description

Implements the production stress test, targeting `api.beta.districtr.org` with a 15-minute simulated influx of 10,000 viewers / 2,500 eval users / 250 editors (~35 rps aggregate). Also fixes a real backend race the harness surfaced during local smoke runs.

- **Locust harness** (`backend/stress_test/`): Viewer/Eval/Editor user classes matching real client behavior — viewers do `GET /api/document` → `GET /api/get_assignments?format=msgpack`; eval users add `GET .../evaluation`; editors create plans via `copy_from_doc` and save with msgpack full-replacement `PUT /api/assignments`, threading `last_updated_at` (409s recorded, non-fatal). Poisson (viewer/eval) and uniform (editor) arrivals pre-sampled with a fixed RNG seed; single `SCALE` env var scales all populations (1% smoke = `SCALE=0.01`); every request tagged `User-Agent: districtr-stress-test/<run-id>`. Editor-created document IDs are flushed write-through to a runtime manifest, so a SIGTERM abort never loses cleanup targets. Deps live in `backend/stress_test/requirements.txt` — nothing added to the prod image.
- **Seed + cleanup CLI** (`backend/cli.py`): `stress-test-seed` creates seed plans over HTTP from the CDN config JSON (slug-validated against `districtrmap` before creating anything), names them with a `[STRESS-TEST]` prefix, and writes a manifest (local or `s3://`). `stress-test-cleanup` deletes by manifest — dropping assignment partitions the same way the reset endpoint does — then offers a confirmed sweep of any leftover `[STRESS-TEST]`-named documents.
- **Observability** (`infra/`): ALB access logging to a new S3 bucket + Athena DDL and canned per-endpoint latency/error queries (`infra/athena/`); a `districtr-prod-stress-test` CloudWatch dashboard (ECS CPU/mem, task count, ALB latency percentiles/5xx, RDS); docs for the temporary SG rule + `/metrics` scrape loop and which existing alarms may fire.
- **Ephemeral runner** (`backend/stress_test/runner/`): `provision.sh` / `run.sh` / `teardown.sh` — one c7i.4xlarge in a prod public subnet (SSM-only, scoped IAM), discovers infra by Pulumi `Name` tags, uploads all artifacts to `s3://<bucket>/stress-test/<run-id>/`, and cleans up after itself including the temp SG rule.
- **Runbook** (`backend/stress_test/RUNBOOK.md`): operator checklist — pre-flight (alarm notification, Sentry sample-rate caveat, RDS snapshot, `pulumi up`, `SCALE=0.01` prod dry run), baseline capture, run monitoring, abort criteria, post-run cleanup.
- **Bug fix — concurrent cold-evaluation race** (`backend/app/evaluation/main.py`): concurrent cache-cold `GET /api/document/{id}/evaluation` requests each ran the full metrics compute and raced on the `Evaluation` INSERT; the loser 500'd with a `UniqueViolation` (~17% of cold evals failed at just 26 users in local smoke). Now a `pg_advisory_xact_lock` keyed on `document_id` serializes computes across all backend tasks: losers wait for the winner's commit and return its cached row — one compute per document, no new infra. `SET LOCAL lock_timeout = '120s'` keeps waiters from tripping the 15s session default. **This fix must be deployed before the full-scale run** (2,500 eval users on 10 cold documents guarantees collisions).

## Reviewers

- Primary:
- Secondary:

## Checklist

  - `backend/stress_test/README.md` (harness usage, seed/cleanup, ECS Exec, runner), `RUNBOOK.md`, `infra/athena/OBSERVABILITY.md`, `STRESS_TEST_PLAN.md`.
  - `backend/tests/test_evaluation_lock.py`: two concurrent cold evaluations on independent connections — asserts the loser blocks on the advisory lock, compute runs exactly once, and both callers get the winner's row. Harness has a self-gating smoke-assert mode (`STRESS_SMOKE_ASSERT=1` fails the run on any non-2xx or a save that doesn't round-trip `updated_at`).

N/A — no frontend changes. Local smoke evidence: seed → `SCALE=0.01` Locust run (290 requests, 0 failures, all 5 routes) → cleanup restored document/partition counts exactly; pre-commit, pytest, and `tsc --noEmit` green.

## Review required

- **The advisory-lock change is the only prod app-code change** — review `update_or_select_document_evaluation` for lock scope (acquired after the cache miss, released by the existing commit/rollback) and the `SET LOCAL lock_timeout` interaction with the 120s `statement_timeout`.
- **Infra changes need a human `pulumi preview --stack prod`** (no AWS creds in the agent environment): expect 4 creates (ALB logs bucket, lifecycle, bucket policy, dashboard) and 1 in-place ALB update. ALB logging must be live before the dry run so Athena has data.
- **Deferred acceptance items** (scripted in `RUNBOOK.md` §1, need operator AWS access): `SCALE=0.01` prod dry run end-to-end (provision → seed via ECS Exec → run → S3 artifacts → cleanup → teardown), and validation of the ECS Exec one-liners.
- **CDN contract assumption**: `assignments_path` in `config.json` is resolved root-relative (`https://tilesets1.cdn.districtr.org/stress-data/...`). The real `config.json` still returns 403 (not yet uploaded), so this is unverified — confirm before the prod dry run.
- Runner IAM grants `ecs:ListTasks/DescribeTasks` on `Resource: *` (needed for the `/metrics` scrape loop's task-IP discovery); acceptable for an ephemeral role but flagging for review.


nofurtherinformation:
Migrating from a noisy but homogenous Poisson process to a Beta curve distribution / Poisson distribution
arrivals svg

nofurtherinformation:
> Migrating from a noisy but homogenous Poisson process to a Beta curve distribution / Poisson distribution 

872f13387a61366bb3a9fc5e154975d0052a9985 implements this change using `random.betavariate` (seeded as `rng`)

backend/app/evaluation/main.py
```diff
@@ -56,6 +49,45 @@ def update_or_select_document_evaluation(
...
+        select(Evaluation).where(Evaluation.document_id == document.document_id)
+    ).one_or_none()
+    if cached := _cached_envelope(evaluation, document):
+        return cached
+
+    # The session default lock_timeout (15s) is shorter than a slow compute;
+    # SET LOCAL scopes the raise to this transaction only.
+    session.execute(text("SET LOCAL lock_timeout = '120s'"))
+    session.execute(
+        text("SELECT pg_advisory_xact_lock(hashtextextended(:doc, 0))"),
```
fangge518: I: Every request that loses this race holds a pooled DB connection — ~15 concurrent cache-cold evals on one document can park a task's entire pool and starve *unrelated* endpoints, the exact pool-exhaustion failure the RUNBOOK watches for. 

Suggestion: try-and-wait instead of blocking — on failure to get the lock, `session.rollback()` (releases the connection back to the pool), sleep 1–2s, re-check the cache, loop with a deadline. Waiters then hold no connection while sleeping, and the `SET LOCAL lock_timeout` bump can be dropped.

backend/stress_test/README.md
```diff
@@ -0,0 +1,247 @@
+# Districtr stress-test harness
+
+Locust harness implementing the traffic model in `STRESS_TEST_PLAN.md` §3:
```
fangge518: PP: `STRESS_TEST_PLAN.md` isn't in the repo, but nine files cite its sections (§3, §5 WS2, WS6, …). It the plan committable to the repo? If not, we probably need to inline the references.

backend/stress_test/RUNBOOK.md
```diff
@@ -0,0 +1,107 @@
...
+- [ ] Tail progress: `aws ssm start-session --target $INSTANCE_ID`, then
+      `tail -f /home/ec2-user/run1.log`. Also watch runner CPU (`top`) — if the
+      single locust process pins a core, note it for WS6 (`--processes 4` next
+      time).
+- [ ] Watch on the dashboard:
+  - **Task count**: expect autoscale 2→N once CPU passes 45% (60 s cooldown).
+  - **DatabaseConnections** vs the ~15/task ceiling (pool 5 + overflow 10);
+    a plateau at 15×tasks with rising latency = pool exhaustion.
+  - **ALB 5xx** (target + ELB) and TargetResponseTime p95/p99.
+  - Known suspects: burst of eval 500s on cache-cold `/evaluation` races,
```
fangge518: PP: nit. This describes the scenario before the advisory-lock commit — that race is fixed with this PR, so an eval-500 burst with the fix is a *finding*, not expected noise. Worth rewording. Related sequencing note: the fix must be live in prod before the run (and the runner pins `origin/main`), so the another merge to main needs to land before stress test.

backend/stress_test/RUNBOOK.md
```diff
@@ -0,0 +1,107 @@
...
+everywhere below.
+
+## 1. Pre-flight (day of run, before provisioning)
+
+- [ ] **Notify alarm recipients** (`alarmEmails` in `Pulumi.prod.yaml`:
+      dhalpern@, gfang@, peter@) of the run window. Likely to fire:
+      `districtr-prod-backend-memory-high` (ECS mem >85%) and
+      `districtr-prod-db-cpu-high` (RDS CPU >80%); each also sends an OK
+      email when it clears. Table in `infra/athena/OBSERVABILITY.md`.
+- [ ] **Sentry quota**: `traces_sample_rate=1.0` is hard-coded in
```
fangge518: I: We can probably make `SENTRY_TRACES_SAMPLE_RATE` an env var (default 1.0, ~3 lines in `app/main.py`). 100% tracing only works while traffic is small — the knob will be needed as prod traffic grows anyway, and in case we need to adjust it for quota issues, the fix is just a config flip instead of a code change + deploy into main.

backend/stress_test/runner/provision.sh
```diff
@@ -0,0 +1,154 @@
...
+INSTANCE_ID=$(aws ec2 run-instances \
+  --image-id "$AMI" --instance-type "$INSTANCE_TYPE" \
+  --subnet-id "$SUBNET_ID" --security-group-ids "$RUNNER_SG" \
+  --associate-public-ip-address \
+  --iam-instance-profile "Name=$NAME" \
+  --metadata-options HttpTokens=required \
+  --block-device-mappings '[{"DeviceName":"/dev/xvda","Ebs":{"VolumeSize":20,"VolumeType":"gp3"}}]' \
+  --user-data "file://$USERDATA" \
+  --tag-specifications "ResourceType=instance,Tags=[{Key=Name,Value=$NAME},{Key=stress-test,Value=true}]" \
+  --query 'Instances[0].InstanceId' --output text)
```
fangge518, line 139: I: Nothing provision.sh creates expires on its own — a forgotten `teardown.sh` leaves a c7i.4xlarge idling (~$17/day) plus the temp 8080 SG rule and live IAM credentials. Cheap backstop: add `--instance-initiated-shutdown-behavior terminate` here and `shutdown -h +720` at the end of user data (12h self-destruct, far beyond any run, cancelable with `shutdown -c`). `teardown.sh` remains the proper cleanup — this just caps the cost of forgetting it.

fangge518 (commented):
I've checked the code under infra/, the documentations, and the evaluation api---everything except for the user simulation with locust. Overall I think the changes look good. There is no absolute blockers, but merely several suggestions.

backend/app/evaluation/main.py
nofurtherinformation: Good catch! Added a less blocking polling strategy in 8bf826a

backend/stress_test/README.md
nofurtherinformation: Good catch -- yes, I decided not to commit the plan when I was wrapping up since this was a process artifact and not relevant to dev ops. I've inlined the relevant info in 161320b996b7b500911901a81c557f5363fb8016

backend/stress_test/RUNBOOK.md
nofurtherinformation: Good catch -- yes since this is resolved we should keep the docs relevant. Clarified in d95d665c0409650d84c614de953a7cfb95f0aeb4

backend/stress_test/RUNBOOK.md
nofurtherinformation: Agreed -- configured in a951769dc4da394d267c0b66f96b461ecfad9c65

backend/stress_test/runner/provision.sh
nofurtherinformation: Nice, this is an elegant backstop :) Added in 23e7737d8547a5fc9afd656fcab19721354526b8

backend/stress_test/scenario.py
```diff
@@ -0,0 +1,61 @@
...
+    """Exactly n session-start offsets in [0, window), distributed as a
+    non-homogeneous Poisson process conditioned on n arrivals, with intensity
+    proportional to a Beta(ARRIVAL_ALPHA, ARRIVAL_BETA) density (spiky early
+    ramp, then decay). Conditioned on the count, arrivals of a non-homogeneous
+    Poisson process are i.i.d. draws from the normalized intensity — i.e. n
+    Beta variates scaled to the window, sorted."""
+    rng = _rng(seed, label)
+    return sorted(
+        rng.betavariate(ARRIVAL_ALPHA, ARRIVAL_BETA) * window for _ in range(n)
+    )
```
peterrrock2, line 46: O: (note for the future) it would be great to try a few different $\beta$ values

fangge518 (approved):
The infra/documentation/eval change look good to me :)

peterrrock2 (approved):
Traffic models are looking good! Did a quick look at the other backend stuff and that all seems in order as well.

Issue #605, opened 2026-07-07
Bug: CSV Uploader incorrectly infers map modules


nofurtherinformation:
# Steps to replicate:
1. Download CSV assignments from Dave's https://davesredistricting.org/maps#viewmap::12a18072-adf1-48ac-a9d1-12280567b824
2. Use uploader
<img width="783" height="300" alt="Image" src="https://github.com/user-attachments/assets/c4bb3e02-b35a-446a-b522-a8e57baa9eb6" />

# Suggested solution:
Just use modifiable district number plan for custom uploads? It is really hard to infer, so we could just make a plan that is titled "Pennsylvania Plan" and use the number of districts in the CSV, enabling modifiable district number

fangge518:
## Action plan

Here's the implementation plan:

### 1. DB migration (prod + dev)
Create new maps for all the districts with `num_districts_modifiable = true` set as true. Note: Currently only `navajo_nation` has this enabled; all other maps have locked number of districts.

### 2. Importer fix (FE, a few lines)
Let importers use the map_modules with flexible number of districts.

nofurtherinformation:
Closed with #608

PR #608, opened 2026-07-07
Prefer custom (modifiable-district) maps in the CSV importer

fangge518:
Part of #605. Built on the v2 modules from #621 (merged).

**Problem:** the CSV uploader infers the target map module from the state FIPS in the uploaded GEOIDs and always lands on the state's congressional map. Those maps have a fixed `num_districts`, so an uploaded plan with a different number of zones (e.g. a state-legislative plan) gets clamped to the congressional district count.

**Fix:**

- The importer's map lookup in `upload.ts` now targets only the state's custom (modifiable-district) map — `{state}_custom_districts_v2`, falling back to the v1 `{state}_custom_districts` slug while pre-v2 environments remain in service — and errors clearly when neither is visible. No fallback to congressional or state-senate maps: their fixed `num_districts` clamping uploads was the bug. The uploader accordingly fetches only the `custom` map group.
- Uploads to a modifiable-district map are no longer clamped to that map's fixed `num_districts`; only the global 538-zone bound applies, and the document's `num_districts` is set to the highest assigned zone after insert.

**Also in this branch:**

- Custom maps' default `num_districts` acts as a **floor**: partial plans keep the default district count and can only grow it, never shrink it.
- **Outlier label detection**: an accidental numeric zone label (>5x a reference value, at most two per upload) is remapped to a low slot instead of inflating the district count; reported and preserved as a district comment. Partial and genuinely spread-out plans are untouched.
- Small fixes: a fixed-district map with `num_districts=NULL` now raises (500) instead of silently uncapping zones; `parseGeoid` classifies split block groups (`-datadem-N`) as `bg` instead of unrecognized; added DC to `US_STATE_META`, closing a gap that broke DC CSV resolution, lowercased DC's community-map label, and (via a companion fix) a dropdown 404 shared with Puerto Rico.

**Scope note:** this branch is importer logic, plus the `US_STATE_META` fix above and its one-line follow-on in the homepage place-picker (`PlaceMap.tsx`) — both discovered testing the importer change live. The `{state}_custom_districts` config and `load_data.py` changes this branch once carried shipped in #621 (v2-based custom maps, group auto-creation, `num_districts_modifiable` pass-through) and were dropped here at the rebase, which also absorbs #625's assignments-table departitioning.

**Deployment constraint:** the module lookup goes through the visible-filtered views API, so this must not reach an environment without a visible custom module per state. AWS dev has the v2 modules visible; Fly dev previews resolve to the v1 custom maps loaded there; prod's v2 modules are loaded but hidden pending the v1→v2 launch flip — promote to prod only after that flip.

## Tested

**Before merge — Fly deployment preview.** Fly's dev DB carries only v1 custom maps today — every check below resolves the *v1* custom slug, not `_v2`. This still exercises the importer logic fully.

- [x] Uploaded a CA block CSV; document landed on `CA Districts` (`ca_custom_districts`, the v1 slug) with the district count following the CSV zones, adjustable.
- [x] A CSV with more zones than the state's congressional count (17-district CO plan) opens with all 17, instead of the previous "Too many districts" error; fewer zones (4-district CO plan) keeps the default 8, reducible manually.
- [x] A CSV with an accidental zone label (e.g. "196" alongside zones 1-8): upload succeeds, count stays at default, "196" reported as remapped and saved as a district comment.

**After merge — AWS dev/prod.** Both AWS backends carry the v2 catalog (unlike Fly); dev is already flipped to v2-visible, prod's is loaded but hidden pending the launch flip.

- [ ] The same CA CSV upload lands on `ca_custom_districts_v2` on dev.districtr.org.
- [ ] `/place/dc`'s community-map button reads "Washington, DC" (was lowercased pre-fix).
- [ ] After the prod launch flip: repeat the CA CSV check against prod.

### DIFF
--- a/app/src/app/components/Uploader/Uploader.tsx
+++ b/app/src/app/components/Uploader/Uploader.tsx
@@ -26,18 +26,13 @@ export const Uploader: React.FC<{
 
   useEffect(() => {
     // Custom (num_districts_modifiable) maps live in the 'custom' group, not
-    // 'states', so they don't clutter the public map picker. Fetch both so
-    // map inference can prefer custom maps with congressional as fallback.
-    Promise.all([getAvailableDistrictrMaps({}), getAvailableDistrictrMaps({group: 'custom'})]).then(
-      ([states, custom]) => {
-        if (states.ok || custom.ok) {
-          setAvailableMaps([
-            ...(custom.ok ? custom.response : []),
-            ...(states.ok ? states.response : []),
-          ]);
-        }
+    // 'states', so they don't clutter the public map picker. CSV imports
+    // target custom maps exclusively.
+    getAvailableDistrictrMaps({group: 'custom'}).then(custom => {
+      if (custom.ok) {
+        setAvailableMaps(custom.response);
       }
-    );
+    });
   }, []);
 
   useEffect(() => {
--- a/app/src/app/components/Uploader/Uploader.tsx
+++ b/app/src/app/components/Uploader/Uploader.tsx
@@ -25,9 +25,19 @@ export const Uploader: React.FC<{
   const inputRef = useRef<HTMLInputElement>(null);
 
   useEffect(() => {
-    getAvailableDistrictrMaps({}).then(result => {
-      if (result.ok) setAvailableMaps(result.response);
-    });
+    // Custom (num_districts_modifiable) maps live in the 'custom' group, not
+    // 'states', so they don't clutter the public map picker. Fetch both so
+    // map inference can prefer custom maps with congressional as fallback.
+    Promise.all([getAvailableDistrictrMaps({}), getAvailableDistrictrMaps({group: 'custom'})]).then(
+      ([states, custom]) => {
+        if (states.ok || custom.ok) {
+          setAvailableMaps([
+            ...(custom.ok ? custom.response : []),
+            ...(states.ok ? states.response : []),
+          ]);
+        }
+      }
+    );
   }, []);
 
   useEffect(() => {
--- a/app/src/app/utils/api/upload.ts
+++ b/app/src/app/utils/api/upload.ts
@@ -85,24 +85,18 @@ const partitionRows = (rows: string[][]): PartitionResult => {
   return {assignments, skippedGeoIds, stateFipsSet};
 };
 
-// Slug patterns must stay in sync with how maps are named in the DB
-// A mismatch will silently break CSV import for that state.
+// CSV imports target the state's custom (modifiable-district) map by naming
+// convention: every deployed environment must have a visible custom module
+// for the importer to work. There is no fallback to fixed-district maps —
+// their locked num_districts silently clamps uploaded plans. The v1 slug is
+// accepted while environments that predate the v2 catalog remain in service.
 const inferMapForCsv = (fips: string, availableMaps: DistrictrMap[]): DistrictrMap | null => {
   const abbr = FIPS_TO_ABBR[fips];
   if (!abbr) return null;
   const lower = abbr.toLowerCase();
-  // Prefer custom_districts maps (num_districts_modifiable=true) so the zone count
-  // from the CSV drives the district count instead of locking to a fixed map.
-  // Fall back to congressional, then name-matching, then state senate for at-large states.
   return (
+    availableMaps.find(m => m.districtr_map_slug === `${lower}_custom_districts_v2`) ??
     availableMaps.find(m => m.districtr_map_slug === `${lower}_custom_districts`) ??
-    availableMaps.find(m => m.districtr_map_slug === `${lower}_congressional_districts`) ??
-    availableMaps.find(
-      m =>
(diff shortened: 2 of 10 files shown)


app/src/app/components/Uploader/Uploader.tsx
```diff
@@ -25,9 +25,19 @@ export const Uploader: React.FC<{
...
+    Promise.all([getAvailableDistrictrMaps({}), getAvailableDistrictrMaps({group: 'custom'})]).then(
+      ([states, custom]) => {
+        if (states.ok || custom.ok) {
+          setAvailableMaps([
+            ...(custom.ok ? custom.response : []),
+            ...(states.ok ? states.response : []),
+          ]);
+        }
+      }
+    );
```
nofurtherinformation: PP: This works, but I wonder if we should just hard code the `{state}_custom_districts` as the target map module and require that the DB by convention be populated for the uploader to work?

app/src/app/utils/api/upload.ts
```diff
@@ -87,16 +87,15 @@ const partitionRows = (rows: string[][]): PartitionResult => {
...
   const abbr = FIPS_TO_ABBR[fips];
   if (!abbr) return null;
   const lower = abbr.toLowerCase();
-  // Prefer the canonical congressional slug, then name-matching, then state senate
-  // as a fallback for at-large states (AK, DE, ND, SD, WY) that have no congressional map.
+  // Prefer custom_districts maps (num_districts_modifiable=true) so the zone count
+  // from the CSV drives the district count instead of locking to a fixed map.
+  // Fall back to congressional, then name-matching, then state senate for at-large states.
   return (
+    availableMaps.find(m => m.districtr_map_slug === `${lower}_custom_districts`) ??
```
nofurtherinformation, line 99: PP: Related to my comment in uploader, should we just enforce it must be the custom districts module or else throw?

backend/management/configs/v1_custom_districts_config.yml
nofurtherinformation, line 1: H: Let's use the v2 modules

nofurtherinformation, line 1: H: This requires the `custom` map group by created, will this config and the load data, as is, handle that automatically or should that be done manually?

backend/management/configs/v1_custom_districts_config.yml
```diff
@@ -0,0 +1,512 @@
...
+# states, since the app requires at least 2 districts) and acts as a floor:
+# uploads can grow the count but never shrink it below this default.
+# gerrydb_views and shatterable_views are omitted — all layers already exist.
+districtr_maps:
+- child_layer: al_districtr_block_view_v1
+  districtr_map_slug: al_custom_districts
+  gerrydb_table_name: al_districtr_view_v1
+  name: AL Districts
+  num_districts: 7
+  num_districts_modifiable: true
```
nofurtherinformation: H: It looks like `load_sample_data` doesn't actually handle `num_districts_modifiable` yet?

nofurtherinformation (commented):
Overall looks great! Nice work! 

A couple questions about if we can deprecate some old required logic and potential tricks with loading the new configs.

backend/management/configs/v1_custom_districts_config.yml
fangge518: Done — this branch no longer ships any custom-maps config: the v2 `{state}_custom_districts_v2` modules landed in #621 and are loaded (visible on dev; dark on prod pending launch). The rebase dropped the v1 config file entirely.

fangge518: Yes, since #621: the loader auto-creates map groups and their display names; verified on the dev and prod batch loads (idempotent re-run creates no duplicates).

backend/management/configs/v1_custom_districts_config.yml
fangge518: Fixed in #621 — the loader passes `num_districts_modifiable` through to `create_districtr_map`; the v2 custom modules are created modifiable.

app/src/app/utils/api/upload.ts
fangge518: Accepted: the importer now resolves only custom modules by convention — `{state}_custom_districts_v2` first, then the v1 `{state}_custom_districts` slug while pre-v2 environments (currently the Fly dev previews) remain in service — and errors with "No custom districts map available for {state}" when neither is visible. No fallback to fixed-district maps, whose locked `num_districts` was the original bug. This also fixes a latent miss: the old preferred branch matched only the v1 slug, which never matched the v2 modules. One deployment note: the lookup goes through the visible-filtered views API, so this shouldn't deploy to prod ahead of the v1→v2 visibility flip (prod's v2 customs are loaded but still hidden).

app/src/app/components/Uploader/Uploader.tsx
fangge518: Done as part of the above — the uploader now fetches only the `custom` map group.

nofurtherinformation (approved):
LGTM!

PR #617, opened 2026-07-09
Draw / Super Draw edit modes + UX batch (elections revert, panel renames, portal routes, CMS place-page ordering)

nofurtherinformation:
## Description

**Draw / Super Draw editing modes**
- The topbar mode switcher becomes **Draw | Super Draw | View | Evaluate**. Both draw entries share the edit route; switching between them while editing is a pure client toggle (no navigation, no save). The preference persists in localStorage (`toolbarStore.superDraw`); **Draw (simplified) is the default**.
- Draw hides super-only surface area: Break/Shatter and Inspector tools (hotkeys included), paint-mask targeting, gear popovers (summary options, choropleth settings, population chart options), side-by-side Comparison mode, per-layer opacity sliders, and advanced settings rows (basemap, block population labels, population-on-map labels, highlight broken precincts, custom district colors). Dropping to Draw normalizes conflicting state (shatter/inspector → pan, side-by-side → overlay).

**Sidebar redesign** (iterated through card-grid and dropdown-stack prototypes with user testing; landed on a hybrid)
- One scrollable **accordion of five sections in both modes** — District overview, Demographics, Elections, Validity check, Overlays — with card-style headers (icon, label, one-line description), framer-motion collapse animation, and open state in `mapControlsStore.sidebarPanels`. Modes gate density *inside* sections, not navigation.
- **Demographics/Elections**: Table | Overlay Layer segmented tabs over a shared `SummaryPanel` (new `sections` prop). Entering an Overlay Layer tab enables the choropleth with that group's last-used (or default) variable; the overlay intentionally **stays on when navigating away** and is toggled from Visual settings ("Toggle overlay layer (demographic | election)", both types offered in Super Draw) via a small `overlayMemory` module.
- **Coalition builder** is attached to Demographics as a full-width gray "Create a coalition (optional)" expander instead of floating as its own tab/card.
- **Validity check**: Contiguity | Completeness segmented tabs. Completeness (ex "Unassigned Areas") drops the show-all button and manual refresh, **auto-refreshes after saves** (keyed on `updated_at`), shows numbered zoom buttons under 10 areas (dropdown otherwise) with labeled Previous/Next and "Zoom to unassigned area" helper text. Contiguity matches (no manual refresh; query already refetches on save) and its table reads District / Number of pieces.
- **Typography/control consistency**: sub-section tabs are size-2 segmented controls; secondary selectors (display mode, summary type, POV, share/count, bar scaling, overlay-mode presets) are size-1 segmented controls; labels are size-2 medium; helper text is size-1 gray.

**Save affordances** (user testing: save was undiscoverable)
- Persistent **topbar save icon** — cloud icon with a red unsaved-changes badge, spinner while saving, tooltip explaining autosave runs after 30 seconds of inactivity. "Save map" leaves the Map actions menu.
- Inline labeled **Save changes** buttons next to tools that read saved state (validation checks kept their own save prompt).

**Visual settings**
- Moved from a Map actions modal to a **dropdown button beside the toolbar** in both modes (left-aligned, size-1, capped popover width, no more toolbar-area horizontal scrollbar). Toolbar-size options removed; labels simplified ("Show" dropped; "District numbers"). Gating: population-on-map labels are Super Draw-only; county-name emphasis is available in Draw.
- Choropleth controls: overlay-mode presets became a segmented control in the panel body ("painted districts" visibility lives with the choropleth); Super Draw gets separate **overlay-layer and districts-layer opacity sliders** (new `zonesOpacity` map option); Draw relies on presets alone.

**Elections data presentation**
- Reverted from `_lean` margins ("D +X%") to **raw two-party vote shares** with a Democratic/Republican POV segmented toggle; cells use the diverging red←white→blue scale keyed to dem share (matching the map) in both POVs, with a note that shares reflect the two major parties only. `X_dem_pct`/`X_rep_pct` rollups replace the removed `_lean` derivation; inspector tooltip and choropleth legend updated ("100% Rep / Even / 100% Dem").

**Interaction & bug fixes**
- **Middle-mouse drag pans the map regardless of the active tool** (manual `panBy`, browser autoscroll suppressed). Custom paintbrush/eraser/break cursor images replaced with standard cursors.
- **Population labels bug**: the parent label layer returned `HIDE_ALL_FILTER` whenever no precinct was shattered, so "show population labels" never rendered on a normal map. Fixed, plus a stale-closure toolbar hotkey listener and a duplicate React key for unassigned/statewide table rows.
- **Custom district colors modal**: with many districts the swatch list filtered to empty, and picking a color truncated the saved scheme to `num_districts` entries (the not-saving bug). Full palette always offered (unused first), duplicates allowed, full scheme preserved.
- Editor topbar title shows the **map module until named**, then the name with a single condensed tooltip (module, data source, "You are drawing {unit}s, which can be broken into {unit}s.", edit hint) replacing the stacked popover + tooltip.

**Tables & wording**
- Panels renamed along the way to final copy: District overview / Demographics / Elections / Validity check / Overlays, each with a one-line description; District overview uses `Component1Icon`.
- Total population listed first (summary selector + Demographics default); uniform percent decimals; tighter table cells vertically centered; unassigned rows show **∅** with a tooltip; Zone → District headers; deviation shows 2–3 decimals with a `<0.001%` floor and "(N people)"; "Democrat" → "Democratic"; choropleth variable selector labeled "Choose Election"/"Choose Population"; overlay opacity slider out of the gear popover; Dave's Redistricting credited for election data on the Data page.

**Routing & CMS**
- Public **tag pages renamed to portal** (`/tag/[slug]` → `/portal/[slug]`, `/tags` → `/portals`) with permanent redirects; footer/admin labels updated; internal `tags` content_type and DB unchanged (no migration).
- **CMS place pages**: the places editor's multi-select is replaced by a type-to-filter **autocomplete with an ordered, reorderable selected list** (`MapMultiSelector`); the place page renders map cards in the saved order. Place pages and the landing page's state-picker section get an **Upload block assignments** button (shared `UploaderModal`).

**Dependencies**: adds `framer-motion` (sidebar/section animations; the previous animation classes referenced keyframes that never existed).

## Reviewers

- Primary: @fangge518 
- Secondary:

## Checklist

  - N/A — no docs affected; UI copy is self-describing.
  - N/A — frontend has Playwright e2e only. Verified via `tsc`, production builds, and extensive live dev-stack sessions (both modes, mode switching, choropleth lifecycle, portal redirects, place-page rendering). Note: e2e selectors referencing the old accordion/testids may need updating since Draw is now the default mode.

## Review required

- **Mode-split judgment calls**: Draw doesn't force-reset hidden Super Draw settings (basemap, custom colors) on switch — only actively conflicting state; paint-mask *release* stays available in both modes; the choropleth persists when leaving Overlay Layer tabs in both modes (toggled off via Visual settings or display-mode None).
- **Elections semantics**: shares are two-party (dem+rep denominator, third parties excluded); deviation floor assumes "0.001" means 0.001 percent; the global percent formatter now always shows one decimal ("20%" → "20.0%") wherever `NUMBER_FORMATS.PERCENT` is used.
- **`zonesOpacity`** multiplies the zone fill expression (`['*', expr, n]`) — sanity-check rendering across basemaps and broken precincts.
- **Mobile**: the mobile top-nav still uses the old `DataPanels` accordion, and Visual settings no longer has a mobile entry point since leaving Map actions — follow-up needed if mobile parity matters for this release.
- **Not exercised live** (dev tiles unavailable): middle-mouse pan feel, cursor appearance, population labels on a painted map, overlay toggles' visual effect, color-modal behavior at 40 districts against a real plan — worth a hands-on pass on preview.
- Pre-existing, untouched: Radix `DialogTitle` a11y warning from the Map actions dialogs.


peterrrock2:
## Big UX batch updates

@nofurtherinformation 

### Notes:

1. Like the reworking of the dropdowns in the sidebar
2. Draw / Superdraw distinction in the topbar dropdown in nice. 
3. Having the save button separate is nice to have.
4. Tabbed layout in the "Demographics" and "Elections" section is more intuitive.
5. Explanations on each of the dropdowns is nice
6. I prefer the hands and cross-hairs to the old icons
7. Middle mouse drag is top tier for anyone with 3D software or QGIS experience
8. Still not able to draw when I set my browser to be the right side of my screen (mobile)

### Nits:

1. (from other user) "Catalog" is confusing and could benefit from an explanation (tool tip)
     - When she clicked on main map, she was worried that she had lost her old map 
2. When clicking on the "Map Layer" in "Demographics" and "Elections", the layer automatically turns on, but it does not turn off when you click back on "Table". I find this unintuitive. It might be worth it to have an "overlay" checkbox in the table section to the right of the "summary type" tabs.
3. "Population on map" is too small even when zoomed in. This would probably be more trouble than it is worth, but it would be nice if we had a few font sized depending on the size of the polygon.
4. Personally, I would prefer if, when switching between draw and super draw,  any super draw settings that are not in draw were disabled. I think it just makes it less confusing if you have something like population on all units. Enabled. It also seems more natural to me to think "oh I need to go to super draw to turn ON some feature" than to think "oh I need to go to super draw to turn OFF some feature" 
5. State module ordering is inconsistent (CO does state house, congressional, state senate, and OK does congressional, state house, state sentate)

nofurtherinformation:
Thanks @peterrrock2 !

1. Worth discussing more!
2. Handled - map no longer comes on automatically, reducing the expectation it might turn off automatically
3. TODO
4. Handled - one-way settings from super draw now use global defaults in basic draw
5. This is a CMS issue not code. However, I've added a CMS ux to make this easier to manage

app/src/app/hooks/useIsDesktop.ts
```diff
@@ -0,0 +1,23 @@
+import {useSyncExternalStore} from 'react';
+
+// Matches Tailwind's `lg` breakpoint, which gates the desktop sidebar.
+const QUERY = '(min-width: 1024px)';
+
+const subscribe = (onChange: () => void) => {
+  const mql = window.matchMedia(QUERY);
```
fangge518, line 7: H: Using width to test for desktop looks good, but right now it introduces a bug:

On a desktop, narrow the window to a point where the mobile layout shows-->Click "District overview" or any side panel tab --> Resize back to full width. The side panel data would still linger on top of the map portion of the window.

app/src/app/globals.css
```diff
@@ -60,20 +60,57 @@ body {
...
+   disabled (i.e. exactly while painting), which let vertical brush strokes
+   scroll and rubber-band the whole page on mobile. */
+.maplibregl-canvas-container {
+  touch-action: none;
+}
+
+/* The canvas is focusable, so every tap/paint drew a focus ring around the
+   whole map. */
+.maplibregl-canvas {
+  outline: none;
```
fangge518: H: These duplicate the configs above.

app/src/app/components/Topbar/ModeSwitcher.tsx
```diff
@@ -164,8 +181,17 @@ export const ModeSwitcher: React.FC = () => {
...
-    // Edit: route straight in when we hold the UUID; otherwise unlock with a password.
-    if (mode === 'edit') {
+    // Draw / Super Draw: same edit route, different client-side toolset flag.
+    // setSuperDraw handles backing out of super-only tools/settings on exit.
+    if (isDrawMode(mode)) {
+      // Only flip (and persist) the mode once we can actually enter the editor —
+      // a cancelled password prompt must not leave superDraw switched on.
+      if (isEditing || editDocId) {
+        setSuperDraw(mode === 'superdraw');
+      }
```
fangge518, line 192: H: If I start at the view mode of a password-protected map and attempt to click superedit mode from there, entering the password correctly will land me in draw mode rather than the superedit mode.

app/src/app/components/Toolbar/Settings.tsx
```diff
@@ -148,40 +137,101 @@ export const ToolSettings: React.FC = () => {
...
+            >
+              {overlayGroups.map(entry => (
+                <CheckboxGroup.Item
+                  key={entry.group}
+                  value={entry.group}
+                  onClick={() => toggleOverlayGroup(entry)}
+                >
+                  Toggle map overlay layer ({entry.label})
+                </CheckboxGroup.Item>
+              ))}
```
fangge518, line 251: I: Overlay group toggles in the visual setting are mutually exclusive, which makes them less like independent checkboxes, but controls for the selection among three states "no overlay/demographic overlay/partisans overlay". Perhaps a dropdown menu selection or a three-state slider switch would be more appropriate here?

Also the per-panel slider does not exactly match the checkbox state here, which is a bit confusing. If I check demographic overlay, the election overlay is automatically unchecked, but the slider in the election panel's menu is still "Overlay".  The election panel's menu's overlay serves a double duty here: passively, it just indicates display mode, but when we actively select it, it will toggle the partisans overlay in addition to changing the display mode to overlay. Should we separate the display toggle and the overlay (or comparison) content selection here? This way the overlay content selection also has a neat correspondance with the checkbox in the visual setting here.

app/src/app/components/Toolbar/Settings.tsx
```diff
@@ -148,40 +137,101 @@ export const ToolSettings: React.FC = () => {
...
+            </CheckboxGroup.Root>
```
fangge518: PP: We have checkboxes for overlay here but no checkboxes for side-to-side comparison. Shall we add them (or remove the overlay checkboxes) for consistency?

app/src/app/components/Cms/ContentEditor/MapMultiSelector.tsx
```diff
@@ -0,0 +1,169 @@
...
+          (m.name.toLowerCase().includes(q) || m.districtr_map_slug.toLowerCase().includes(q))
+      )
+      .slice(0, 10);
+  }, [maps, query, selectedSlugs]);
+
+  const nameFor = (slug: string) => maps.find(m => m.districtr_map_slug === slug)?.name ?? slug;
+
+  const addMap = (slug: string) => {
+    // The `multiple` branch appends, preserving insertion order.
+    handleChange('districtr_map_slugs', true)(slug);
```
fangge518: PP: Nit. The interface of handleChange() is not very intuitive. Since we are adding multiple references to it, perhaps split the multiple branch and the default branch of handleChange() in cmsFormStore and use better method names?

app/src/app/components/sidebar/DataCards.tsx
```diff
@@ -0,0 +1,258 @@
...
+  open,
+  children,
+}) => {
+  const [mounted, setMounted] = useState(open);
+  useEffect(() => {
+    if (open) {
+      setMounted(true);
+      return;
+    }
+    const timeout = setTimeout(() => setMounted(false), 200);
```
fangge518: I: We can extract "200" as a constant and construct "duration-200" from this constant to keep them logically in sync.

app/src/app/utils/demography/overlayMemory.ts
```diff
@@ -0,0 +1,38 @@
...
+
+/**
+ * Last-used choropleth config per layer type, so the Visual settings overlay
+ * toggles can restore it after the overlay is turned off. Module state:
+ * survives panel unmounts and mode switches, resets on page load.
+ */
+export const overlayMemory: {
+  variables: Partial<Record<SummaryType, DemographyVariable>>;
+  lastGroup: SummaryType | null;
+} = {variables: {}, lastGroup: null};
```
fangge518: I: Memory of "overlay layer opacity" does not seem to work when we toggle overlay with the checkboxes in the visual setting. But it is memorized when we toggle them in the panel. Should we adjust overlayMemory's shape here to make the overlay memory more consistent across different toggles?

fangge518 (changes requested):
Thank you Dylan! Our site is really shaping up with this change.

I have listed some unexpected behavior, mostly in edge cases, but I think it should be quite convenient to fix in this change.

app/src/app/hooks/useIsDesktop.ts
nofurtherinformation: Good catch — fixed in b72eed8

app/src/app/globals.css
nofurtherinformation: Fixed in 2a3eb6d

app/src/app/components/Topbar/ModeSwitcher.tsx
nofurtherinformation: Fixed in cf3f431 -- using a transient `pendingSuperDraw` property to handle this case

app/src/app/components/Toolbar/Settings.tsx
nofurtherinformation: Agreed on both points. In general, the control in visual settings is meant as an escape hatch for Draw users if they can't remember how to turn off the demographic layer. There is now a more unified control in both Draw and SuperDraw

app/src/app/components/Toolbar/Settings.tsx
nofurtherinformation: I made this more generic in Super Draw to be `Toggle Choropleth Layer` so this should be handled a bit better. This case should again be a backup for if users forget where the map settings are managed

app/src/app/components/Cms/ContentEditor/MapMultiSelector.tsx
nofurtherinformation: I cleaned this up in abc3546. Since we are moving to a full CMS I think it is not very valuable to update the current CMS too much

app/src/app/components/sidebar/DataCards.tsx
nofurtherinformation: Done in abc3546

nofurtherinformation: Worth noting: Tailwind JIT doesn't work with templated / interpolated strings (eg `duration-{n}`) so I just used the vanilla CSS prop

app/src/app/utils/demography/overlayMemory.ts
nofurtherinformation: Handled in fce12d8

fangge518 (approved):
All my previous concerns have been resolved. The new change looks good to me!

PR #619, opened 2026-07-13
Security audit

nofurtherinformation:
## Description

Security hardening from a full audit of the API surface and AWS stack: edge protection (WAF), a silent-captcha session layer to fast-fail bot traffic, and three quick-win fixes.

- **AWS WAFv2 on the ALB** (`infra/waf.ts`, wired in `index.ts`): per-IP rate limit (2,000 req / 5 min, block), `AmazonIpReputationList` + `KnownBadInputsRuleSet` (block), `CommonRuleSet` in **count mode** — its 8KB `SizeRestrictions_BODY` rule would break msgpack assignment saves, so it soaks in CloudWatch before we flip it to block with exclusions. Optional `wafAllowlistIps` IPSet for the stress-test runner. `BlockedRequests` alarm on the existing SNS topic. Shield Standard applies automatically. ~$11/mo per stack.
- **Silent reCAPTCHA v3 session tokens**: new `POST /api/session` verifies an invisible reCAPTCHA v3 token and mints a stateless 4-hour HS256 JWT (signed with the existing `SECRET_KEY`, no DB). A `require_session` dependency verifies signature+expiry in microseconds and gates writes + expensive endpoints (`create_document`, `assignments` PUT/reset, colors, num_districts, metadata, `stats`, `evaluation`, `unassigned`, `contiguity*`, export, comment-flag). **`SESSION_ENFORCE=false` by default** — invalid/missing tokens only log a warning until the flag is flipped, so nothing breaks on deploy. A `RESEARCH_API_KEY` header (constant-time compared) bypasses for known scripts.
- **Frontend** (`app/src/app/utils/api/session.ts`): lazily mints on first gated call — injects the grecaptcha script once, single-flight refresh, cached in localStorage with a 5-min expiry buffer, attached as `X-Districtr-Session` in both the JSON and msgpack clients, one retry on 401 `session_required`. Fully self-disables when `NEXT_PUBLIC_RECAPTCHA_V3_SITE_KEY` is unset; never blocks the UI on Google.
- **Audit fixes**: removed `print()`s leaking plaintext share passwords + hashes into logs (`save_share/main.py`); fixed `CMSModel.auth` → `.author` filter typo in the public CMS list; added security-headers middleware (HSTS, nosniff, X-Frame-Options DENY, Referrer-Policy).

**Not done in this PR (deploy-time steps):** create a reCAPTCHA **v3** key pair (existing key is v2-checkbox); `pulumi config set --secret recaptchaV3SecretKey` / `researchApiKey` on dev+prod; GitHub secret `RECAPTCHA_V3_SITE_KEY`; after soak, flip CommonRuleSet to block and `SESSION_ENFORCE` to `"true"` (one line in `backendtask.ts` — also the rollback lever).

## Reviewers

- Primary:
- Secondary:

## Checklist

  - Rollout/tuning notes live as comments in `waf.ts` and `backendtask.ts`; no user-facing docs affected.
  - `backend/tests/test_session.py` (8 tests: mint/verify/expired/garbage, enforce on/off, research key, endpoint-level 401). Full backend suites pass (176 tests); frontend `tsc` + `next build` clean; infra `tsc` clean.

N/A — no UI changes; the captcha is invisible by design.

## Review required

- **Gated-endpoint list** in `backend/app/main.py` — judgment call on what requires a session vs. stays open (document GET/list, share, comment creation stay open). Worth a scan for anything over- or under-gated.
- **WAF rate limit (2,000/5min/IP)** — sanity-check against real classroom/NAT usage before prod; sampled requests on the dev stack will show false positives.
- **401-retry flow** in `app/src/app/utils/api/session.ts` + msgpack client — exercise a save in the browser once enforcement is on in dev.
- **Human-only steps above** (v3 keys, Pulumi secrets, GitHub secret) — nothing enforces until these exist, so the merge itself is low-risk.


backend/app/exports/main.py
```diff
@@ -105,7 +106,11 @@ def build_districts_shapefile(
                 zf.write(os.path.join(tmpdir, fname), arcname=fname)
 
 
-@router.get("/api/document/{document_id}/export", status_code=status.HTTP_200_OK)
+@router.get(
+    "/api/document/{document_id}/export",
+    status_code=status.HTTP_200_OK,
+    dependencies=[Depends(require_session)],
+)
```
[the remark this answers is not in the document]
nofurtherinformation: Resolved in cb77a7f895b43ac90d8fee9407bf57baf5c9154c

backend/app/core/security.py
```diff
@@ -129,3 +135,63 @@ async def verify_recaptcha(self, token: str, host: str):
...
+    if (
+        settings.RESEARCH_API_KEY
+        and x_districtr_session
+        and secrets.compare_digest(x_districtr_session, settings.RESEARCH_API_KEY)
+    ):
+        return
+    if x_districtr_session:
+        try:
+            jwt.decode(x_districtr_session, settings.SECRET_KEY, algorithms=["HS256"])
+            return
```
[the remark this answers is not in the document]
nofurtherinformation: Resolved in cb77a7f895b43ac90d8fee9407bf57baf5c9154c

backend/app/main.py
```diff
@@ -271,7 +290,28 @@ async def db_is_alive(session: Session = Depends(get_session)):
...
+
+@app.post("/api/session")
+async def create_session(data: SessionCreate, request: Request):
+    """Mint a stateless session token, verifying reCAPTCHA v3 when configured."""
+    if settings.RECAPTCHA_V3_SECRET_KEY:
+        # Behind the ALB, request.client is the LB node; the user is the
+        # first X-Forwarded-For entry.
+        forwarded = request.headers.get("x-forwarded-for")
+        client_ip = (
+            forwarded.split(",")[0].strip()
```
fangge518: I: The comment says "behind the ALB, request.client is the LB node; the user is the first X-Forwarded-For entry" and takes `forwarded.split(",")[0]`. That's backwards for how ALB populates this header: when a request already carries an XFF header, ALB *appends* the real client IP it saw at the TCP layer to the end of the existing list rather than overwriting it — so the trustworthy entry is the *last* one, and the first entry is exactly what an attacker's original request supplied. As written, a client can set `X-Forwarded-For: 8.8.8.8` on the request and have that value passed straight through as `remoteip` to Google's siteverify call.

Impact here is limited — the reCAPTCHA v3 token itself still has to verify and clear the score threshold, so this doesn't bypass the session gate on its own — but it undermines the accuracy of the one bot-quality signal this endpoint is supposed to add, and the wrong assumption is stated as settled fact in a comment other endpoints may copy later. Worth pulling the last entry instead (or, if there's any chance of another proxy/CDN in front of the ALB later, being explicit about which hop is trusted).

app/src/app/components/Topbar/MapActionsDropdown.tsx
```diff
@@ -31,15 +32,33 @@ export const MapActionsDropdown: React.FC<{
...
-    a.remove();
+    // Fetch via the session-aware client (plain anchor navigation can't attach
+    // the X-Districtr-Session header) and save the blob through a transient
+    // anchor. Filename comes from the backend's Content-Disposition.
+    try {
+      const response = await fetchWithSession(
+        `${process.env.NEXT_PUBLIC_API_URL}/api/document/${exportId}/export?export_type=${exportType}`
+      );
+      if (!response.ok) {
+        console.error('Export failed', response.status);
```
fangge518: O: The previous implementation navigated a bare anchor to the export URL, so a failed request (404, 5xx) at least surfaced the browser's own download-failure UI. The new fetch-based version does `console.error` and returns on a non-OK response or thrown error, with nothing shown to the user — a failed export now just looks like nothing happened. Worth a toast or inline error state, though this is a UX regression rather than a security concern.

fangge518 (approved):
Traced the full enforcement path for the new session gate (`require_session` → JWT decode with `aud`/`exp` required → the research-key bypass), the WAF rule set (priority ordering, rate-limit key selection, the allowlist), and confirmed the CMS typo fix and export CORS header exposure against other call sites. One item worth addressing (non-blocking, since nothing in this PR is enforced yet by default), plus a separate UX-only note.

backend/app/main.py
nofurtherinformation: Good catch - fixed in 74a468f00d4b1db2e27b46fa302a471cd15c6907

app/src/app/components/Topbar/MapActionsDropdown.tsx
nofurtherinformation: Made louder in 1538810d913a39a61ccd239aa81574132e54b5c5 !

PR #620, opened 2026-07-14
Dot density demographic layer

nofurtherinformation:
## Description

- Adds a **dot density population map** as a new "Dot density" option under Display mode: classic multi-color racial dot map (one dot ≈ N people, colored by race category, Okabe-Ito palette), rendered entirely on the GPU from the vector tiles the client already downloads. No new pipeline outputs, no raster tiles, no deck.gl — ~1,100 lines, all frontend.
- Density is a live join: dots re-color and re-density **instantly** on demography changes (TOTPOP ↔ VAP, data refresh) because per-feature race densities live in small GPU textures rebuilt from the client-side Arquero demography table — geometry is never re-fetched or re-tessellated.
- **Shattering-aware**: shattering a VTD zeroes its densities (texture update only) and decodes just the exposed child blocks (path-filtered), which then stipple at block resolution; healing reverses both.
- Legend with category swatches and a zoom-tracking "1 dot = N people" caption in the sidebar MapPanel, plus a new **floating on-map legend for mobile viewports** (hidden ≥ `md`, same convention as `MobileTopNav`) that also serves the existing choropleth modes — the MapPanel legend was extracted into a shared self-contained `ChoroplethLegend`.
- Debug gate `?dotdensity=1` mounts the layer regardless of display mode.

### GPU strategy

The layer is a MapLibre `CustomLayerInterface` (WebGL2 — guaranteed by maplibre-gl 4.x). The trick is that **no dot exists as geometry**; dots are generated per-fragment:

1. **Geometry**: a Comlink worker (`DotDensityWorker`) fetches tiles from the same PMTiles archive MapLibre reads (`pmtiles` was already a dependency), decodes MVT (`@mapbox/vector-tile` + `pbf`), tessellates polygons with `earcut`, and transfers per-tile GPU buffers: tile-local `[0,1]` vertex positions, triangle indices, a per-vertex feature index, plus feature `path` ids and tile-clipped mercator areas.
2. **Stippling**: the fragment shader lays a global power-of-two grid over the mercator `[0,1]²` world (cell size `2^-n`, `n` from zoom targeting ~7 px cells). Each cell's integer world id feeds a `pcg2d` uint hash → jittered dot center (jitter margin keeps discs inside cells: no overlap, complete circles), Bernoulli thinning against the feature's total-population density (`expected = density × cellArea / peoplePerDot`), category selection from an independent hash channel over the cumulative category densities, and a `fwidth`-based anti-aliased disc edge.
3. **Stability & precision**: cell ids are world-stable integers ≤ 2^24 (exact in fp32), so dots never move under pan/zoom and match identically across tile borders. Vertices stay tile-local; the tile→clip matrix is composed in float64 on the CPU so fp32 never sees a full mercator coordinate — no jitter at deep zoom.
4. **Tile borders**: tippecanoe's buffered geometry would double-stipple at seams, so fragments outside the tile's `[0,1]` square are discarded; tiles then partition the plane exactly and edge dots are completed by the neighbor (same world hash).
5. **Data join**: per-tile `RGBA32F` textures hold 2 texels per feature (6 category densities + total). Rebuilt from `demographyService`'s table on `dataHash`/variable/shatter changes; a texture upload is the *only* cost of any data change.
6. **LOD**: people-per-dot halves per zoom level (`2^(22-n)`, ~250 at state view → 1 person/dot from ~z15), so cities saturate at low zoom and resolve into individuals as you zoom in — standard dot-map behavior.

### Major computational costs

- **Network**: PMTiles range reads for the viewport cover at data zoom (3–12), ≤ 64 tiles, ≤ 1.5 MiB/tile. Double-fetches the archive MapLibre already reads, but HTTP cache + PMTiles directory caching make warm loads cheap.
- **Worker CPU (per tile, off main thread)**: MVT decode + earcut tessellation — the dominant one-time cost. CO VTD tiles (~3k features) decode in tens of ms; the worst case is dense block-level tiles (~50–70k polygons/tile), currently only decoded path-filtered for shattered children, so small in practice. A full-blocks state (TX) has not been perf-tested.
- **Main thread**: one Arquero table scan per demography change to build the `path → counts` index (O(rows), ~10³–10⁴ for VTDs), then O(features-in-tile) texel fills. No per-frame main-thread work beyond uniform setup.
- **GPU memory**: per tile — vertex positions (8 B/vertex) + feature indices (4 B/vertex) + u32 indices, roughly 0.5–3 MB per dense VTD tile; density textures are negligible (32 B/feature).
- **GPU per frame**: one draw call per resident tile (≤ ~10 typical); fragment cost is 2 `texelFetch`es + an integer hash + a `smoothstep` per covered pixel — roughly one screen of overdraw total, ~1 ms-class on integrated GPUs, visually smooth panning at z10–z16.
- **What's free**: variable/universe switches, shatter/heal, and repaints reuse all geometry — texture re-upload only.

### Known limits (deliberate, marked with `ponytail:` comments)

- Features spanning tile borders use tile-clipped areas → slight dot overcount along seams (fix: cross-tile area registry).
- Dot positions re-randomize when the zoom crosses a grid-level boundary (fix: cross-fade between adjacent levels).
- Ultra-dense cells clamp at one dot per cell and read as a regular grid patch (visible on some downtown blocks).
- Districts mode only; COI mode would need the `coiAssignmentsStore` variant. Not mounted on public read-only side-by-side demographic views.

## Reviewers

- Primary:
- Secondary:

## Checklist

  - N/A — experimental prototype; architecture documented in code comments and this PR.
  - N/A — rendering-heavy code with no existing GL test harness; verified end-to-end in the browser. The pure math (`tileMath.ts`) is the natural target if we productionize.

Screenshot 2026-07-14 at 8 10 10 AM

Screenshot 2026-07-14 at 8 11 08 AM

## Review required

- **Visual QA on real hardware**: dot readability/density at state, metro, and street zooms (one constant, `DOT_DENSITY_PPD_ANCHOR_LEVEL`, tunes overall density); Safari and Firefox untested; map pitch untested.
- **Perf on a blocks-heavy state** (TX/CA): worker decode time and GPU buffer sizes for parent-level tiles at z12.
- **Category semantics**: confirm the six-category breakdown and Okabe-Ito color assignments are appropriate (Hispanic overlaps race columns per census definitions — we render the columns as-shipped in the tabular data).
- **Shatter interaction**: dots swap parent→block resolution on shatter; confirm the UX reads correctly alongside zone painting.
- **Mobile legend**: placement (bottom-left, above attribution) and the lack of a dismiss control; it also now serves overlay/comparison choropleths on small screens.
- WebGL context loss (laptop GPU switch) is not handled — the layer re-adds on `styledata` but not on context restore.


nofurtherinformation:
This viz strategy is shockingly effective for identifying zero population blocks while painting 
Screenshot 2026-07-14 at 8 19 01 AM

PR #621, opened 2026-07-14
Add v2 map modules for 52 geographies (50 states + DC + PR)

fangge518:
A new release of source geopackages (v2) is on S3 under `gerrydb/`, named uniformly `{geo}_districtr_{layer}_view_v2.gpkg`, for 52 geographies: the 50 states plus DC and PR (both new — v1 covered 50 states only). This PR adds the config and loader changes to bring v2 online as a v1-shape batch import: parent layer (vtd, or `bg` for CA/HI/OR/WV, which have no v2 vtd source — same exception v1 had — and for ME, whose v2 vtd layer does not cover the unorganized territories, leaving most blocks there parentless) + block layer, per geography.

**The v2 data fix:** discontiguous units in the source data were split into contiguous parts with `-datadem-N` suffixed geo_ids — this applies to block groups as well as VTDs (confirmed in `hi_districtr_bg_view_v2.gpkg`: `150099800001-datadem-1`/`-2`); all v2 VTD geometries are single-part (verified on sampled states; BG single-part status was not independently checked). No special handling was needed in the configs — ogr2ogr imports the view as-is — but one backend fix was: the `BLOCK_GROUP` predicate in `backend/app/utils.py` (`GEOID_PREDICATES`) only accepted bare 12-digit ids, so split-BG parents misclassified in the evaluation context's parent-vs-child check and `infer_geo_unit_type` could fail if its single sampled row was a split part. It now matches `\d{12}(-datadem-\d+)?`; the VTD predicate (prefix check) was already suffix-safe, and BLOCK stays strict. Parametrized unit tests cover all three predicates with plain and suffixed ids.

**What's here:**
- `backend/management/configs/v2_backend_config.yml` — the whole v2 catalog in one file, matching v1's single-file shape (imports, shatterable views, map modules, groups).
- `pipelines/tilesets/configs/v2_tileset_config.yml`, `pipelines/tabular/configs/v2_tabular_config.yml`, `pipelines/transforms/configs/v2_graph_config.yml` — 52-geography batch configs mirroring the v1 batch configs.
- `backend/management/load_data.py` — ports PR 608's edge-copy speedup for parent-child edges.
- `backend/management/load_data.py` — loader now auto-creates map groups and their display names.
- `backend/app/models.py` — documents the `DistrictrMap.visible` contract (gates picker discovery only).

A few facts worth flagging: CA is now included in the graph batch — v1 excluded it because its stripped-down CA gpkg lacked the graph pipeline's input layer, and the v2 CA gpkg ships the full layer set. The `load_data.py` edge-copy port also fixes a pre-existing `view.districtr_map_uuid` `AttributeError`. And the loader's auto-created groups, plus passing `num_districts_modifiable` through to `create_districtr_map`, remove the manual `create-map-group` step PR 608 needed.

**Deployment ordering:**
1. Load `v2_backend_config.yml` via `batch-create-districtr-maps` — one run covers imports, shatterable views, all 198 modules, and groups (created automatically — no manual group-creation step). Config-only reruns skip the slow imports with `--skip-gerrydb-loads`.
2. Verify the v2 modules in the running app.
3. Retire v1 from the picker: a set-based `UPDATE districtrmap SET visible = false` selecting v1-generation maps structurally (`parent_layer LIKE '%\_districtr\_%\_view\_v1'`), preceded by a pre-flight `SELECT count(*)` on the same predicate that the operator compares against the expected count (143 state modules; 193 if the v1 custom maps are loaded) before updating. This is a separate, explicit operator step — not run automatically by loading any config here. Retirement also removes v1 maps from the daily graph-integrity check, which monitors visible maps only: documents on v1 maps keep working, but their graphs are no longer watched. Deliberate — v1 enters sunset at retirement, not indefinite support.

**Picker mechanism:** `GET /api/gerrydb/views` (backend/app/main.py, `get_projects()`) filters on `DistrictrMap.visible` within a `group_slug` ("states" by default, matching the frontend's default query). v2 modules are added to the *same* `states` group v1 uses, with the default `visible = true`, so they show up in the picker as soon as they're loaded — no frontend change needed. Setting v1 rows' `visible = false` hides them from the picker only; documents already created on v1 maps are looked up directly by slug (not filtered on `visible`) and keep working.

## Review required
- Proposed module set — DC: one map, Council wards (8); PR: Senate districts (8) + House districts (40); no congressional map for either (Delegate/Resident Commissioner are at-large). Reasonable?
- Should DC/PR appear in the same "states" picker group as the 50 states, or do you want a separate grouping?


nofurtherinformation (approved):
LGTM!

peterrrock2 (approved):

PR #623, opened 2026-07-15
Perf/stress test findings revisions

nofurtherinformation:
## Description

Stress test `run1` (2026-07-15, 12,750 simulated users vs api.beta.districtr.org) failed 94% of requests while compute sat idle (ECS CPU 15%, RDS 17%). Every failure was config, not hardware: autoscaling watched CPU (which an I/O-bound workload never raises), each task's DB pool defaulted to 15 connections, and the anyio threadpool capped sync routes at 40.

- **Autoscale on request count** (`infra/backend.ts`, `infra/config.ts`): adds a second target-tracking policy on `ALBRequestCountPerTarget` alongside the existing CPU policy — ECS scales out when either fires. Target is 600 req/min/task, overridable per stack via the new `backendRequestsPerTarget` config. Run1's traffic (~390 req/min/target) would have triggered scale-out immediately; instead the fleet stayed at 2 tasks the whole run.
- **Raise per-task DB pool** (`backend/app/core/db.py`): `pool_size=40, max_overflow=20` (60/task vs SQLAlchemy's default 15, which was run1's hard concurrency wall). At 60/task, RDS `max_connections` (~900) allows ~14 tasks — above the current max of 6.
- **Raise anyio threadpool limiter to 80** (`backend/app/main.py`): the default 40 tokens would have capped sync-route concurrency below the new pool; the extra 20 threads cover non-DB work (S3 graph loads, evals).
- **Graph LRU cache 10 → 15** (`backend/app/evaluation/graph.py`): run1 touched ~11 distinct maps against a maxsize of 10, so evictions forced multi-second cold S3 reloads.
- **Guard `RESULTS_BUCKET` in the stress runner** (`backend/stress_test/runner/provision.sh`, `run.sh`): fail fast if it matches `*alb-logs*` — run1's IAM policy was accidentally provisioned against the ALB access-logs bucket.

## Reviewers

- Primary: @fangge518 
- Secondary:

### DIFF
--- a/backend/app/core/db.py
+++ b/backend/app/core/db.py
@@ -6,6 +6,10 @@ from app.core.config import settings
 engine = create_engine(
     str(settings.SQLALCHEMY_DATABASE_URI),
     echo=settings.ECHO_DB,
+    # 60/task; RDS max_connections ~900 caps us at ~14 tasks. Keep in step
+    # with the anyio threadpool limiter (app/main.py lifespan).
+    pool_size=40,
+    max_overflow=20,
     pool_pre_ping=True,
     pool_recycle=3600,
 )
--- a/backend/app/evaluation/graph.py
+++ b/backend/app/evaluation/graph.py
@@ -53,7 +53,9 @@ def get_gerrydb_graph(file_path: str) -> Graph:
         return pickle.load(f)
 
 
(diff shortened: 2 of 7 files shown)


fangge518 (approved):
LGTM.

PR #625, opened 2026-07-16
Departition assignments tables + fix stress-test viewer traffic model

nofurtherinformation:
## Description

Stress testing against beta showed ~93% request failure at 12,750 simulated users with both app and DB CPU idle. Root cause: `create_document` ran `CREATE TABLE ... PARTITION OF document.assignments` per document, which takes an ACCESS EXCLUSIVE lock on the parent table — so every document creation (and every reset, which did DROP+CREATE) globally blocked all assignment reads and writes. Under load this became a lock convoy: connections piled up on the parent-table lock, pools exhausted, and everything hit the 120s ALB timeout. Confirmed via RDS Performance Insights (lock waits concentrated on `get_assignments`) and a viewers-only run where removing writes dropped failures from ~92% to ~34%.

- **Migration `7e57b49573e0`**: converts `document.assignments` and `document.community_assignments` from per-document LIST partitions to plain tables (copy → drop old parents CASCADE → rename → rebuild composite PKs post-load → ANALYZE). Reads/writes now take row locks only; document creation does no DDL. Also drops the legacy `create_document(TEXT)` / `create_assignment_partition_sql(TEXT)` DB functions, which are unused and would break against plain tables. **Irreversible** — downgrade raises; rollback is a DB snapshot. Run at low traffic: the migration holds an exclusive lock for the copy + index build (~1s at 260k rows locally; pre-flight row counts on beta/prod before deploying). Fixed HASH partitioning was measured and rejected: the composite-PK lookup is ~4ms of a ~115ms `get_assignments` query (the rest is the `parentchildedges` join), so partitioning the cheap part buys nothing.
- **`main.py`**: partition create/reset helpers deleted; the reset endpoint is now a parameterized row `DELETE` (response shape unchanged, message text updated).
- **`models.py` / alembic `env.py`**: dropped the partition declarations, the phantom `document_geo_id_unique` UniqueConstraint (it was actually the PK in the live DB), the duplicate `community_assignments` unique constraint, and the autogenerate exclusion hacks. The two `community_assignments` secondary indexes are intentionally not recreated — every query filters `document_id` first, so the PK prefix covers them and the save path sheds index write amplification.
- **`thumbnails/main.py`**: was the only code addressing partitions by name; now queries the parent table (document_id predicate in the LEFT JOIN's ON clause to preserve unassigned-geo rows).
- **Stress harness traffic model fix**: the Viewer and EvalUser classes called `get_assignments` (306 KB, heaviest read endpoint), but real viewers call `/api/document/{id}/stats` (~1 KB) and eval users call `/evaluation`. 12,500 of 12,750 simulated users were driving synthetic load, inflating every prior run. Viewer → `get_document` + `/stats`; EvalUser → `get_document` + `/evaluation`; editor path unchanged. Stress seed/cleanup now deletes rows instead of dropping partitions.

## Reviewers

- Primary:
- Secondary:

## Checklist

  - Migration docstring documents the lock rationale and the alternatives measured; endpoint behavior unchanged.
  - No new tests: the test suite builds its DB via `alembic upgrade head`, so the migration is exercised from scratch; existing reset/copy/shatter/export/thumbnail tests cover every touched path (293 passed, 117 skipped).

N/A — backend/schema only.

## Review required

- **The migration** (`7e57b49573e0_departition_assignments_tables.py`) is the high-stakes piece: it rewrites both assignment tables under an exclusive lock and is irreversible. Verify the copy-swap-drop sequence and the deploy plan (low-traffic window; old tasks briefly 500 on create/reset until the rolling deploy replaces them).
- **Intentional index/constraint removals** on `community_assignments` — confirm no out-of-band query depends on bare `geo_id`/`community_id` lookups.
- **Thumbnail SQL rewrite** — the document_id predicate moved into the LEFT JOIN's ON clause; covered by `test_thumbnails`, but worth a second pair of eyes.

### DIFF
--- a/backend/app/alembic/env.py
+++ b/backend/app/alembic/env.py
@@ -86,12 +86,8 @@ def include_object(object, name, type_, reflected, compare_to):
 
     if name and (
         name in POST_GIS_ALPINE_RESERVED_TABLES
-        or re.match(r"document.assignments_.+", name)
-        or re.match(r"document.community_assignments_.+", name)
         or re.match(r"parentchildedges_.+", name)
         or re.match(r".*_districtr_view+", name)
-        # For whatever reason alembic fails to recognize it already exists
-        or name == "document_geo_id_unique"
     ):
         return False
     return True
--- a/backend/app/alembic/versions/7e57b49573e0_departition_assignments_tables.py
+++ b/backend/app/alembic/versions/7e57b49573e0_departition_assignments_tables.py
@@ -1,0 +1,116 @@
+"""departition assignments tables
+
+Replace per-document LIST partitioning of document.assignments and
+document.community_assignments with plain tables. CREATE TABLE ... PARTITION OF
+takes an ACCESS EXCLUSIVE lock on the parent, so every document creation (and
+the reset endpoint's DROP+CREATE) globally blocked all assignment reads and
+writes — the lock convoy behind the July 2026 stress-test collapses (~93%
+failure with app and DB CPU idle; RDS lock waits concentrated on
+get_assignments). Plain tables take row locks only.
+
+Alternatives measured against the resulting big table (260k rows dev / EXPLAIN
+ANALYZE on the largest document, 38k rows): the composite-PK lookup resolves a
+document in ~4ms of a ~115ms get_assignments query — the rest is the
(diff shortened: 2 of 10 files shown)


peterrrock2 (approved):
LGTM!

PR #629, opened 2026-07-17
Overlays update - CLI and FE county mask

nofurtherinformation:
## Description

Adds a re-runnable CLI command for attaching overlays to maps, and a county paint mask for Super Draw.

**Backend**:
- New `link-overlays-to-maps` CLI command — idempotently links overlays to maps (`INSERT … ON CONFLICT DO NOTHING`), selecting overlays by name, source URL, id,  or `--all-overlays`, and filtering maps by `--statefps` (array-overlap against `districtrmap.statefps`). Selecting by source URL exists because overlay display names repeat across states while source URLs are unique, so per-state scripts can re-link without ids. Safe to re-run any time to pick up newly created maps; maps with NULL `statefps` never match a filter.

**Frontend — county paint mask (Super Draw)**:
- The built-in county boundaries pseudo-overlay now supports the same choose-an-area-to-paint-within mask as real overlays: mask button on the county row (Super Draw only), county branch in `setPaintConstraint` (`querySourceFeatures` on the county source with a `GEOID` filter; `promoteId` added so feature ids resolve), and click/hover routing to the always-on `counties_fill` layer via a shared `getSelectingClickLayer` helper.
- Matching visuals: feature-state hover highlight while selecting, semi-opaque white dim over non-selected counties, and an orange outline on the selected county. Turning the county layer off releases an active county mask.
- Enforcement reuses the existing geometry-based `filterFeatures` path unchanged.

## Reviewers

- Primary: @fangge518 
- Secondary:

## Checklist

- [x] Added/Updated related documentation
      (CLI `--help` text documents all selectors and the statefps filter).
- [x] Added/Updated related unit tests (5 CLI tests for `link-overlays-to-maps`:
      by name, by source, statefps filtering, idempotent re-run, invalid selectors —
      10/10 passing in `backend/tests/test_cli.py`).

## Review required

- **Linker semantics**: `--statefps` uses array overlap (`&&`) against `districtrmap.statefps`, and NULL-statefps maps are deliberately excluded from filtered links — confirm that matches how prod map modules are populated.
- **UI/UX validation**: county mask feel in Super Draw — hover highlight while selecting, dim opacity (0.75), selected outline, and that releasing the mask (button or county-layer toggle) behaves as expected alongside the existing  County Brush.

### DIFF
--- a/app/src/app/components/sidebar/OverlaysPanel.tsx
+++ b/app/src/app/components/sidebar/OverlaysPanel.tsx
@@ -18,18 +18,7 @@ import {useMemo} from 'react';
 import {useMapStore} from '@/app/store/mapStore';
 import {COUNTY_SOURCE_ID} from '@/app/constants/map/layerIds';
 import type {Overlay} from '@/app/utils/api/apiHandlers/types';
-import {OverlayMetadataModal} from './OverlayMetadataModal';
-
-// Enacted-plan datasets (from the source-URL suffix, e.g. `al_cd` -> "cd") are
-// grouped under a "Legislative Districts" heading; everything else follows.
-const LEGISLATIVE_DATASETS = new Set(['cd', 'sldu', 'sldl', 'sld']);
-const datasetKey = (source: string | null) =>
-  source
-    ?.split('/')
-    .pop()
-    ?.replace(/\.geojson$/, '')
-    .split('_')
-    .pop() ?? '';
+import {OverlayMetadataModal, isLegislativeOverlay} from './OverlayMetadataModal';
 
 export const OverlaysPanel = () => {
   const availableOverlays = useMapStore(state => state.mapDocument?.overlays ?? []);
@@ -62,12 +51,8 @@ export const OverlaysPanel = () => {
     return fastUniqBy(sortedOverlays, 'name');
   }, [availableOverlays]);
 
-  const legislativeOverlays = overlaysGroupedByName.filter(o =>
-    LEGISLATIVE_DATASETS.has(datasetKey(o.source))
-  );
-  const otherOverlays = overlaysGroupedByName.filter(
-    o => !LEGISLATIVE_DATASETS.has(datasetKey(o.source))
-  );
+  const legislativeOverlays = overlaysGroupedByName.filter(isLegislativeOverlay);
+  const otherOverlays = overlaysGroupedByName.filter(o => !isLegislativeOverlay(o));
 
   const renderOverlayRow = (overlay: Overlay) => (
     <Flex key={overlay.overlay_id} justify="between" align="center" gap="2">
--- a/app/src/app/components/sidebar/OverlaysPanel.tsx
+++ b/app/src/app/components/sidebar/OverlaysPanel.tsx
@@ -14,6 +14,7 @@ import {getFeaturesInBbox} from '@utils/map/getFeaturesInBbox';
 import {fastUniqBy} from '@/app/utils/arrays';
 import {useMemo} from 'react';
 import {useMapStore} from '@/app/store/mapStore';
+import {COUNTY_SOURCE_ID} from '@/app/constants/map/layerIds';
 
 export const OverlaysPanel = () => {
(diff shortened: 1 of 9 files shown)


backend/cli.py
```diff
@@ -856,6 +981,81 @@ def delete_overlay(session: Session, overlay_id: str):
...
+
+    updated_count = 0
+    for overlay in overlays:
+        key = overlay.source.rsplit("/", 1)[-1].removesuffix(".geojson")
+        entry = metadata_by_key.get(key)
+        if entry is None:
+            logger.info(f"No metadata entry for {key}; leaving overlay unchanged")
+            continue
+
+        new_name = entry.get("name")
```
fangge518: I: Use `entry["name"]` / `entry["description"]` instead of `.get()`. `overlay_metadata.json` is self-authored, so a missing key means a malformed entry, not an intentional partial update — `[]` raises `KeyError` right at the lookup instead of letting `None` flow through to `UPDATE ... SET name = NULL`, which currently fails `Overlay.name`'s `NOT NULL` constraint several lines later with a much less legible Postgres error.

backend/cli.py
```diff
@@ -670,6 +670,131 @@ def add_overlay_to_map(session: Session, districtr_map_slug: str, overlay_id: st
...
+
+    # Dedupe while preserving order
+    resolved_overlay_ids = list(dict.fromkeys(resolved_overlay_ids))
+
+    statefps_filter = list(statefps) if statefps else None
+
+    stmt = text(
+        """INSERT INTO districtrmap_overlays (districtr_map_id, overlay_id)
+        SELECT uuid, CAST(:overlay_id AS uuid) FROM districtrmap
+        WHERE (CAST(:statefps AS varchar[]) IS NULL OR statefps && CAST(:statefps AS varchar[]))
```
fangge518, line 785: O: FYI, not something to fix here — while reviewing this we found every `_v2` `districtrmap` row (all 50 states + DC + PR, the new modules from #621) had `statefps = NULL`, which would have made `--statefps`-filtered runs silently match zero v2 maps (`statefps && ...` is never true against `NULL`). Already backfilled on both dev and prod, so this shouldn't cause any surprises. No changes needed on your end — just flagging so the history here makes sense if this comes up again.

app/src/app/components/sidebar/OverlaysPanel.tsx
```diff
@@ -73,71 +121,65 @@ export const OverlaysPanel = () => {
           </Flex>
         </Callout.Root>
       )}
+
+      {/* Congressional + state legislative districts, grouped under a heading */}
+      {legislativeOverlays.length > 0 && (
+        <>
+          <Text size="2" weight="bold">
+            Legislative Districts
```
fangge518, line 129: PP: Consider "Enacted Districts" instead of "Legislative Districts." Two reasons: (1) "legislative" colloquially tends to mean the *state* legislature specifically, but the group (`LEGISLATIVE_DATASETS = {cd, sldu, sldl, sld}`) includes congressional too — "enacted" applies cleanly to both without that narrowing. (2) It's the standard redistricting-community term for "the currently legally-effective plan" as opposed to a drawn/ proposed one — the exact distinction this heading is trying to draw against a user's own districts — and it matches the framing your own metadata entries already use (`plan_name`, `year`, `modified`). Non-blocking either way. (Same heading also appears in OverlayMetadataModal.tsx:201.)

fangge518 (approved):
Nice, re-runnable ops command and a clean retrofit of the county layer onto the existing overlay-selection machinery. A couple of non-blocking suggestions:

app/src/app/components/sidebar/OverlaysPanel.tsx
nofurtherinformation: Thanks! _Enacted_ in 71992e7bb33d1d00d5c4c03f159656573cfe0ade

backend/cli.py
nofurtherinformation: Good call - updated in 002f1e9144c1504184b168f4a3024af2b37b0858

PR #634, opened 2026-07-21
Make undo/redo per-gesture instead of time-throttled

nofurtherinformation:
Drop the 3s MIN_DIFF_MS throttle in temporalDiff: it didn't merge rapid edits into one undo step — it silently stored no snapshot at all for any edit within 3s of the previous one, so a burst of clicks was un-undoable individually and could even swallow the pre-shatter snapshot. Every paint gesture already funnels through exactly one ingestAccumulatedAssignments set(), so one snapshot per clientLastUpdated bump is naturally one per click/stroke.

To keep timestamp-only bumps (comment/metadata edits, save syncs) from creating dead undo steps, snapshot only when a tracked collection ref actually changed — all mutation paths replace Maps/Sets wholesale, so ref inequality is a reliable O(keys) change check.

Also null out pendingShatterUndoState in the stored pre-shatter entry so restoring it can't leave a stale pending snapshot via partial merge.

### DIFF
--- a/app/src/app/constants/document/temporal.ts
+++ b/app/src/app/constants/document/temporal.ts
@@ -6,8 +6,5 @@
  */
 export const OFFSET_FACTOR: number = 15;
 
-/** Minimum milliseconds between undo/redo history snapshots. */
-export const MIN_DIFF_MS = 3000;
-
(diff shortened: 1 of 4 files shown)


fangge518 (approved):
LGTM.

PR #635, opened 2026-07-21
Refine district count editing: draft input, confirm dialog

nofurtherinformation:
- Input now holds a local draft so it can be emptied while typing; the value commits on blur or Enter and snaps back if invalid.
- All paths (+/- buttons and input) route through one handler that clamps to MIN/MAX and removes assignments above the new count on decrease — previously the input path left stale assignments in place.
- Decreases that would delete existing assignments prompt an AlertDialog before applying.

### DIFF
--- a/app/src/app/components/Toolbar/DistrictZonePicker.tsx
+++ b/app/src/app/components/Toolbar/DistrictZonePicker.tsx
@@ -20,6 +20,8 @@ import {
 } from '@/app/constants/document/limits';
 import {MinusIcon, PlusIcon, Pencil1Icon} from '@radix-ui/react-icons';
 import {ACCESS_STATES} from '@constants/document/state';
+import {temporalManager} from '@/app/utils/temporal';
+import {MAP_MODES} from '@constants/map/mode';
 
 export const DistrictsZonePicker: React.FC = () => {
   const [showNumDistrictEditor, setShowNumDistrictEditor] = useState(false);
@@ -54,15 +56,21 @@ export const DistrictsZonePicker: React.FC = () => {
     setNumDistricts(newNumDistricts);
     if (newNumDistricts < numDistricts) {
       removeAssignmentsForZonesAbove(newNumDistricts);
+      // num_districts isn't part of the undo/redo history, so rewinding could
+      // resurrect assignments in zones that no longer exist. Clear history at
+      // this boundary — the dialog already tells the user this can't be undone.
+      temporalManager.clear(MAP_MODES.DISTRICTS);
       if (selectedZone > newNumDistricts) {
         setSelectedZone(1);
       }
     }
   };
 
-  const requestNumDistricts = (requested: number) => {
+  /** Clamps and applies a requested count. Returns the committed value, or
+   * null when nothing changed yet (no-op, or deferred to the confirm dialog). */
+  const requestNumDistricts = (requested: number): number | null => {
     const clamped = Math.max(MIN_NUM_DISTRICTS, Math.min(MAX_NUM_DISTRICTS, requested));
-    if (clamped === numDistricts) return;
+    if (clamped === numDistricts) return null;
     if (clamped < numDistricts) {
       const {zoneAssignments} = useAssignmentsStore.getState();
       let hasAssignmentsAbove = false;
@@ -71,10 +79,11 @@ export const DistrictsZonePicker: React.FC = () => {
       });
       if (hasAssignmentsAbove) {
         setPendingDecrease(clamped);
-        return;
+        return null;
       }
     }
     commitNumDistricts(clamped);
+    return clamped;
   };
 
   const handleCommitDraft = () => {
@@ -83,10 +92,10 @@ export const DistrictsZonePicker: React.FC = () => {
       setDraftCount(String(numDistricts));
       return;
     }
-    requestNumDistricts(parsed);
-    // If the request was clamped or needs confirmation, snap the draft back to
-    // the current value; a confirmed change re-syncs via the effect above.
-    setDraftCount(String(numDistricts));
+    const committed = requestNumDistricts(parsed);
+    // Sync the draft to the value just committed; on a no-op or a change
+    // pending confirmation, snap back to the still-current count.
+    setDraftCount(String(committed ?? numDistricts));
   };
 
   const isReadOnly = access === ACCESS_STATES.READ;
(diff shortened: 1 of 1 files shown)


app/src/app/components/Toolbar/DistrictZonePicker.tsx
```diff
@@ -27,26 +35,58 @@ export const DistrictsZonePicker: React.FC = () => {
...
+  const handleCommitDraft = () => {
+    const parsed = parseInt(draftCount, 10);
+    if (isNaN(parsed)) {
+      setDraftCount(String(numDistricts));
+      return;
+    }
+    requestNumDistricts(parsed);
+    // If the request was clamped or needs confirmation, snap the draft back to
+    // the current value; a confirmed change re-syncs via the effect above.
+    setDraftCount(String(numDistricts));
```
fangge518: I: Suggestion: have `requestNumDistricts` return the clamped value when it commits immediately (or `null` when it's a no-op or deferred to the confirm dialog), and in `handleCommitDraft` set `draftCount` from that return value instead of unconditionally resetting to `numDistricts`. Reason: `numDistricts` in `handleCommitDraft` is this render's closure value, which hasn't caught up yet when `requestNumDistricts` just committed a new one synchronously — so the field is briefly reset to the *old* count instead of the one just committed, correct only because a later effect papers over it. The no-op and pending-dialog cases are already correct as written (nothing changed yet, so resetting to current `numDistricts` is right) — only the immediate-commit case needs the fix.

app/src/app/components/Toolbar/DistrictZonePicker.tsx
```diff
@@ -27,26 +35,58 @@ export const DistrictsZonePicker: React.FC = () => {
...
+  }, [numDistricts]);
+
   const handleRadioChange = (index: number, _color: string) => {
     const value = index + 1;
     setSelectedZone(value);
   };
 
-  const handleIncreaseDistricts = () => {
-    const newNumDistricts = numDistricts + 1;
+  const commitNumDistricts = (newNumDistricts: number) => {
```
fangge518, line 55: H: Reported repro: reduce from 6 districts to 5, then undo/redo — a phantom district 6 (grey-colored) reappears on what's nominally a 5-district map. Root cause: `commitNumDistricts` calls `setNumDistricts` (in `mapStore`, wrapped with `createWithDevWrapperAndSubscribe` — no undo/redo history at all) and `removeAssignmentsForZonesAbove` (in `assignmentsStore`, which **is** wrapped with zundo undo/redo) as two separate, uncoordinated state changes. Undoing/redoing only ever rewinds `assignmentsStore`'s tracked history (`zoneAssignments`, `shatterIds`, etc. — `num_districts` isn't part of it, confirmed via `middlewareConfig.ts`'s `partialize`), so a temporal navigation can restore old zone assignments referencing a zone number that `num_districts` no longer includes.

This looks pre-existing (the old `handleDecreaseDistricts` made the same two uncoordinated calls) rather than introduced by this PR — but this PR makes it much easier to trigger: direct multi-digit entry (e.g. 6→2) removes assignments across several zones at once where previously only one zone at a time was removable via the buttons, so the desync is far more likely to be hit in practice now. Flagging as blocking for this PR given that increased exposure, not because this diff caused the root cause.

app/src/app/components/Toolbar/DistrictZonePicker.tsx
```diff
@@ -68,25 +108,23 @@ export const DistrictsZonePicker: React.FC = () => {
...
-                    MIN_NUM_DISTRICTS,
-                    Math.min(MAX_NUM_DISTRICTS, Number(e.target.value))
-                  );
-                  if (!isNaN(val)) {
-                    setNumDistricts(val);
+                onChange={e => setDraftCount(e.target.value)}
+                onBlur={handleCommitDraft}
+                onKeyDown={e => {
+                  if (e.key === 'Enter') {
+                    e.currentTarget.blur();
```
fangge518, line 142: H: Reported repro: typing e.g. "4" into a 6-district map (with assignments above 4) via **Enter** sometimes shows no confirm dialog at all — the field just reverts to "6" as if nothing happened. Confirmed does not reproduce via click-away (blurring the field by clicking elsewhere works fine every time) — only Enter.

Likely mechanism: `onKeyDown`'s Enter handler calls `e.currentTarget.blur()`, which runs `handleCommitDraft` → `requestNumDistricts` → opens the `AlertDialog` (`setPendingDecrease`). Radix's `AlertDialogContent` (confirmed in the installed `@radix-ui/react-alert-dialog` source, `onOpenAutoFocus`) explicitly moves focus to the **Cancel** button as soon as the dialog opens. If the same physical Enter press repeats even slightly (OS key-repeat while held, or a fast habitual double-press right after typing a number) — the second Enter now lands on the focused Cancel button and activates it, closing the dialog via Cancel, which never commits the change. That would produce exactly this symptom, and explains why only Enter triggers it: a mouse click has no equivalent "second keypress" that could land on newly-focused dialog content.

The focus-to-Cancel behavior is confirmed from the library source; the "second Enter lands on Cancel" trigger is inferred from that plus the reported Enter-vs-click distinction, not reproduced live — worth confirming with a keyboard event log or by testing whether a single, deliberately brief Enter tap (vs. a slightly held one) changes the failure rate.

fangge518 (changes requested):
Traced the new draft/commit/confirm flow end to end. The clamp-and-confirm logic and dialog wiring are correct in isolation, and the multi-digit-entry / missing-cleanup bugs this PR fixes are real (confirmed against the pre-PR code). Three inline comments: a minor fragile-but-currently-invisible render glitch (I), plus two reported-reproducible issues (H) — an undo/redo desync that leaves a phantom colored district, and the confirm dialog sometimes not appearing on a genuine decrease.

app/src/app/components/Toolbar/DistrictZonePicker.tsx
nofurtherinformation: Good catch -- resolved in 48423b1c5a8248a6770c58764db7e9f4b3917b22 by clearing undo/redo history. This is a niche feature to begin with (supporting mid-plan district changing) so this should not be that unpleasant of an edge case

app/src/app/components/Toolbar/DistrictZonePicker.tsx
nofurtherinformation: Fixed keyDown -> keyUp in 48423b1c5a8248a6770c58764db7e9f4b3917b22 and blur in 4f02a990099021a3d0b2bc51aa7956ff89f77d7d

app/src/app/components/Toolbar/DistrictZonePicker.tsx
nofurtherinformation: Good call, implemented in 48423b1c5a8248a6770c58764db7e9f4b3917b22

fangge518 (approved):
LGTM.

PR #636, opened 2026-07-21
Only display public IDs to user

nofurtherinformation:
Rework routing so only the public ID displays the users.


fangge518:
I fixed five issues with the recent commits.

1. **Blank map / false Draw availability** — visiting an edit URL by public_id with no (or a stripped) `private_edit_id` didn't degrade to read-only the way it should have: the page still mounted the editor and hydrated via the private fetch path, producing a blank map, and the mode switcher showed Draw/Superdraw as available even without real edit access (`useEditableDocId` trusted a UUID-shaped `document_id` as proof of access, which the private fetch path happened to satisfy regardless). `MapPage.tsx`/`CoiMapPage.tsx` now redirect to the plain display route whenever there's no real edit capability, instead of trying to render the editor inline against a document we can't edit.
2. **`SaveConflictModal.tsx`** — `onNavigate` wasn't updated to this PR's new signature (the whole `DocumentObject`, not a plain id string), so resolving a save conflict by forking would have pushed a broken `/map/edit/[object%20Object]` URL. Fixed to use `editPath()`, matching every other call site.
3. **`MapSelector.tsx`** — a leftover raw `/edit/{document_id}` URL (the "attach one of my maps" comment-form picker) migrated to `editPath()`.
4. **Naming nit** — renamed `coi/edit/[document_id]` to `coi/edit/[map_id]`, matching `map/edit/[map_id]`'s naming now that the segment can hold either id type.
5. **Production build failure** — `useSearchParams()` (added by this PR) needs a Suspense boundary on pages eligible for static prerendering; `/map/edit` and `/coi/edit` (the bare landing pages) qualify, and the deploy preview's build failed prerendering `/coi/edit` as a result. Wrapped both pages' content in `<Suspense>`.

nofurtherinformation:
LGTM! 

A couple small nits - none blocking:

E2E fixtures should be updated with the new paths:
- `map-fixture.ts:171` — waitForURL(/\/map\/edit\//)
- `map-creation.spec.ts:51` — /\/map\/edit\/[a-zA-Z0-9-]+/
- `share-map.spec.ts:76` — same pattern
- `topbar-navigation.spec.ts:102` — /\/map\/eval\/\d+/

`CoiMap.tsx:30` has a stale comment

PR #638, opened 2026-07-21
Refine zoom to unassigned and discontiguous 

nofurtherinformation:
When a component contains exactly one geometry, snap (no animation) to its approximate area, wait for tiles to load, query the map for the geometry's rendered pieces, and fit to their true bbox. Centroid-derived unassigned bboxes collapse to a point for single units, so the old flow could only zoom to the general area (capped at z10). Multi-geometry components keep the existing fitBounds behavior.

The contiguity bbox endpoint now wraps single-node components in a GeoJSON Feature carrying {bbox, geo_ids} properties so the client can do the same.


app/src/app/components/sidebar/MapValidation/ZoomToFeature.tsx
```diff
@@ -60,43 +65,110 @@ export default function ZoomToFeature({
...
+        ],
+        {
+          maxZoom: 16,
+          speed: 2.4, // default 1.2; zippier since we already snapped nearby
+          ...(padding ? {padding: getFitBoundsPadding(mapRef, padding)} : {}),
+        }
+      );
+    };
+    pendingIdleHandler.current = onIdle;
+    mapRef.once('idle', onIdle);
```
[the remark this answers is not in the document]
nofurtherinformation: Resolved in 3eb4062d91731eaed90d9713adaef7199f30b498

app/src/app/store/unassignedFeatures.ts
```diff
@@ -41,11 +41,44 @@ export const useUnassignFeaturesStore = create<UnassignedFeatureStore>((set, get
...
+    // overlap. `reset()` (new document) starts the list over.
+    const prevFeatures = get().unassignedFeatureBboxes;
+    let merged = newFeatures;
+    if (prevFeatures.length) {
+      const idToNewIndex = new Map<string, number>();
+      newFeatures.forEach((f, i) =>
+        f.properties?.geo_ids?.forEach((id: string) => idToNewIndex.set(id, i))
+      );
+      const matched = new Set<number>();
+      merged = prevFeatures.map(f => {
```
fangge518: H: The merge can silently mark a still-unassigned area `resolved: true` — when two separate unassigned regions dissolve into one, only the first old row claims the shared new index; the second falls into the `newIndex === undefined` branch and gets marked done anyway. Resolved rows also short-circuit permanently (`if (f.properties?.resolved) return f;`), so this doesn't self-heal, and it has a second symptom: since a resolved row is never re-checked against new data, undoing a fix doesn't un-resolve it — the reverted area just reappears as a new row further down the list instead.

An alternative solution to the problem of number mismatch after fixes: the root problem is that unassigned areas have no stable identity — they merge, split, and can be undone — so any per-row `resolved` state is tracking something that doesn't durably exist. I think we should drop the attempt to number the areas and marking them as resolved, and instead show the live, unnumbered set of currently-unassigned regions on each refresh — a number or checkmark implies a stable area-to-slot mapping the data can't actually promise.

app/src/app/components/sidebar/MapValidation/ZoomToFeature.tsx
```diff
@@ -45,58 +63,156 @@ export default function ZoomToFeature({
...
+  });
+
+  // After the snap, wait for the map to go idle (tiles loaded), then fly in.
+  // With geoIds, the map is queried for the geometries' rendered pieces and the
+  // fly targets their union bbox — needed because unassigned-area bboxes are
+  // centroid-derived: they understate the true extent, and a single-unit
+  // component collapses to a point. Without geoIds the fly targets approxBounds
+  // directly; waiting for idle anyway keeps the pacing identical to the
+  // query path, so every zoom reads as snap → beat → fly.
+  const flyInAfterIdle = (approxBounds: LngLatBoundsLike, geoIds?: string[]) => {
```
fangge518, line 115: PP: The snap→beat→fly pacing's "beat" is whatever time it takes `mapRef.once('idle', ...)` to fire — there's no engineered minimum hold at the wide context view. On a fast/cached tile load, idle can fire almost immediately, so the pause can be too brief to actually orient in before the fly-in to the tight final bounds starts. Worth a minimum dwell (e.g. the later of "tiles idle" and a fixed ~300-500ms timer) so the pause is consistent regardless of cache state.

fangge518 (changes requested):
Traced the zoom/query call chain end to end: the new snap→idle→fly sequence in `ZoomToFeature.tsx`, the backend's `geo_ids` addition in `main.py`/`contiguity/main.py` (confirmed graph node ids are the same `geo_id`/`path` strings the frontend filters vector tiles on), and the new checklist-merge logic in `unassignedFeatures.ts`. One holding issue regarding the new labeling model for unassigned and district fragments — it can permanently mislabel a still-unassigned area as resolved.

app/src/app/store/unassignedFeatures.ts
nofurtherinformation: Good catch and good call!

nofurtherinformation: Simplified in 6a2b9b807851ad71d1a8807c24e79edd6a893420

app/src/app/components/sidebar/MapValidation/ZoomToFeature.tsx
nofurtherinformation: 400ms min dwell time added in 6a2b9b807851ad71d1a8807c24e79edd6a893420

fangge518 (approved):
LGTM

PR #639, opened 2026-07-21
Fix Alaska map bbox and prevent recurrence on reload

fangge518:
Alaska maps were opening zoomed all the way out to the whole world instead of framing Alaska. `DistrictrMap.extent` (the bbox fed straight into MapLibre's `fitBounds`) was `[-179.23, 51.18, 179.86, 71.44]` — a ~359°-wide box.

Root cause: `add_extent_to_districtrmap` (`backend/app/utils.py`) computes the bbox as a naive `ST_XMin`/`ST_XMax` over `ST_Extent(geometry)`. That's fine for any ordinary state, but Alaska's Aleutian islands have vertices on both sides of the antimeridian (mainland ~-179° to -130°, westernmost islands ~+172° to +180°) — min/max across mixed signs produces a near-global span instead of the true ~50° one. This function reruns on every map (re)creation, so the bug would resurface the next time Alaska's map data gets reloaded.

Already hand-fixed the 6 affected rows on dev and prod (`districtrmap.extent` set to a standard non-wrapping Alaska bbox). This PR is the code fix so it doesn't come back.

**Fix:** when the naive extent comes out wider than 180° (impossible for any single non-antimeridian-crossing state), recompute with `ST_ShiftLongitude`, which maps negative longitudes into the 0–360° range so the data becomes contiguous. The resulting bbox can have coordinates outside ±180°, which MapLibre's `fitBounds` renders correctly for antimeridian-crossing regions.

Added a regression fixture (`antimeridian_geos.geojson`, two small polygons straddling 180°) and a test asserting the computed bbox stays narrow instead of ballooning to ~360°.

### DIFF
--- a/backend/app/utils.py
+++ b/backend/app/utils.py
@@ -481,6 +481,36 @@ def add_extent_to_districtrmap(
                 rec.parent_layer
             ) INTO layer_extent;
 
+            -- A >180-degree-wide bbox is impossible for a legitimate single-state
+            -- layer. It means the geometry straddles the antimeridian (e.g.
+            -- Alaska's Aleutian islands, recorded with longitudes on both the
+            -- -179.x and +179.x side), and the naive ST_XMin/ST_XMax above
+            -- produced a near-global span instead of the true, much narrower,
+            -- extent. Recompute using ST_ShiftLongitude, which maps negative
+            -- longitudes into the 0..360 range so the data becomes contiguous;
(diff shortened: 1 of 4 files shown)


nofurtherinformation (approved):
LGTM! Sometime we should clean up some of the lingering unhinged SQL statements if possible and convert to in-memory / sqlalchemy

PR #640, opened 2026-07-21
Fix slight color mismatch between VTDs and broken blocks

nofurtherinformation:
Broken parents keep rendering their gray unassigned background (zone is null once shattered), which tinted the child blocks painted on top of them. Hide the background for broken features, and unify the child unassigned opacity with the parent's so unassigned areas match too.


fangge518 (approved):
LGTM.

PR #641, opened 2026-07-21
Fix coalition builder breaking the population chart in view mode

fangge518:
In view mode (a public/read-only plan view), toggling any demographic coalition checkbox made the population chart go blank; unchecking it didn't restore the original numbers either. Draw/edit mode was unaffected.

View mode fed demography through the same block-level assignment/join pipeline edit mode uses, via a fabricated identity-assignment map built in `PublicSource.tsx`. Every later population recompute triggered by the coalition builder (a checkbox toggle) called `updatePopulations()` with no `zoneAssignments` arg, which fell back to the live assignment stores — empty in view mode, since view mode never populates them for performance. The join against an empty zone table zeroed out every district's population, and `PopulationChart` skips rendering any bar with `total_pop_20 <= 0`.

**Fix:** the backend already returns fully-aggregated per-district demography (`document.district_unions` sums every numeric gerrydb column per zone, including race/ethnicity columns). Give view mode its own population-assembly path in `DemographyService` that uses that data directly — no block/zone join, no assignments concept, no live-store dependency — so a coalition toggle in view mode is a pure recompute over already-fetched data.

## Tested

- [x] `bun run ts` and `bun run build` clean.
- [x] Local seed data, public view `/map/311` (CA custom districts, 72k+ real assignments): population chart renders on load, coalition checkbox toggle updates bars/numbers per-district, unchecking restores totals.
- [x] Draw mode: coalition toggle still updates the population chart correctly (its code path, `calculatePopulations`'s non-public branch, is unchanged by this diff — not independently re-run).

### DIFF
--- a/app/src/app/components/Map/GeoSources/PublicSource.tsx
+++ b/app/src/app/components/Map/GeoSources/PublicSource.tsx
@@ -80,10 +80,17 @@ export const PublicSource: React.FC<{children: React.ReactNode}> = ({children}) 
       setStateFp(publicDistrictsQuery.data.statefp);
     }
 
-    // Feed demographic data into demographyService for sidebar stats
-    const {columns, demographicData, assignments} = publicDistrictsQuery.data;
+    // Feed demographic data into demographyService for sidebar stats. The backend
+    // already aggregates demography per district (document.district_unions), so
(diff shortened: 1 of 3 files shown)


nofurtherinformation (approved):
LGTM!

PR #643, opened 2026-07-21
Feat/misc cleanup

nofurtherinformation:
* County brush is too hard to find
* Text - remove you can draw in limited states
* Draw on top nav, Main Map in hamburger
* District plan buttons more prominent on place pages
* Clarify what is ideal population
* When looking at unassigned / contiguity, the red text notifying about a save needed might not be big enough

### DIFF
--- a/app/src/app/components/Static/Interactions/PlaceMapGrid.tsx
+++ b/app/src/app/components/Static/Interactions/PlaceMapGrid.tsx
@@ -1,39 +1,109 @@
 'use client';
-import {useState} from 'react';
-import {SegmentedControl, Flex, Grid} from '@radix-ui/themes';
-import {CreateButton} from './CreateButton';
+import {Card, Flex, Grid, Heading, Spinner, Text} from '@radix-ui/themes';
+import {ArrowRightIcon, Component1Icon, LayersIcon, PersonIcon} from '@radix-ui/react-icons';
 import {DistrictrMap} from '@/app/utils/api/apiHandlers/types';
-import {MAP_TABS, type MapTab} from '@constants/document/tabs';
 import {sanitizeCommunityMaps} from '@/app/utils/communities';
+import {useCreateMapDocument} from './CreateButton';
+import {ImportBlockAssignments} from './ImportBlockAssignments';
 
-export const PlaceMapGrid: React.FC<{maps: Partial<DistrictrMap>[]}> = ({maps}) => {
-  const [activeTab, setActiveTab] = useState<MapTab>(MAP_TABS.DISTRICTS);
-  const filteredMaps = activeTab === MAP_TABS.DISTRICTS ? maps : sanitizeCommunityMaps(maps);
+const MapStartCard: React.FC<{
+  view: Partial<DistrictrMap>;
+  isCommunity: boolean;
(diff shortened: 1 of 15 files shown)


app/src/app/components/Static/Interactions/PlaceMapGrid.tsx
```diff
@@ -1,39 +1,109 @@
...
+        </Flex>
+        <CardGrid>
+          {maps.map((view, i) => (
+            <MapStartCard key={i} view={view} isCommunity={false} />
+          ))}
+        </CardGrid>
+      </section>
+      <section>
+        <Heading as="h3" size="3" mb="1">
+          Community maps
```
fangge518: I: The new layout renders the "Community maps" heading, description text, and `CardGrid` unconditionally over `communityMaps = sanitizeCommunityMaps(maps)`. `sanitizeCommunityMaps` filters out any map missing a `name` and can return an empty array for places where no map qualifies as a community map. In that case the page shows a "Community maps" section header and description with no cards underneath — a plausible dead-end for a visitor on some place pages. Worth gating the section on `communityMaps.length` the same way the outer `PlaceMapGrid` render is already gated on `availableMaps.length` in the parent page.

app/src/app/components/Static/Interactions/PlaceMapGrid.tsx
```diff
@@ -1,39 +1,109 @@
...
+  return (
+    <Flex direction="column" gap="5">
+      <section>
+        <Flex justify="between" align="start" gap="3" wrap="wrap" mb="2">
+          <Flex direction="column">
+            <Heading as="h3" size="3" mb="1">
+              District plans
+            </Heading>
+            <Text as="p" size="2" color="gray">
+              Split a map into districts, one for each seat.
```
fangge518: I: New copy: "Split a map into districts, one for each seat." This asserts a 1:1 district-to-seat mapping that doesn't hold for multi-member districts. Suggest dropping the seats claim or hedging it.

fangge518 (approved):
Traced each of the unrelated changes against its actual call sites rather than reviewing the diff in isolation: the `DEFAULT_COLOR_SCHEME_GRAY` rewrite against its sole consumer in `demographyService.ts` (index bounds match exactly), the `CreateButton`/`PlaceMapGrid` hook extraction against both its usage sites for double-submit safety, and the `ImportBlockAssignments` relocation against the component's own state assumptions. All check out clean. Two non-blocking items below: an unconditional "Community maps" section that can render with zero cards, and new text that needs some edits.

app/src/app/components/Static/Interactions/PlaceMapGrid.tsx
nofurtherinformation: Good catch - dead end handled in ebf08ff3ac235d36d5172889a8aaef23d47abc1d

app/src/app/components/Static/Interactions/PlaceMapGrid.tsx
nofurtherinformation: Improved in ebf08ff3ac235d36d5172889a8aaef23d47abc1d

PR #644, opened 2026-07-21
Add a Total column to the demography table; fix nested scrollbar and dark-cell text contrast

fangge518:
Four related gaps in the demography sidebar's Table view.

### Missing Total column

The table showed each district's race/ethnicity breakdown but never the row's own denominator. Total population has a home in the District overview panel, but total voting-age population had no UI surface anywhere once on the VAP tab.

**Fix:** add a Total column (raw `total_pop_20`/`total_vap_20`) to the TOTPOP/VAP table configs — never color-shaded, unaffected by the share/count toggle.

### Scrollbar covering table data

The nested table's scrollbar overlapped the rightmost column's numbers.

**Fix:** reserve padding inside `ConditionalScrollArea`'s scrolled content so the scrollbar has empty space to sit in instead of overlapping the data.

### Scrollbars sitting too close together

The nested table's scrollbar sat close enough to the sidebar's own outer scrollbar that reaching the next panel meant scrolling the inner list all the way down first.

**Fix:** shrink `ConditionalScrollArea`'s `ScrollArea` box width, so its scrollbar — which Radix renders at that box's edge, a sibling of the scrolled content, not inside it — sits a real gap away from the sidebar's outer scrollbar (padding inside the content, the fix above, never moves the scrollbar itself; these are two different reservations). Separately, make the sidebar's own outer scrollbar (`Sidebar.tsx`) sit flush against the true window edge instead of a padding-width short of it — the easiest possible target to grab, and it further separates it from the inner scrollbars for free. Took three passes to actually land: padding inside scrolled content doesn't touch the scrollbar itself (sibling, not descendant); a negative margin to bleed the ScrollArea past its container's padding got closer but left a remainder, because Radix's own built-in scrollbar margin sits on top of any box-level offset; landed on not applying right padding to the container at all, restoring it only where content actually needs the inset, plus zeroing Radix's own scrollbar-margin variable directly.

### Unreadable text on dark cells

Cells shaded dark by "color cells by values" kept default dark text — hard to read against a dark background. The same bug independently existed in `PartisanSection.tsx` on the separate Map Evaluation page.

**Fix:** a shared `getReadableTextColor` helper, used in both the demography table and `PartisanSection`, so dark cells get white text.

Along the way, renamed `Evaluation.tsx` → `DemographyTable.tsx` (and its `EVAL_MODES`/`EvalColumnConfiguration`/`evaluationConfig.ts` neighbors) — it was colliding in name with the unrelated Map Evaluation page.

## Tested

- [x] `bun run ts`, `bun run build`, and `docker-compose up pre-commit` clean.
- [x] Total column renders correct raw counts on both TOTPOP and VAP tabs, matches the District overview panel's numbers, and the "Overall" row shows the correct statewide sum.
- [x] Scrollbar no longer overlaps the rightmost column's data.
- [x] Sidebar's outer scrollbar sits flush against the true window edge (mouse reaches it at the screen's rightmost pixel, maximized window) and the gap to the inner per-panel scrollbar reads as comfortable (visual check — third pass, see the Fix note above for why the first two didn't fully close this).
- [x] Dark cells get white text and light cells keep default dark text, in both the demography table's scales and `PartisanSection` on the Map Evaluation page.

## Review required

- Double check the tested items above with other maps.

### DIFF
--- a/app/src/app/components/sidebar/DemographyTable/DemographyTable.tsx
+++ b/app/src/app/components/sidebar/DemographyTable/DemographyTable.tsx
@@ -526,7 +526,12 @@ const DemographyTableCell: React.FC<DemographyTableCellProps> = ({
         color: textColor,
       }}
     >
-      {numericValue === undefined
+      {/* An empty/unpainted district has no data for the Total column's underlying
+          zone, which arrives as a real 0 (needed elsewhere for chart scaling and
+          min/max stats) rather than undefined — display it the same as "no data"
+          (`--`) so it doesn't read as a legitimate zero, and so TOTPOP and VAP
+          render consistently for the same empty district. */}
+      {numericValue === undefined || (isTotalColumn && numericValue === 0)
         ? '--'
         : formatNumber(numericValue, isTotalColumn ? NUMBER_FORMATS.STRING : numberFormat)}
     </Table.Cell>
--- a/app/src/app/components/sidebar/DemographyTable/DemographyTable.tsx
+++ b/app/src/app/components/sidebar/DemographyTable/DemographyTable.tsx
@@ -23,7 +23,7 @@ import {
   modeButtonConfig,
   numberFormats,
   summaryStatLabels,
-} from '@/app/store/demography/evaluationConfig';
+} from '@/app/store/demography/demographyTableConfig';
 import {GearIcon, InfoCircledIcon} from '@radix-ui/react-icons';
 import {useColorScheme} from '@/app/hooks/useColorScheme';
 import {demographyService} from '@/app/utils/demography/demographyService';
@@ -42,20 +42,26 @@ import {
   isCoalitionUniverse,
   TOTAL_COLUMN,
 } from '@constants/demography/summary';
-import {type NumberFormat} from '@constants/demography/format';
-import {EVAL_MODES, type EvalMode} from '@constants/demography/evalMode';
+import {NUMBER_FORMATS, type NumberFormat} from '@constants/demography/format';
+import {
+  TABLE_DISPLAY_MODES,
+  type TableDisplayMode,
+} from '@constants/demography/demographyTableMode';
 import {MAP_MODES} from '@constants/map/mode';
 import {PovSwitcher, type Pov} from '@components/Shared/PovSwitcher';
 import {ConditionalScrollArea} from '../ConditionalScrollArea';
+import {getReadableTextColor} from '@/app/utils/colors';
 
 type ColumnConfig = {
   label: string;
   column: string;
   sourceCol?: string;
   tooltip?: string;
+  /** Denominator column: always a raw count, never shaded, never a share of itself. */
+  isTotal?: boolean;
 };
 
-type EvaluationProps = {
+type DemographyTableProps = {
   summaryType: SummaryType;
   setSummaryType: (summaryType: SummaryType) => void;
   displayedColumnSets?: Array<SummaryType>;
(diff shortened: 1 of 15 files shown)


app/src/app/components/sidebar/DemographyTable/DemographyTable.tsx
```diff
@@ -478,28 +484,36 @@ const EvaluationTableCell: React.FC<EvaluationTableCellProps> = ({
   maxValues,
   pov,
 }) => {
-  const column = evalMode === EVAL_MODES.COUNT ? columnConfig.column : `${columnConfig.column}_pct`;
+  const isTotalColumn = Boolean(columnConfig.isTotal);
```
nofurtherinformation, line 487: H: Total population shows zero for empty districts but Voting Age Population show `--`. I think `--` is clearer

app/src/app/store/demography/demographyTableConfig.ts
```diff
@@ -32,18 +40,20 @@ export const TOTPOPColumnConfig: EvalColumnConfiguration<
...
-    {column: 'hvap_20', label: 'Hispanic'},
-    {column: 'amin_vap_20', label: 'AMIN'},
-    {column: 'asian_nhpi_vap_20', label: 'Asian'},
-    {column: 'white_vap_20', label: 'White'},
-    {column: 'other_vap_20', label: 'Other'},
-  ];
+export const VAPColumnConfig: DemographyTableColumnConfiguration<
+  SummaryStatConfig[typeof SUMMARY_TYPES.VAP]
+> = [
+  {label: 'Total', column: 'total_vap_20', isTotal: true},
```
nofurtherinformation: I: I think total should go at the end

app/src/app/store/demography/demographyTableConfig.ts
```diff
@@ -59,56 +69,56 @@ export const VoterColumnConfig: EvalColumnConfiguration<
...
 ];
 
-export const numberFormats: Record<EvalMode, NumberFormat> = {
-  [EVAL_MODES.SHARE]: NUMBER_FORMATS.PERCENT,
-  [EVAL_MODES.COUNT]: NUMBER_FORMATS.STRING,
-  [EVAL_MODES.TOTPOP]: NUMBER_FORMATS.PERCENT,
+export const numberFormats: Record<TableDisplayMode, NumberFormat> = {
+  [TABLE_DISPLAY_MODES.SHARE]: NUMBER_FORMATS.PERCENT,
+  [TABLE_DISPLAY_MODES.COUNT]: NUMBER_FORMATS.STRING,
+  [TABLE_DISPLAY_MODES.TOTPOP]: NUMBER_FORMATS.PERCENT,
```
nofurtherinformation: PP: Is number format dead code?

nofurtherinformation (changes requested):
Code looks good! A couple nits but otherwise good to go!

app/src/app/components/sidebar/DemographyTable/DemographyTable.tsx
fangge518: Good catch — fixed, both now show `--` for empty districts.

app/src/app/store/demography/demographyTableConfig.ts
fangge518: Moved it to the end.

app/src/app/store/demography/demographyTableConfig.ts
fangge518: Yep, dead — nothing ever sets evalMode to TOTPOP. Removed it.

nofurtherinformation (approved):
LGTM!

PR #645, opened 2026-07-21
Legacy flow

nofurtherinformation:
Handles legacy routes

### DIFF
--- a/app/src/app/not-found.tsx
+++ b/app/src/app/not-found.tsx
@@ -1,0 +1,28 @@
+'use client';
+import {usePathname} from 'next/navigation';
+import {Box, Flex, Heading, Link, Text} from '@radix-ui/themes';
+import {Header} from '@components/Static/Header';
+import {Footer} from '@components/Static/Footer';
+
+export default function NotFound() {
+  const pathname = usePathname();
+  const legacyUrl = `https://districtr.org${pathname ?? ''}`;
+
+  return (
+    <Flex direction="column" className="min-h-[100vh]" justify="center">
+      <Header />
+      <Box className="w-full flex-grow p-4 pt-0 max-w-screen-lg mx-auto px-4 xl:px-0">
+        <Flex direction="column" gapY="4" align="center" py="9">
+          <Heading as="h1">Page not found</Heading>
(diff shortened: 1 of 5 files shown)


app/src/app/not-found.tsx
```diff
@@ -0,0 +1,29 @@
+'use client';
+import {usePathname} from 'next/navigation';
+import {Box, Flex, Heading, Link, Text} from '@radix-ui/themes';
+import {Header} from '@components/Static/Header';
+import {Footer} from '@components/Static/Footer';
+import {LEGACY_DISTRICTR_URL} from '@/app/constants/legacy';
+
+export default function NotFound() {
+  const pathname = usePathname();
+  const legacyUrl = `${LEGACY_DISTRICTR_URL}${pathname ?? ''}`;
```
fangge518, line 11: PP: The global 404 page suggests the equivalent `districtr.org` URL for every not-found path, including ones that never existed on the old site (e.g. a typo'd `/map/edit/...` URL) — in that case the suggested legacy link will itself 404. Recommend scoping the "try the classic version" messaging to legacy-shaped paths (`/plan/`, `/group/`, etc.) rather than showing it unconditionally for every 404.

fangge518 (approved):
Checked for route collisions with existing v2 pages (none), confirmed the new global `not-found.tsx` correctly hand-renders `Header`/`Footer` since it sits outside the `(static)` route group's layout, and confirmed there's no open-redirect risk (legacy URLs are always built by appending the current path/id to a fixed, hardcoded origin). One small PP recommendation below — nothing blocking.

app/src/app/not-found.tsx
nofurtherinformation: This is worth considering - the trick here is that a lot of old districtr routes don't have a strong identity -- eg. `https://districtr.org/chicago/current-wards` is a plan of the chicago wards, but uses the place name as the first level route. In that context, what do you think?

nofurtherinformation: I made this a bit more clever, a lightweight API route probes the URL to see if it exists on the legacy site, and if so, suggests the legacy page

PR #646, opened 2026-07-21
Fix map removal in Catalog

nofurtherinformation:
The delete confirmation dialog never opened because AlertDialog.Trigger slotted its onClick onto the Tooltip component, which doesn't forward it to the IconButton. Swap the nesting so the Trigger slots directly onto the button, matching the working CopyLinkDropdown pattern.

### DIFF
--- a/app/src/app/components/RecentMapsList.tsx
+++ b/app/src/app/components/RecentMapsList.tsx
@@ -464,8 +464,8 @@ const RecentMapCard: React.FC<{
           <CopyLinkDropdown editUrl={editUrl} publicUrl={publicUrl} />
           {!active && (
             <AlertDialog.Root>
-              <AlertDialog.Trigger>
-                <Tooltip content="Remove from list">
+              <Tooltip content="Remove from list">
+                <AlertDialog.Trigger>
                   <IconButton
                     variant="ghost"
                     color="gray"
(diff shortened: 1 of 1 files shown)


fangge518 (approved):
I've traced the Radix composition to confirm the trigger fix is structurally sound, and it matches the `CopyLinkDropdown` component a few lines above (`Tooltip > *.Trigger > IconButton`), which already works today. Manually tested it locally, and found one follow-up issue: confirming "Remove" navigates into the map that was just deleted — noted below, preferably fixed here but fine as a fast-follow if that's easier.

---

**I — `app/src/app/components/RecentMapsList.tsx:492`, the "Remove" button inside `AlertDialog.Content`**

Tested locally — clicking the trash icon now correctly opens the confirmation dialog (the bug this PR targets), but clicking "Remove" inside it navigates to the map's edit route instead of just removing it from the list. Root cause: `AlertDialog.Content` is portaled to `document.body` in the DOM, but React's synthetic events bubble along the *component* tree, not the physical DOM tree — so the click on "Remove" still bubbles up through `AlertDialog.Root` to the card's own `onClick={() => !active && onSelect(data)}`, which calls `router.push` into the document that was just deleted. `CopyLinkDropdown`'s `DropdownMenu.Content` already guards against exactly this with `onClick={e => e.stopPropagation()}`; `AlertDialog.Content` here doesn't have the equivalent. Suggest adding the same guard to `AlertDialog.Content` (or to the "Remove" button directly) — preferably as part of this PR since it's small and this fix is what makes the path reachable for the first time, but not a blocker if you'd rather land it as a quick follow-up.

PR #648, opened 2026-07-22
Hotfix: Set num_districts_modifiable: false on fixed v2 map modules

nofurtherinformation:
The fixed legislative modules (congressional/senate/house) omitted the flag, so the loader's DistrictrMapPublic default made them modifiable. Custom maps keep their explicit true.


fangge518 (approved):
LGTM.

PR #650, opened 2026-07-22
Draw mode UI/UX enhancements

nofurtherinformation:
WIP PR, based on feedback from user testing


fangge518:
## Notes:
I like the new design of tool buttons: feels informative and clean

## Bugs:
The eval view doesn't work on my end for now. Seem to have problems talking to the s3 server. Might be just a fly-preview thing and will resolve after we migrate to aws, but need some confirmation.
After switching back from eval to draw, the item "Understand demographic and voter histories" in "Improve you plan" is no longer finished. Also "Evaluate your plan further" isn't marked as finished. Thus there is no way to truly finish everything under "Improve your plan"

## General issues
When a district's population exceeds ideal population, the current marker for the fact that it exceeds and how much it exceeds is not sufficiently visible to me. It looks very similar to a finished bar. Would it help if we extend the bar and let it fade or use rectangular dots, if it grows to be too long?
The currently sticky part of the the side panel occupies 50% of my sidebar (I'm using relatively small screen with relatively large font), making the room for checking Data Panels very small (see screenshot). This will make validity checking a bit inconvenient. The district selector should still be sticky, but can we make brush size config and current painting indication a part of the scrollable area and not sticky?

## Nits:
The radio button in front of the TODO items makes me want to click them, but these are not clickable.
The left edge of the "Move" button is no longer aligned with the left edge the other components. (also the right edge of "Redo")
District: 8 following immediately "Painting district 1" might be confusing. Perhaps "Number of Districts: 8"?
The inspector button in the SuperDraw is bit larger than the other buttons. I like the previous design where all the buttons have similar width. If it helps, perhaps we can shorten it to "Inspect" to align with the verb form of all the other tools.

PR #651, opened 2026-07-23
Surface the guide page and add contextual help across the map editor

fangge518:
User-testing feedback: people don't know how to do things in Draw mode, and nobody finds the `/guide` page.

### Guide discoverability

Previously, the top-nav "Learn" link and its section both defaulted to About, not the guide, and once inside the editor there was no path back to `/guide` short of navigating away entirely.

**Fix:** "Learn" now lands on `/guide` by default, and the drawing page's hamburger menu gained a "Guide" item that opens it in a new tab.

### Contextual help across the editor

Adds `HelpTip`, a hover-triggered help card. Most cards contain a "Watch video" link opening a nearly-full-window modal with the guide clip and a link to the matching guide section. For consistence, this change has unifies every explanation-on-hover in the editor to use this help card: toolbar tools, district locks, panel headers, Visual settings, district color picker, population stats, map title, save status, access status, Mode switcher, Map actions, and Super Draw. The old dark `InfoTip` tooltips and ad-hoc explanatory `Tooltip`s are replaced and `InfoTip` removed.

Hover delays come in two tiers: near-instant for read-mostly triggers (stats, panel explainer icons, occasional settings/edit buttons), and a full second for controls users click through repeatedly (toolbar tools, locks, the dropdown triggers), so help never gets in the way of doing the actual work.

### Also in this branch

The sidebar accordion headers drop their description line (one row: icon, label, help icon, chevron); the boundaries panel's provenance modal gets a labeled "Data info" button so the bare (i) icon stays reserved for hover help; `LoopVideoPlayer` cycles multi-clip sequences in one player and shows loading/error states; the video modal is code-split so non-users never load it; and the guide's "Saving your Map" section is renamed "Edit map names and details" — saving is automatic, and the old name suggested otherwise.


app/src/app/components/HelpTip/HelpTip.tsx
```diff
@@ -0,0 +1,225 @@
...
+/** The video modal, split out of `HelpTip` so a caller that needs its own trigger
+ * for the video (rather than the in-card link) can still reuse it directly. */
+export const HelpTipVideoDialog: React.FC<{
+  tip: HelpTipKey;
+  open: boolean;
+  onOpenChange: (open: boolean) => void;
+}> = ({tip, open, onOpenChange}) => {
+  const entry: HelpTipEntry = helpTipContent[tip];
+  const videoFiles = entry.videoFiles ?? (entry.videoFile ? [entry.videoFile] : []);
+  const videoUrls = videoFiles.map(
```
nofurtherinformation: I: On parent state updates (eg. background save), this can trigger a re-render reseting the video, we should memoize for reference stability

app/src/app/components/HelpTip/HelpTip.tsx
```diff
@@ -0,0 +1,225 @@
...
+        (children as React.ReactElement<TriggerProps>).props.onClickCapture?.(event);
+      },
+    })
+  ) : (
+    // A plain span, not IconButton: ghost IconButton pads itself and cancels that
+    // with negative margins, which can leave it internally taller than the text
+    // line it sits in. A fixed 16px flex box matches size="1" Text's line-height
+    // exactly, so it can never make its row taller than the text beside it.
+    <span
+      role="button"
```
nofurtherinformation, line 194: H: For accessibility, we should also add a label like `aria-label={entry.title}`. Keyboard usability is also a potential concern here, but I think in general we don't have hopes of a real keyboard-only workflow

app/src/app/components/HelpTip/HelpTip.tsx
```diff
@@ -0,0 +1,225 @@
...
+
+// Code-split: most hovers never reach "Watch video", so the player (and the video
+// bytes themselves, loaded lazily inside it) shouldn't be in everyone's initial bundle.
+const LoopVideoPlayer = dynamic(
+  () => import('@/app/components/Static/LoopVideoPlayer').then(mod => mod.LoopVideoPlayer),
+  {ssr: false}
+);
+
+const COLLAPSED_WIDTH = 220;
+const VIDEO_MODAL_MAX_WIDTH = 720;
```
nofurtherinformation, line 16: I: We should get Peter to UX test this before merging to main to see how the scaling / max width works on big monitor + Linux (probably the worst case scenario)

nofurtherinformation (changes requested):
## Notes:

Overall looks good and works well!

## Bugs:

- Click the `?` icon hides the tooltip. Not a bug per se but an interaction I suspect will occur
- On the branch preview, the first '?' click and open per session consistently doesn't load the video. I think this is a state update issue -- try adding `onCanPlay={() => setIsLoading(false)}` instead of or in addition to `onLoadedData` on the `video` element in `LoopVideoPlayer.tsx:133`

## General issues

- Accessibility and small changes noted inline
- On button level tooltips, I think it is important that we start each text with the action. Eg for the mode selector, I think it should be like this:
```
Click to change the map mode.

Move between Draw (build your plan), View (a clean read-only display), and Evaluate (a dashboard of stats about it).
```
Or for the paint tool:
```
Click to use the paintbrush tool.

Use the paintbrush to click and drag on the map to add units to your district.
```
- Worth a copy edit review, will do that later today but wanted to get these notes out sooner

## Nits:
- None; all important feedback captured above/inline

app/src/app/components/HelpTip/HelpTip.tsx
fangge518: Fixed — memoized on `entry` instead of the derived array, so an unrelated ancestor re-render (autosave, etc.) doesn't reset the video mid-watch anymore. Also added progress bar for clips.

app/src/app/components/HelpTip/HelpTip.tsx
fangge518: Added `aria-label={entry.title}` here, and did the same pass across the other components this PR touches that had the identical icon-only-no-name problem (ToolButtons, EditStatus, Header's mobile menu trigger, MapTitleDisplay's close button).

app/src/app/components/HelpTip/HelpTip.tsx
fangge518: Filed #661 to track it in case we want to test other UI/UXs on big monitor and/or Linux.

fangge518:
Responding to the top-level review:

**Bugs**

> Click the `?` icon hides the tooltip.

Fixed — clicking a trigger no longer dismisses the card at all now; only the pointer actually leaving both the trigger and the card closes it.

> On the branch preview, the first '?' click and open per session consistently doesn't load the video... try adding `onCanPlay`.

Fixed exactly as suggested — added `onCanPlay` alongside `onLoadedData`.

**Copy convention (start with the action)**

Addressed, with slightly different phrasing: "Click to select the X tool. With this tool selected, ..." — applied to pan/paint/erase/break, plus "Click X above to..." for the panel/topbar entries that lost their icon.

Inline replies posted on the three discussion threads. 👍 on the promised copy-edit review, will watch for it.

app/src/app/components/HelpTip/helpTipContent.ts
```diff
@@ -0,0 +1,157 @@
...
+    videoFile: 'undo_redo.webm',
+    guideAnchor: 'drawing-the-districts',
+  },
+  brushSize: {
+    title: 'Brush size',
+    text: 'Drag this slider to change how many units the paintbrush or eraser affects at once.',
+    videoFile: 'brush_size.webm',
+    guideAnchor: 'drawing-the-districts',
+  },
+  countyBrush: {
```
nofurtherinformation, line 57: H: This tip isn't attached to the county brush, but could be helpful

app/src/app/components/Toolbar/DistrictZonePicker.tsx
```diff
@@ -173,7 +173,13 @@ export const DistrictsZonePicker: React.FC = () => {
             </>
           )}
         </Flex>
-        <ColorPicker onValueChange={handleRadioChange} defaultValue={0} value={selectedZone - 1} />
+        <Flex align="start" gap="1">
```
nofurtherinformation: Q: Is this change required?

app/src/app/components/sidebar/DataCards.tsx
```diff
@@ -204,31 +208,45 @@ const AccordionSection: React.FC<{
   onToggle: () => void;
 }> = ({section, open, onToggle}) => {
   const Icon = section.icon;
+  // The row itself is the hover trigger — no separate icon — so its own onClick
+  // (toggling the accordion) has to survive being cloned by HelpTip below.
+  const headerRow = (
+    <div
```
nofurtherinformation: Q: Why not a button?

nofurtherinformation (approved):
A few small follow up questions/suggestions, but otherwise LGTM! I'll follow up sometime this week with a fine toothed comb on text copy.

app/src/app/components/HelpTip/helpTipContent.ts
nofurtherinformation: P.s. A tooltip on the "Boundaries and areas" would be great too

app/src/app/components/sidebar/DataCards.tsx
fangge518: Good catch. This is now a useless change from earlier attempts. I've reverted to the button to make the code change more surgical.

app/src/app/components/Toolbar/DistrictZonePicker.tsx
fangge518: Not exactly. I've reverted this change.

app/src/app/components/HelpTip/helpTipContent.ts
fangge518: Tips for county brush and "Boundaries and Areas" have been added.

PR #652, opened 2026-07-24
Fix ideal_population on maps with a user-modified district count

fangge518:
`DocumentEvaluationContext.ideal_population` (`backend/app/evaluation/context.py`) divided total population by `DistrictrMap.num_districts` — the map's static default — regardless of what the user actually set. On a `num_districts_modifiable` map, the real district count lives on `Document.num_districts`, which the frontend lets users change per plan. Any document that overrode the count got `ideal_population` (and the `population_deviation` metric that depends on it) computed against the wrong denominator.

**Fix:** coalesce `Document.num_districts` over `DistrictrMap.num_districts`, matching the same pattern already used in `dependencies.py` and `main.py` for reading the effective district count elsewhere in the backend.

Added regression tests in `test_validity_metrics.py` covering both the `ideal_population` and `population_deviation` metrics under a document-level override, alongside the existing map-default-only case.

Also bumped the `ideal_population` and `population_deviation` metric versions in `registry.py`, since existing cached evaluation rows computed the old (map-default-only) value — this forces recomputation on next read instead of serving stale numbers.

### DIFF
--- a/backend/app/evaluation/context.py
+++ b/backend/app/evaluation/context.py
@@ -209,13 +209,29 @@ class DocumentEvaluationContext:
 
     @cached_property
     def ideal_population(self) -> int:
-        """Ideal population per district (total population ÷ map's number of districts)."""
-        num_districts = self._districtr_map.num_districts
+        """Ideal population per district (total population ÷ document's number of districts).
+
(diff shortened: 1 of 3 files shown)


nofurtherinformation (approved):
LGTM

PR #654, opened 2026-07-27
Button refinements

nofurtherinformation:
Update button styling


app/src/app/components/Toolbar/ToolControls/ToolControls.tsx
```diff
@@ -1,39 +1,29 @@
...
+  redo: {},
   brush: {
     Component: BrushControls,
   },
   eraser: {
     Component: BrushControls,
   },
+  // The break flow is guided by the on-map BlockModePill (which also hosts
+  // the exit control); the sidebar keeps the paint controls, since breaking
+  // leads straight into painting blocks.
```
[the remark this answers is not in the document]
nofurtherinformation: This is re-integrated in in #659

app/src/app/components/Toolbar/ToolButtons.tsx
```diff
@@ -1,101 +1,144 @@
...
-                    </span>
-                  ))}
-                  <Tooltip.Arrow className="fill-gray-900" />
-                </Tooltip.Content>
-              </Tooltip.Portal>
-            </Tooltip.Root>
-          </Tooltip.Provider>
-        );
-      })}
+      <Flex direction="row" wrap="nowrap" className="flex-grow" gap="1">
```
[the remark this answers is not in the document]
nofurtherinformation: Wrapped in f471dc83dbac17d5a23accbff6c61375dcf4a63a

app/src/app/components/Toolbar/ToolControls/ToolControls.tsx
```diff
@@ -1,39 +1,29 @@
...
   shatter: {
-    Component: () => {
-      const focusFeatures = useMapStore(state => state.focusFeatures);
-      if (focusFeatures.length) {
-        return <Text>Focused on {focusFeatures[0].id}</Text>;
-      } else {
-        return <Text>Click a feature to show the census blocks within it</Text>;
-      }
-    },
+    Component: BrushControls,
```
fangge518: I: This hunk (dropping `<ExitBlockViewButtons />`, pointing at an "on-map BlockModePill") is identical to one already in PR #659, which is the PR that actually adds `BlockModePill.tsx`. To enhance PR independence, perhaps remove this part from #654 and letting #659 carry it, so the exit-control handoff lands in one place with its supporting component rather than duplicated across two branches.

app/src/app/components/Toolbar/ToolButtons.tsx
```diff
@@ -1,101 +1,145 @@
...
+      <Flex direction="row" wrap="wrap" className="flex-grow" gap="1">
+        {/* flexBasis 0 (not auto) so every tool gets the same width regardless
+            of label length. Wraps because the sidebar resizes down to 140px,
+            below the five Super Draw tools' combined minimum. */}
+        {mainTools.map(tool =>
+          renderTool(tool, {minWidth: TOOLBAR_SIZE, flexGrow: 1, flexBasis: 0})
+        )}
+      </Flex>
+      <Flex direction="row" gapX="1">
+        {historyTools.map(tool => renderTool(tool, {width: HISTORY_BUTTON_WIDTH}))}
```
fangge518: I: Nit. The width balance between the larger buttons and Undo/Redo is perfect at the default sidebar width (or narrower), but at a wider sidebar (e.g. when checking wide tables), the main buttons look oversized next to a fixed-width Undo/Redo. Can Undo/Redo scale with the other buttons too, instead of a fixed `HISTORY_BUTTON_WIDTH`?

app/src/app/components/Toolbar/ToolButtons.tsx
```diff
@@ -1,101 +1,145 @@
...
+  // only; the hotkeys themselves still work in plain Draw.
+  const showHotkeyHints = useToolbarStore(state => state.superDraw);
   const activeTools = useActiveTools();
+  const mainTools = activeTools.filter(tool => !HISTORY_TOOLS.includes(tool.mode));
+  const historyTools = activeTools.filter(tool => HISTORY_TOOLS.includes(tool.mode));
+
+  const renderTool = (tool: ActiveToolConfig, buttonStyle: React.CSSProperties) => {
+    const IconComponent = tool.icon;
+    const isActive = activeTool === tool.mode;
+    const singleKey = tool.hotKeyLabel.length === 1;
```
fangge518: PP: `singleKey = tool.hotKeyLabel.length === 1` currently matches `mainTools` (M/P/E/B/I) exactly, and is false for every `historyTools` entry (⌘Z, ⌘⇧Z) — is that a coincidence, or intentional? Either way, I think the corner-badge/tooltip choice should key off which group a tool is in (main vs. history), not off the incidental length of its hotkey label — that way it stays correct even if a future tool's shortcut format changes.

app/src/app/globals.css
```diff
@@ -141,3 +141,14 @@ body {
...
+
+.tool-button {
+  transition:
+    background-color 120ms ease,
+    color 120ms ease,
+    box-shadow 120ms ease,
+    transform 80ms ease;
+}
+.tool-button:active:not(:disabled) {
+  transform: scale(0.96);
```
fangge518, line 153: PP: Missing a trailing newline.

fangge518 (commented):
I really like the button design of the new UI. I added a few non-blocking thoughts below. This should be ready to be merged without further review.

app/src/app/components/Toolbar/ToolControls/ToolControls.tsx
nofurtherinformation: Good call, reverted in 4c2151f8834702dd28f5c9f4790e55db91155e56

app/src/app/globals.css
nofurtherinformation: Fixed in cab9e5fb23af513b7a7ef16cc30a29a5cba818f9

app/src/app/components/Toolbar/ToolButtons.tsx
nofurtherinformation: Good catch - handled in 8f632e4477be5547477af1bea72efa2125009282

app/src/app/components/Toolbar/ToolButtons.tsx
nofurtherinformation: I tweaked the scaling but kept undo/redo smaller - I like them to be more minimal since they are momentary actions instead of more active/full tool-modes

PR #657, opened 2026-07-27
Population chart improvements

nofurtherinformation:
- Add option to collapse not yet started districts
- Clamps x-axis scale

### DIFF
--- a/app/src/app/components/sidebar/PopulationPanel/PopulationChart/PopulationChart.tsx
+++ b/app/src/app/components/sidebar/PopulationPanel/PopulationChart/PopulationChart.tsx
@@ -207,7 +207,6 @@ export const PopulationChart: React.FC<{
     [data.length, yMax]
   );
 
-  const [hoveredIndex, setHoveredIndex] = useState<number | null>(null);
   const [isHovered, setIsHovered] = useState(false);
 
   // Render helpers (not components) — they close over shared state (scales, hover, colors)
@@ -261,9 +260,6 @@ export const PopulationChart: React.FC<{
                 height={barHeight + 10}
                 className="opacity-0 hover:opacity-10 transition-opacity duration-300 cursor-pointer"
                 onClick={() => onBarSelect?.(entry.zone)}
-                onMouseEnter={() => setHoveredIndex(index)}
-                onMouseMove={() => setHoveredIndex(index)}
-                onMouseLeave={() => setHoveredIndex(null)}
               />
               <Bar
                 key={`bar-${entry.zone}`}
@@ -323,13 +319,10 @@ export const PopulationChart: React.FC<{
       height={height}
       style={{display: 'block'}}
       onMouseEnter={() => setIsHovered(true)}
-      onMouseLeave={() => {
-        setIsHovered(false);
-        setHoveredIndex(null);
-      }}
+      onMouseLeave={() => setIsHovered(false)}
     >
       <OverageHatch />
-      <Group left={margins.left} top={margins.top} onMouseLeave={() => setHoveredIndex(null)}>
+      <Group left={margins.left} top={margins.top}>
         {renderIdealReference()}
         {renderBars()}
       </Group>
--- a/app/src/app/components/sidebar/PopulationPanel/PopulationChart/PopulationChart.tsx
+++ b/app/src/app/components/sidebar/PopulationPanel/PopulationChart/PopulationChart.tsx
@@ -71,8 +71,6 @@ const usePopulationXScale = (
   const effectiveIdealPopulation = isCommunityMode ? undefined : idealPopulation;
 
   const maxPop = Math.max(...data.map(r => r.total_pop_20));
-  // ponytail: with an ideal, the domain is pinned to 130% of it so one outsized district
-  // can't squash every other bar. Bars past the cap are clipped and hatched.
   const xMaxValue = scaleToCurrent
     ? maxPop * 1.05
     : effectiveIdealPopulation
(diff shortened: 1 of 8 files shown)


app/src/app/components/sidebar/PopulationPanel/PopulationChart/PopulationLabels.tsx
```diff
@@ -46,11 +49,19 @@ export const PopulationLabels: React.FC<{
         : _popDiffLabel;
   const popLabel = formatNumber(entry.total_pop_20, NUMBER_FORMATS.STRING);
   if (popLabel === undefined) return null;
-  const [left, top] = [xScale(entry.total_pop_20), yScale(index) + 5 + barHeight / 2];
+  // Over-ideal bars anchor their labels to the ideal line — deviation to its left, population
+  // to its right — so the numbers stay put instead of chasing the (clipped) bar end.
+  const isOverIdeal = hasIdealPopulation && entry.total_pop_20 > idealPopulation;
+  const [left, top] = [
+    isOverIdeal ? xScale(idealPopulation) : Math.min(xScale(entry.total_pop_20), xMax),
```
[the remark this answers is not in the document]
nofurtherinformation: Handled in f82db8582ae6e6c04030a11f7a96e72c81f8c483

fangge518 (commented):
Overall the code looks good to me. I have only a few nits on the styles of displayed elements.

(Two more style preferences on `PopulationLabels.tsx` — posting here since the specific code they touch (the label render block, lines 76-108) isn't part of this PR's diff, so GitHub won't let me anchor inline comments there.)

PP: Currently the population number renders on the right of the anchor (`x={5}`) and the deviation number on the left (`x={-5}`, with the white halo). I'd prefer these flipped, for three reasons: (1) near completion, the deviation number matters more than the absolute population — putting it on the side that lands against the plot's white background makes it more visible than tucking it into the halo treatment. (2) For an overflow bar, the deviation number would then sit in the overflow portion (right of the ideal-line anchor) — the exact region it's describing. (3) For a deficient bar, the deviation number would sit in the gap between the bar's end and the ideal line (right of the bar-tip anchor) — again, the exact region it's describing.

O: `DemographyTable.tsx` has the same "text over a colored/shaded background" problem and solves it with `getReadableTextColor(backgroundColor, alpha)` (`app/src/app/utils/colors.ts`) — computes the background's luminance and picks white or default dark text accordingly, rather than this component's current halo technique (drawing the deviation text twice: once white with a heavy white stroke, then black on top). For whichever label lands on the bar's own solid color, that color is already known (same one used to paint the bar), so this would make the chart match the table's existing convention. The halo's one advantage — robustness to messy/mixed backgrounds without needing to know what's underneath — matters less for the solid-fill region, but probably still applies where a label could straddle the ideal-line marker or the hatch pattern.

app/src/app/components/sidebar/PopulationPanel/PopulationPanel.tsx
```diff
@@ -79,7 +79,15 @@ export const PopulationPanel = () => {
...
   const superDraw = useToolbarStore(state => state.superDraw);
-  const shouldUseScrollableRows = populationData.length > 10;
+  // Unstarted zones are padded in with total_pop_20 = 0; hide them unless asked for.
+  const startedData = useMemo(
+    () => populationData.filter(d => d.total_pop_20 > 0),
+    [populationData]
+  );
+  const showAllZones = chartOptions.popShowAllZones;
+  const visibleData = showAllZones ? populationData : startedData;
+  const hasUnstartedZones = startedData.length < populationData.length;
```
fangge518, line 89: I: `startedData` treats any zone with `total_pop_20 === 0` as unstarted and hides it by default. A district or community that's been painted but happens to cover zero-population area (water, an unpopulated commercial block) would disappear from the default view even though it's a real, assigned zone — indistinguishable from one nobody has touched. This mirrors the existing `allPainted` heuristic a few lines up, so it's not a new assumption in the codebase, but it's now driving what's shown vs. hidden rather than just a completion percentage. Worth a gut check on whether a zero-population but assigned zone is common enough to matter here.

app/src/app/components/sidebar/PopulationPanel/PopulationPanel.tsx
```diff
@@ -155,6 +163,18 @@ export const PopulationPanel = () => {
...
+            size="1"
+            variant={showAllZones ? 'solid' : 'surface'}
+            color="gray"
+            highContrast={showAllZones}
+            aria-pressed={showAllZones}
+            onClick={() => setChartOptions({popShowAllZones: !showAllZones})}
+          >
+            {`Show all ${zoneLabelPlural}`}
+          </Button>
+        )}
```
fangge518, line 177: PP: Slight preference: instead of a header button, put this where the collapsed rows themselves would be — e.g. if districts 1-3 are painted (shown) and 4+ are hidden, show it in the list at position 4; if 3-5 are painted, at position 6. The current header gets a bit crowded once the Super Draw settings button (`PopulationPanelOptions`) is also showing alongside the heading and this button.

app/src/app/components/sidebar/PopulationPanel/PopulationChart/PopulationChart.tsx
```diff
@@ -55,9 +71,13 @@ const usePopulationXScale = (
   const effectiveIdealPopulation = isCommunityMode ? undefined : idealPopulation;
 
   const maxPop = Math.max(...data.map(r => r.total_pop_20));
+  // ponytail: with an ideal, the domain is pinned to 130% of it so one outsized district
```
fangge518: PP: The comment above `xMaxValue` opens with "ponytail: with an ideal, the domain is pinned..." — looks like a leftover word that doesn't belong; the rest of the comment reads fine without it.

nofurtherinformation: Good catch - these are leftover process comments I missed - cleaned up in d1b0f4d13c4b00555d953e6547bd0c62dad9c2c1

PR #659, opened 2026-07-27
Break mode enhancements

nofurtherinformation:
- Limits view bounds when in break (focused) mode
- Adds pill on-map instructions to select a unit to break and helper to exit break mode


app/src/app/components/Map/MainMap.tsx
```diff
@@ -104,6 +105,7 @@ export const MainMap: React.FC = () => {
             <MetaLayers isDemographicMap={false} />
           </PointSource>
           <NavigationControl showCompass={false} showZoom={true} position="bottom-right" />
+          <BlockModePill />
```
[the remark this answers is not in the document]
nofurtherinformation: Mounted in 0a8c46f6e0b613923876398ff9297035e03c2d90

app/src/app/components/Toolbar/ToolControls/ToolControls.tsx
```diff
@@ -25,15 +22,11 @@ const ToolControlsConfig: Record<
...
-    Component: () => {
-      const focusFeatures = useMapStore(state => state.focusFeatures);
-      if (focusFeatures.length) {
-        return <Text>Focused on {focusFeatures[0].id}</Text>;
-      } else {
-        return <Text>Click a feature to show the census blocks within it</Text>;
-      }
-    },
+    Component: BrushControls,
   },
```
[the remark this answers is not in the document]
nofurtherinformation: Handled in 432ee7d078824effef9d44aa4a55d6532f564d0f

fangge518, line 32: I: Routing `shatter` straight to `BrushControls` means the "Brush Size" slider and zone picker render as soon as the shatter tool is armed — before a unit has been broken and there's anything to paint. That's a step ahead of what the new on-map pill is telling the user to do ("Choose a unit to break into blocks"). Worth conditioning `BrushControls`'s paint-specific children (or this config entry) on `captiveIds.size > 0` / `inBlockView`, so the sidebar only shows painting controls once painting is actually possible.

app/src/app/components/Map/BlockModePill.tsx
```diff
@@ -0,0 +1,90 @@
...
+  }, [inBlockView, bounds, getMapRef]);
+
+  useEffect(() => {
+    const onKeyDown = (e: KeyboardEvent) => {
+      if (e.key !== 'Escape') return;
+      if (inBlockView) {
+        exitBlockView();
+      } else if (activeTool === ACTIVE_TOOLS.SHATTER) {
+        setActiveTool(ACTIVE_TOOLS.BRUSH);
+      }
```
fangge518, line 72: I: This listens on `document` unconditionally. `GeocodeSearchBar.tsx` and `DistrictZonePicker.tsx` both handle `Escape` locally (closing a suggestion dropdown, blurring a field) but neither calls `stopPropagation()`, so their Escape keydowns bubble up here too — pressing Escape to close a search dropdown while inside block view would also silently exit block view. Worth ignoring the event when `document.activeElement` is an input/textarea (the pattern `BrushSizeSelector.tsx` already uses for its own bracket-key shortcuts), or calling `stopPropagation()` at the source.

fangge518 (changes requested):
Thank you for implementing this feature. I've tried using it and I think this break mechanism has far less routes for users to be confused. Here are several rouge edges that I've discovered, which should all be relatively easy fixes.

(Two more findings on files this PR doesn't touch — posting here since GitHub won't let me anchor inline comments outside this PR's diff.)

PP: Nice touch that breaking a unit automatically returns to the previous tool. One gap in the other direction: while painting inside a captive/block view (`captiveIds.size > 0`), the Break tool itself is still selectable in the toolbar — clicking it does nothing, since you're already inside the one unit you'd be breaking further into. Right now `disabled` here is only `!mapDocument?.child_layer` (line 120); could this also disable while `captiveIds.size > 0` (`useMapStore` is already in scope in this hook), so the tool doesn't invite a click that has no effect?

I: Here is a known bug that exists before this PR, but I think might be suitable to be fixed here. After breaking a unit, painting its exposed blocks into more than one zone, then exiting block view, the population labels on those blocks stay on screen (exiting without painting, or painting them all the same zone, works fine). Root cause: on exit `captiveIds` becomes an empty `Set` (correctly), but the child `PopulationTextLayer`'s filter is `['match', ['get', 'path'], Array.from(captiveIds), true, false]` — an empty `captiveIds` makes the label list `[]`, and MapLibre's `match` expression requires at least one branch label (`@maplibre/maplibre-gl-style-spec`'s `match.ts`, "Expected at least one branch label" — this is the exact console error I hit on exit). The filter update fails validation and never takes effect, so the layer keeps whatever filter it last had while `captiveIds` was still populated — which still matches those specific blocks. Normally invisible because a successful heal (all children same zone) also clears the population *source*'s own data, so there's nothing left for the stale filter to match; painting into different zones skips the heal, the source keeps its data, and the stuck filter keeps matching it. Fix is narrow: guard the empty case, e.g. `captiveIds.size ? ['match', ['get', 'path'], Array.from(captiveIds), true, false] : ['literal', false]`.

If this feels out of scope, we can always file another small PR to fix it.

fangge518 (approved):
Just flipping the earlier verdict because the bug is not necessarily in scope.

app/src/app/components/Toolbar/ToolControls/ToolControls.tsx
nofurtherinformation: Cleaned up in 29df5b97d655b8fdf42aecb5ede8a2bb417d5f6c

app/src/app/components/Map/BlockModePill.tsx
nofurtherinformation: Handled in 0b059c47a32748432a18a88de31c45e6f57bf027

nofurtherinformation:
Thanks Ge! Orphan population labels are handled in 2efc8cddc1249854d6fd5947defe6609f1a4bdc7  and 29df5b97d655b8fdf42aecb5ede8a2bb417d5f6c adds the child layer check to disable break

PR #660, opened 2026-07-27
Show all variables in choropleth map

nofurtherinformation:
Currently, only totpop/VAP variables show based on the segmented control in the table view. This allows them all to be displayed


fangge518 (approved):
LGTM.

PR #662, opened 2026-07-27
Render district numbers above population numbers at shared centroids

fangge518:
District-number and block/precinct population-count labels are both symbol layers placed at the same centroid, with identical `text-anchor`/`text-offset`. Neither passed a `beforeId`, so each was appending itself to the absolute top of the layer stack the moment its own data source became ready — a race, not something set by render order. Population's source resolves later in practice, so it consistently won and hid the district number underneath.

**Fix:** give the population-text layers `beforeId: MAP_LAYER_ANCHOR_IDS.reference` (the same anchor `CountyLayers` already uses). That anchor is created before any real layer mounts, so population now lands in a fixed position instead of racing for the top; the zone-number layer, left unanchored, always ends up above it.

Correct stacking order wasn't quite enough on its own: the district-number circle wasn't fully opaque, so the population digits underneath still showed through.

**Fix:** `circle-opacity` on the district-number background circle, `0.8 -> 0.9`, so it sufficiently hides the population number beneath it.

### DIFF
--- a/app/src/app/components/Map/PointLayers/MetaLayers.tsx
+++ b/app/src/app/components/Map/PointLayers/MetaLayers.tsx
@@ -4,6 +4,7 @@ import {
   SELECTION_POINTS_SOURCE_ID_CHILD,
   ZONE_LABEL_SOURCE_ID,
   ZONE_LABEL_LAYER_IDS,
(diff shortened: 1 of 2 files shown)


nofurtherinformation (approved):
LGTM

PR #663, opened 2026-07-28
Turnstile migration

nofurtherinformation:
## Description

Replaces both reCAPTCHA flows with Cloudflare Turnstile, using two widgets that mirror the old setup:

- **Comment form** (was reCAPTCHA v2 checkbox): a Turnstile **Managed** widget rendered by the renamed `useTurnstile` hook. Tokens are single-use, so the widget resets when the store clears the token after submit. The `react-google-recaptcha` dependency is removed — a small shared loader (`app/src/app/utils/turnstile.ts`) renders widgets explicitly, no new dependency.
- **Silent session minting** (was reCAPTCHA v3): a Managed widget rendered with `appearance: interaction-only` — silent for trusted browsers, but surfaces a manual challenge bottom-right when Cloudflare requires interaction, instead of silently losing the session.
- **Backend**: one `siteverify` helper against Cloudflare's endpoint behind two thin wrappers, one per widget secret. The v3 score threshold and action check are dropped — Turnstile has no scores, and per-widget secrets already prevent cross-widget token replay.
- **Env renames** (deploy workflows, frontend Dockerfile args, Pulumi config all updated):
  - `RECAPTCHA_SECRET_KEY` → `TURNSTILE_SECRET_KEY`
  - `RECAPTCHA_V3_SECRET_KEY` → `TURNSTILE_SESSION_SECRET_KEY`
  - `NEXT_PUBLIC_RECAPTCHA_SITE_KEY` → `NEXT_PUBLIC_TURNSTILE_SITE_KEY`
  - `NEXT_PUBLIC_RECAPTCHA_V3_SITE_KEY` → `NEXT_PUBLIC_TURNSTILE_SESSION_SITE_KEY`
  - `RECAPTCHA_V3_SCORE_THRESHOLD` deleted
- The `recaptcha_token` wire field (and the `*WithRecaptcha` model names mirroring it) are deliberately kept — they now carry Turnstile tokens; renaming would churn every client and test for no behavior change.
- `.env.docker.example` files now use Cloudflare's public always-pass dummy keys, so local dev works out of the box.

**Before merging** (see migration runbook, kept out of the repo): create the two Turnstile widgets in the Cloudflare dashboard, set the new GitHub Actions secrets (`TURNSTILE_SITE_KEY`, `TURNSTILE_SESSION_SITE_KEY`, `TURNSTILE_SECRET_KEY`), and set the Pulumi stack secrets (`turnstileSecretKey`, `turnstileSessionSecretKey`) for dev and prod. If the Pulumi secrets are missing at deploy time, comment endpoints reject everything and session minting skips verification.

**Expected cutover blip:** browser tabs holding the old bundle will fail captcha (Google tokens sent to Turnstile verification) until reload. Already-minted session JWTs keep working (HMAC signing unchanged, 4h TTL).

## Reviewers

- Primary:
- Secondary:

## Checklist

- [x] Added/Updated related documentation (backend/infra READMEs, `.agents` guides, env examples).
- [x] Added/Updated related unit tests (existing suite renamed to the Turnstile helpers; 327 passed / 117 skipped).

## Review required

- **Cloudflare dashboard assumptions**: the two-widget split (Managed for the form, Invisible for sessions) and hostname coverage need human validation against the actual Cloudflare account — code can't verify widget modes match.
- **Secret rotation ordering**: confirm GitHub secrets + Pulumi config are set *before* merge (runbook step 2–3); wrong ordering deploys a backend with captcha unset.
- **Comment form UX**: manual check on dev that the Managed widget renders, the submit button gates on the token, and a second submission after the first works (widget reset path).
- **Silent session flow**: verify on dev that `POST /api/session` returns 200 with a fresh browser profile and that gated endpoints work — headless verification of the invisible widget wasn't possible locally.
- **Session fallback UX:** verify on dev (using the 3x…FF dummy site key) that the challenge surfaces bottom-right, completing it mints a session, and ignoring it degrades gracefully after the timeout.


nofurtherinformation:
# Turnstile Migration Runbook (env + secrets)

Branch `fix/turnstile-migraiton` swaps Google reCAPTCHA for Cloudflare Turnstile. Code, deploy scripts, and Pulumi config are done; the steps below are the manual secret/dashboard changes only you can do.

## Env name mapping

| Old | New | Where |
|---|---|---|
| `RECAPTCHA_SECRET_KEY` | `TURNSTILE_SECRET_KEY` | backend (comment-form widget) |
| `RECAPTCHA_V3_SECRET_KEY` | `TURNSTILE_SESSION_SECRET_KEY` | backend (invisible session widget) |
| `RECAPTCHA_V3_SCORE_THRESHOLD` | *(deleted — Turnstile has no scores)* | backend |
| `NEXT_PUBLIC_RECAPTCHA_SITE_KEY` | `NEXT_PUBLIC_TURNSTILE_SITE_KEY` | frontend build |
| `NEXT_PUBLIC_RECAPTCHA_V3_SITE_KEY` | `NEXT_PUBLIC_TURNSTILE_SESSION_SITE_KEY` | frontend build |

The API wire field `recaptcha_token` was deliberately kept (it now carries
Turnstile tokens) — no client contract change.

## 1. Create the Turnstile widgets (Cloudflare dashboard)

Cloudflare dashboard → account home → **Turnstile** → Add widget, twice:

1. **"districtr comment form"** — widget mode **Managed** (visible checkbox-style, auto-passes most humans). Replaces reCAPTCHA v2.
2. **"districtr session"** — widget mode **Managed** as well (NOT Invisible). Replaces silent reCAPTCHA v3 for `/api/session` minting. The frontend renders it with `appearance: interaction-only`, so it's invisible for Cloudflare trusts; when Cloudflare demands interaction, it bottom-right so the user can complete the challenge manually an Invisible-mode widget could never show this fallback).

Hostnames for both: `districtr.org` (covers subdomains, so dev/app subdomains are included) plus any other apex domains you serve. Note: `fly.dev` is on the Public Suffix List and can't be added wholesale — for Fly PR previews either add individual preview hostnames or just set the preview GH secrets to Cloudflare's public dummy keys (below).

Each widget gives you a **Site Key** (frontend) and **Secret Key** (backend).

## 2. GitHub Actions secrets (repo → Settings → Secrets and variables → Actions)

Add:

| Secret | Value | Consumed by |
|---|---|---|
| `TURNSTILE_SITE_KEY` | comment-form widget site key | `deploy-app.yml`, `fly-deploy-app.yml`, `fly-preview.yml` |
| `TURNSTILE_SESSION_SITE_KEY` | session widget site key | `deploy-app.yml` |
| `TURNSTILE_SECRET_KEY` | comment-form widget secret | `fly-preview.yml` (fullstack previews, optional) |

Then delete the old `RECAPTCHA_SITE_KEY`, `RECAPTCHA_V3_SITE_KEY`,
`RECAPTCHA_SECRET_KEY`.

## 3. Pulumi backend secrets (AWS stacks — repeat for `dev` and `prod`)

```bash
cd infra
export PULUMI_BACKEND_URL='s3://districtr-v2-pulumi-state?region=us-east-2'
pulumi stack select prod   # then again with dev
pulumi config set --secret turnstileSecretKey        '<comment-form widget secret>'
pulumi config set --secret turnstileSessionSecretKey '<session widget secret>'
pulumi config rm recaptchaSecretKey
pulumi config rm recaptchaV3SecretKey
```

The old `RECAPTCHA_*` SSM parameters are Pulumi-managed and get deleted automatically on the next `pulumi up` (the resources are gone from the program). No manual SSM cleanup needed.

## 4. Fly backend (only if the Fly API apps are still live)

```bash
fly secrets set TURNSTILE_SECRET_KEY='<form secret>' TURNSTILE_SESSION_SECRET_KEY='<session secret>' -a districtr-v2-api
fly secrets unset RECAPTCHA_SECRET_KEY RECAPTCHA_V3_SECRET_KEY -a districtr-v2-api
# repeat for districtr-v2-api-dev
```

## 5. Local dev

Already done in this checkout: both `.env.docker` files (untracked) and the committed `.env.docker.example` files now use Cloudflare's **public dummy keys** (always pass, official test values — safe to commit):

- Site keys: `1x00000000000000000000AA` (always pass); swap the session one to
  `3x00000000000000000000FF` to force the interactive fallback for testing
- Secret: `1x0000000000000000000000000000000AA`

Restart the containers to pick them up: `docker-compose restart backend frontend`. (The running backend container still has the old `RECAPTCHA_*` env baked in; they're harmlessly ignored by the new config until restart.)

## 6. Deploy order + expected blip

Merge deploys backend and frontend from the same push. Any browser tab holding the **old** bundle across the cutover will fail captcha (Google tokens sent to Turnstile verification) — comment submits and *new* session mints 400/401 until the tab reloads. Already-minted session JWTs keep working (HMAC signing unchanged, 4h TTL). No action needed; it self-heals on reload.

If the backend deploys with the Pulumi secrets **not yet set**: comment endpoints reject everything (400) and `/api/session` mints without verification (verification is skipped when `TURNSTILE_SESSION_SECRET_KEY` is unset). So do step 3 **before** merging.

## 7. Verify

```bash
# Backend env landed (per stack):
aws ssm get-parameters-by-path --path /districtr/prod/backend --query 'Parameters[].Name'
# expect .../TURNSTILE_SECRET_KEY and .../TURNSTILE_SESSION_SECRET_KEY, no RECAPTCHA_*

# Session minting end-to-end: load the site, then in devtools console:
#   localStorage.removeItem('districtr_session'); location.reload()
# open a map — Network tab should show POST /api/session → 200 {token,...}

# Comment form: submit a test comment on dev; widget should render and pass.
```

backend/app/core/config.py
```diff
@@ -79,12 +79,12 @@ def server_host(self) -> str:
     POSTGRES_DB: str = ""
     DATABASE_URL: str
 
-    # reCAPTCHA
-    RECAPTCHA_SECRET_KEY: str | None = None
+    # Cloudflare Turnstile (comment-form widget)
+    TURNSTILE_SECRET_KEY: str | None = None
```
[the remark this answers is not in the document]
nofurtherinformation: Handled in e50deae21874748a32ad1ae91f48467f48ce8edc, although, let's deprecate fly ASAP.

infra/config.ts
```diff
@@ -40,8 +40,8 @@ export const config = {
   auth0ClientSecret: cfg.requireSecret("auth0ClientSecret"),
   auth0SessionSecret: cfg.requireSecret("auth0SessionSecret"),
   openaiApiKey: cfg.getSecret("openaiApiKey"),
-  recaptchaSecretKey: cfg.getSecret("recaptchaSecretKey"),
-  recaptchaV3SecretKey: cfg.getSecret("recaptchaV3SecretKey"),
+  turnstileSecretKey: cfg.getSecret("turnstileSecretKey"),
+  turnstileSessionSecretKey: cfg.getSecret("turnstileSessionSecretKey"),
```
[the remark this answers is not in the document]
nofurtherinformation: Same as above

backend/app/comments/models.py
```diff
@@ -108,6 +108,9 @@ class CommenterCreate(BaseModel):
     zip_code: str | None = None
 
 
+# ponytail: the recaptcha_token wire field (and the *WithRecaptcha names that
+# mirror it) now carry Cloudflare Turnstile tokens; renaming would churn every
+# client and test for no behavior change.
```
peterrrock2: O: This seems like a reasonable note, but we make the change from recaptcha to turnstile pretty much everywhere else, so I would argue that leaving it breaks parallelism in a bit of a confusing way.

backend/app/main.py
```diff
@@ -291,13 +291,15 @@ async def db_is_alive(session: Session = Depends(get_session)):
 
 
 class SessionCreate(BaseModel):
+    # ponytail: wire name kept from the reCAPTCHA era; it now carries a
+    # Turnstile token. Renaming would churn every client/test for no behavior.
     recaptcha_token: str
```
peterrrock2: O: (dup) We make the rename everywhere else, so I don't see what this is buying us. An lsp should handle this rename quickly

backend/app/comments/main.py
```diff
@@ -456,7 +456,7 @@ async def create_commenter(
 ):
     """Create a new commenter with upsert on conflict for name + email."""
     client_host = request.client.host if request.client else ""
-    await recaptcha.verify_recaptcha(commenter_data.recaptcha_token, client_host)
+    await turnstile.verify_turnstile(commenter_data.recaptcha_token, client_host)
```
peterrrock2: I: Wait, this is on the BE behind the load balancer, right? So won't `client_host` be the private IP of the load balancer rather than the user's IP?

peterrrock2 (commented):
Overall this is looking good. Only one "I" point

Other notes:
- the correct secrets exist for the fly deploy
-  it looks like we don't intend to turn on turnstile for fly, but if we do, we are missing the `TURNSTILE_SESSION_SECRET_KEY` secret

backend/app/comments/main.py
fangge518: Yes. I've fixed it and extracted a function to be called by several references of client.host.

backend/app/comments/models.py
fangge518: The comments have been tidied up.

fangge518 (approved):
The code looks good to me. I've also verified that the two keys have been set on dev and prod on aws. I've also set the two keys on fly dev. This should allow this PR to be safely merged.

PR #664, opened 2026-07-28
Updated political contests

nofurtherinformation:
Adds missing political contests to FE config.

At runtime, this gets filtered to only valid/existing contests in the data.


fangge518 (approved):
LGTM.

PR #666, opened 2026-07-29
Count county pieces by connected components, and add a forced-minimum-split precompute

fangge518:
`county_pieces` counted a county as split into as many pieces as the number of distinct district zones touching it. A district split into two disconnected areas within a county (e.g. wrapping around a lake, or an island precinct) counted as one piece, not two — the metric only measured how many zones touched a county, not how many separate physical pieces it was actually cut into.

**Fix:** `county_pieces` now counts geographically connected components. For each (county, zone) pair, it intersects the zone's assigned units with the county, expands any non-contiguous parent units to their block children, and counts connected components on the adjacency graph.

The "forced minimum" split count is now `sum(ceil(component_pop / ideal_pop))` over the county's own pre-existing connected components, computed independently of any plan and precomputed/cached per gerrydb table (`evaluation.county_demographics.component_populations`). The frontend's "unnecessarily split" determination and Splits panel both use it.

## Tested

- [x] Backend suite passes (337 passed) including new/updated tests in test_splits.py, test_county_context.py, test_partisan.py, test_contiguity.py
- [x] Manual verification against real Colorado data (co_custom_districts_v2): all 64 counties' component_populations sum to their total_pop; found real disconnected counties (Arapahoe County: 8 components, Boulder County: 2 components)
- [x] Splits panel UI: "Disconnected Parts" column and updated County Splits description render correctly in the browser (not re-checked after the latest text/column edits)


app/src/app/components/EvalPanel/CountySplitsSection.tsx
```diff
@@ -39,13 +39,28 @@ export const CountySplitsSection: React.FC<CountySplitsSectionProps> = ({evaluat
...
     .sort((a, b) => a.name.localeCompare(b.name));
 
+  // Forced minimum: sum of ceil(component pop / ideal) over the county's own
+  // connected components (islands/exclaves force at least one district each,
+  // independent of the plan) — a tighter bound than ceil(total_pop / ideal).
+  // Falls back to the single-total bound when component data is unavailable.
+  const forcedMinimum = (e: (typeof allEntries)[number], ideal: number) =>
+    e.componentPopulations.length > 0
+      ? e.componentPopulations.reduce((sum, p) => sum + Math.ceil(p / ideal), 0)
+      : Math.ceil(e.pop / ideal);
```
peterrrock2, line 58: H: I think that this computation is slightly off since some counties have 0 population pieces. For example, Jefferson County in CO has 3 such components and will always appear as "unnecessarily split" no mater what you do. I think that you need
```suggestion
  const forcedMinimum = (e: (typeof allEntries)[number], ideal: number) =>
    e.componentPopulations.length > 0
      ? e.componentPopulations.reduce((sum, p) => sum + Math.max(1, Math.ceil(p / ideal), 0))
      : Math.ceil(e.pop / ideal);
```

app/src/app/components/EvalPanel/CountySplitsSection.tsx
```diff
@@ -39,13 +39,28 @@ export const CountySplitsSection: React.FC<CountySplitsSectionProps> = ({evaluat
...
+
   const splitCounties = allEntries.filter(e => e.actual >= 2).length;
```
peterrrock2, line 60: H: I think that this will now be wrong too. I think that we need to invert the `district_county_membership` for this check

peterrrock2 (changes requested):
BE checked against my own, independent implementation for a 10k step ReCom chain on CO, and everything appears to be lining up.

Only other nit is that "Disconnected Parts" reads strangely in the table. It might be better to use something like "Connected Pieces" or "Contiguous Pieces".

I am going to double-check that this is exactly how we want to count before we merge this in.

peterrrock2:
:warning: This is the computation for the "County Parts" metric, not "County Pieces". Apologies, this was a miscommunication on my end. Let's leave the branch up but close the PR.

fangge518:
Closing per discussion — this branch's computation is actually a new metric (county_parts, connected-component counting), not a fix to county_pieces (which was correct all along). Will revert county_pieces to its original behavior and land county_parts as a new metric, then open a fresh PR from this branch with a title/description matching that scope.

PR #667, opened 2026-07-29
Fix: VAP overlay memory

fangge518:
Picking a VAP variable (e.g. "VAP Black") as the map's choropleth overlay, then toggling the overlay off and back on in Visual Settings, resets the selection to a generic "Total" instead of restoring "VAP Black." Doing the identical sequence with a TOTPOP variable works correctly.

Visual Settings' "Demographic" toggle only ever calls `activateOverlayGroup('TOTPOP')` — there's no separate VAP button. `overlayMemory` kept a variable slot per `SummaryType` (TOTPOP/VAP/VOTERHISTORY — a statistical-universe taxonomy), so a VAP pick was written correctly but never read back: the reactivation path only ever checks the TOTPOP slot. TOTPOP round-tripped fine only because its own write and read already shared a key. Using `SummaryType` to drive the toggle's own, coarser concept (which of two buttons is active) is what made this possible in the first place.

**Fix:** introduce `OverlayGroup` (`'demography' | 'election'`) as the toggle's own type, independent of `SummaryType` — alongside `SummaryType` in `constants/demography/summary.ts`, since it's pure taxonomy with no store dependency. `overlayMemory` stores its remembered variable as `Record<OverlayGroup, DemographyVariable | null>` — a dict again, but keyed by the correct concept this time, so it carries none of the taxonomy conflation the old `SummaryType`-keyed one did. `toOverlayGroup(summaryType)` maps a variable's actual universe to its overlay group; `activateOverlayGroup` and the sidebar's Map Layer tab both index `overlayMemory.variables` by it directly instead of touching `SummaryType` values as memory keys. TOTPOP and VAP now always land in the same (`'demography'`) slot, so a VAP pick survives a toggle off/on exactly like a TOTPOP pick does.

Also removes `overlayMemory.lastGroup`: `activateOverlayGroup`'s own memory write was already unconditional, making it redundant on that path.

### DIFF
--- a/app/src/app/components/Toolbar/Settings.tsx
+++ b/app/src/app/components/Toolbar/Settings.tsx
@@ -19,14 +19,12 @@ import {ColorChangeModal} from './ColorChangeModal';
 import {useAssignmentsStore} from '@/app/store/assignmentsStore';
 import {ACCESS_STATES} from '@constants/document/state';
 import {DEMOGRAPHIC_MODES} from '@constants/map/demographicMode';
-import {SUMMARY_TYPES} from '@constants/demography/summary';
+import {SUMMARY_TYPES, type OverlayGroup} from '@constants/demography/summary';
 import {OVERLAY_OPACITY} from '@/app/constants/document/limits';
 import {
   activateOverlayGroup,
-  getOverlayVariable,
   overlayGroupVariables,
   overlayMemory,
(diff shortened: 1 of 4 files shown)


nofurtherinformation:
LGTM! Code changes are well scoped and UX works on my end

PR #669, opened 2026-07-30
Save before export

### LINKED ISSUE #613: If you export the assignments from an unsaved map with changes, you will export the last save, and then the map will save itself. So if you hit export twice without making any changes, you will get two different exports.
If you export the assignments from an unsaved map with changes, you will export the last save, and then the map will save itself. So if you hit export twice without making any changes, you will get two different exports.

nofurtherinformation:
## Description
- Resolves #613, adding a save before exporting
- Adds better response interfaces to the assignments and coi assignment store save handlers, to more easily access the business logic flow while chaining operations

### DIFF
--- a/app/src/app/components/Topbar/MapActionsDropdown.tsx
+++ b/app/src/app/components/Topbar/MapActionsDropdown.tsx
@@ -24,11 +24,11 @@ export const MapActionsDropdown: React.FC<{
   const setNotification = useMapStore(state => state.setNotification);
   const {isOutdated, save} = useMapSaveStatus();
 
-  const notifyExportFailed = (reason: string) =>
+  const notifyExportFailed = (reason: string, message?: string) =>
     setNotification({
       importance: 2,
       type: 'error',
-      message: 'Exporting this map failed. Please try again in a moment.',
+      message: message ?? 'Exporting this map failed. Please try again in a moment.',
       id: `export-failed-${exportId}-${reason}`,
     });
 
@@ -49,10 +49,18 @@ export const MapActionsDropdown: React.FC<{
     // Fetch via the session-aware client (plain anchor navigation can't attach
     // the X-Districtr-Session header) and save the blob through a transient
     // anchor. Filename comes from the backend's Content-Disposition.
+
+    if (isOutdated && access === ACCESS_STATES.EDIT) {
+      try {
+        await save();
+      } catch (e) {
+        notifyExportFailed(
+          `Failed during save: ${e}`,
+          'Exporting this map failed while saving. Please try saving your map manually and then try exporting again.'
+        );
+      }
+    }
     try {
-      // The backend exports the last saved state, so flush unsaved changes first
-      // or the download lags one save behind (issue #613).
-      if (isOutdated && access === ACCESS_STATES.EDIT) await save();
       const response = await fetchWithSession(
         `${process.env.NEXT_PUBLIC_API_URL}/api/document/${exportId}/export?export_type=${exportType}`
       );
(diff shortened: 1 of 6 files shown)


app/src/app/components/Topbar/MapActionsDropdown.tsx
```diff
@@ -47,6 +49,8 @@ export const MapActionsDropdown: React.FC<{
     // Fetch via the session-aware client (plain anchor navigation can't attach
     // the X-Districtr-Session header) and save the blob through a transient
     // anchor. Filename comes from the backend's Content-Disposition.
+    const saveResponse = await save();
```
fangge518: I: Every other call site of `useMapSaveStatus().save` — `SaveButton.tsx` and `useAutoSave.ts` — gates on `access === ACCESS_STATES.EDIT` before calling it. This one doesn't, and the export menu is explicitly built to work for view-only viewers (the `public_id` fallback above `exportId`). Checked the backend: `PUT /api/assignments` doesn't check edit access either, only document existence and map-type match — so on a fresh page load this PUT is harmless, just re-confirming unchanged assignments. But if the document was edited elsewhere since this viewer's page loaded, the stale `updated_at` trips the optimistic-concurrency conflict check, and the save-conflict modal (Keep local / Use local / Use server / Fork) pops up in front of a read-only viewer who never asked to save anything. `downloadExport` then silently returns — no toast, export just doesn't happen, no explanation.

app/src/app/components/Topbar/MapActionsDropdown.tsx
```diff
@@ -47,6 +49,8 @@ export const MapActionsDropdown: React.FC<{
...
+    if (!saveResponse.ok) return;
```
fangge518: PP: `if (!saveResponse.ok) return;` exits with no call to `notifyExportFailed`, unlike the fetch-failure and network-failure paths a few lines below that both notify. A save failure right before export leaves the user with no toast at all.

app/src/app/components/Topbar/MapActionsDropdown.tsx
```diff
@@ -21,12 +22,13 @@ export const MapActionsDropdown: React.FC<{
   const access = useMapStore(state => state.mapStatus?.access);
   const handleReset = useMapStore(state => state.handleReset);
   const setNotification = useMapStore(state => state.setNotification);
+  const {isOutdated, save} = useMapSaveStatus();
```
fangge518: PP: `isOutdated` is destructured from `useMapSaveStatus()` but never read in this file. Separately, `notifyExportFailed`'s new `message?: string` parameter is never passed by either call site — both just pass a `reason` and fall back to the default message.

fangge518 (commented):
The fix correctly closes the stale-export race from #613. One I and two PP below — the I is a real gap in the view-only export path.

app/src/app/components/Topbar/MapActionsDropdown.tsx
nofurtherinformation: Gated in 0db86cac95bdaec2b48127394959520910e27dbb-- good catch!

app/src/app/components/Topbar/MapActionsDropdown.tsx
nofurtherinformation: Dead code removed in 0db86cac95bdaec2b48127394959520910e27dbb

app/src/app/components/Topbar/MapActionsDropdown.tsx
nofurtherinformation: Because there is an unhappy path on save that we don't want to toast -- conflict resolution -- I'm handling the notifications in the store, rather than in this function that calls save as a preliminary task. All bad pathways are toasted in 8e6d8b88af19f736f98e5fb58567ead9d9e66a61

PR #671, opened 2026-07-30
Static site launch banner and text updates

nofurtherinformation:
## Description
- Misc text updates
- Changelog and update page
- Adds an announcement banner for folks wanting old districtr

### DIFF
--- a/app/src/app/(static)/changelog/changelog.mdx
+++ b/app/src/app/(static)/changelog/changelog.mdx
@@ -1,7 +1,7 @@
 
 ## v2.3.0
 
-**July 2026**
+**August 2026**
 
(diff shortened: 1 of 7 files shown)


app/src/app/(static)/changelog/changelog.mdx
```diff
@@ -1,4 +1,76 @@
 
+## v2.3.0
+
+**July 2026**
```
fangge518: PP: Assuming we will release it on August 3rd, this file should describe the latest change as if it is release day, so we don't need to change it again when shipping.

fangge518 (approved):
LGTM.

app/src/app/(static)/changelog/changelog.mdx
nofurtherinformation: Updated!

PR #672, opened 2026-07-30
Improve submission form

nofurtherinformation:
## Description
- In general, submission portals are in decent shape
- This PR cleans up a small bug with the tags persisting between portals
- This PR improves the map selection dropdown, which was a bit of a mess previous


fangge518 (commented):
I've added some fixes after merging in dev and now this PR looks good to me.

PR #673, opened 2026-07-30
Env-scope thumbnail S3 keys and fix hardcoded OG image domain

fangge518:
### Thumbnail S3 keys collide across environments

Every environment — dev, preview, Fly, AWS — writes map thumbnails to the same S3 bucket at `thumbnails/{public_id}.png`, with no environment segment in the key. `public_id` is a small sequential integer assigned independently per environment, so generating a thumbnail in dev for map #1005 silently overwrites production's real #1005 thumbnail.

**Fix:** the S3 key now includes an environment folder. Split the shared path helper in two: `get_thumbnail_file_path` (public_id-keyed, environment-scoped) and `get_blank_thumbnail_file_path` (`districtr_map_slug`-keyed, intentionally left unprefixed since that string is the same in every environment by design). Only production gets its own folder — `local`, `development`, `qa`, and `test` all collapse into one shared `development` folder, so non-prod thumbnails don't spread across more S3 folders than needed. The three frontend call sites that build the CDN URL directly (`MapSelector`, `PlanGalleryRenderers`, the admin thumbnails page) now go through a shared `thumbnailUrl()` helper mirroring the same collapse logic, reading a new `NEXT_PUBLIC_ENVIRONMENT` set alongside `NEXT_PUBLIC_API_URL` in each deployment's env config (both the Fly and AWS frontend deploy workflows). Kept as direct CDN reads rather than routing through the backend.

### OG image/logo URLs hardcoded to production's domain

`pageMetadataUtils.ts` hardcoded both the per-map OG image URL and the Districtr logo fallback to `beta.districtr.org`, regardless of which environment served the page. Sharing a `dev.districtr.org` link could unfurl a same-numbered production document's thumbnail instead of the one actually being shared.

**Fix:** derive the origin from the current request via `headers()` instead of a hardcoded string, so the image URLs resolve against whichever environment actually served the page.

## Tested

- [x] Backend: `pytest` passes for `test_thumbnails.py`, including new tests asserting the S3 key differs by environment for the same `public_id`, and that `local`/`development`/`qa`/`test` all resolve to the same folder.
- [x] Backend: full `pytest` suite passes (328 passed, 117 skipped — no regressions).
- [x] Frontend: `bun run ts` and `bun run build` both pass clean.
- [x] `docker-compose up pre-commit` passes.
- [x] Live: a freshly-generated thumbnail lands at the new environment-scoped key in a deployed environment.
- [x] Live: sharing a dev/preview link unfurls the correct document's OG image.

### DIFF
--- a/app/src/app/components/Forms/MapSelector.tsx
+++ b/app/src/app/components/Forms/MapSelector.tsx
@@ -259,7 +259,11 @@ const MapSelectorInner: React.FC<MapSelectorProps> = ({allowListModules}) => {
       {notification?.type === 'success' && (
         <object data="/home-megaphone-square.png" type="image/png" className="size-32">
           <img
-            src={thumbnailUrl(dataResponse?.mapInfo?.public_id ?? '')}
+            src={
+              dataResponse?.mapInfo?.public_id != null
+                ? thumbnailUrl(dataResponse.mapInfo.public_id)
+                : undefined
+            }
             alt="Map thumbnail"
             className="size-32"
           />
--- a/app/src/app/components/Forms/MapSelector.tsx
+++ b/app/src/app/components/Forms/MapSelector.tsx
@@ -3,7 +3,7 @@ import {useEffect, useRef, useState} from 'react';
 import {useFormState} from '@/app/store/formState';
 import {getDocument} from '@/app/utils/api/apiHandlers/getDocument';
 import {DocumentObject} from '@/app/utils/api/apiHandlers/types';
-import {TILESET_URL} from '@/app/utils/api/constants';
+import {thumbnailUrl} from '@/app/utils/api/thumbnailUrl';
 import {queryClient} from '@/app/utils/api/queryClient';
 import {
   Blockquote,
@@ -259,7 +259,7 @@ const MapSelectorInner: React.FC<MapSelectorProps> = ({allowListModules}) => {
       {notification?.type === 'success' && (
         <object data="/home-megaphone-square.png" type="image/png" className="size-32">
           <img
-            src={`${TILESET_URL}/thumbnails/${dataResponse?.mapInfo?.public_id}.png`}
+            src={thumbnailUrl(dataResponse?.mapInfo?.public_id ?? '')}
             alt="Map thumbnail"
             className="size-32"
           />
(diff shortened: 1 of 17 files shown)


app/src/app/utils/metadata/pageMetadataUtils.ts
```diff
@@ -1,18 +1,22 @@
...
 export type MetadataProps = {
   params?: Promise<{public_id?: string}>;
   searchParams?: Promise<{document_id?: string | string[] | undefined}>;
 };
 
+/** The serving environment's own origin — never a hardcoded domain, so OG
+ * images/logo resolve correctly on dev/preview, not just production. */
+async function getRequestOrigin(): Promise<string> {
+  const requestHeaders = await headers();
+  const protocol = requestHeaders.get('x-forwarded-proto') ?? 'https';
```
nofurtherinformation: PP: Claude suggests this _can_ be a comma-separated list in the save of forwarding, suggested cheap hardening: `.split(',')[0]`.

app/src/app/components/Forms/MapSelector.tsx
```diff
@@ -259,7 +259,7 @@ const MapSelectorInner: React.FC<MapSelectorProps> = ({allowListModules}) => {
       {notification?.type === 'success' && (
         <object data="/home-megaphone-square.png" type="image/png" className="size-32">
           <img
-            src={`${TILESET_URL}/thumbnails/${dataResponse?.mapInfo?.public_id}.png`}
+            src={thumbnailUrl(dataResponse?.mapInfo?.public_id ?? '')}
```
nofurtherinformation: PP: `thumbnailUrl` yields ` .../development/.png` with an empty string, maybe better to leave the `src` undefined when public ID is missing (should never happen)?

nofurtherinformation (commented):
Code LGTM! 

One blocker -- `publish_district_stats_to_s3` (`backend/app/utils.py:978`) is going to have the same issue at line `:1032`. While we are fixing this, can we update there as well with the same approach?

A couple other nits in the code that are unlikely to ever be problems :)

app/src/app/components/Forms/MapSelector.tsx
fangge518: Good catch. Fixed.

app/src/app/utils/metadata/pageMetadataUtils.ts
fangge518: Removed this change as this is cover by PR 670's use of metadataBase

fangge518:
Reply to Dylan's flag of general issue: also applied the scoping to district_union's upload to S3.

Issue #676, opened 2026-07-31
Launch 02 - Draft-status-driven helper box (Get started → Refine and validate → Advanced)


nofurtherinformation:
### Context

`feat/draw-mode-enhancements` prototyped a "Getting started" checklist card (`app/src/app/components/sidebar/GettingStarted.tsx` on that branch, 411 lines) driven purely by live map state. This issue reworks that concept so the checklist is staged by the document's `draft_status` (scratch / in_progress / ready_to_share), which already exists in metadata and is editable from the topbar (`MapStatus`). Use the branch as reference for the checklist mechanics and contextual hints — **don't merge the branch itself** (it contains unrelated toolbar/meter work).

## Requirements

1. The helper box shows one of three stages, keyed to `draft_status`:
   - **scratch — "Get started"**: two checklist items:
     - Start drawing all districts (k/n painted)
     - Assign all population (unassigned remaining)
     - When both complete, an option appears to advance status to **In Progress**.
   - **in_progress — "Refine and validate"**: two checklist items:
     - Balance district populations — done when every district is within **10%** of ideal (drop the branch's 1% "fine" tier)
     - Keep districts contiguous (k/n contiguous)
     - When both complete, an option appears to advance to **Ready to Share**.
   - **ready_to_share — "Advanced"**: no checkmarks; inline, well-integrated pointer buttons to: Share (Save/Share modal), Demographics, Elections, Super Draw, Evaluate mode.
2. **Advancement is opt-in and gated**: status only moves forward via the user clicking the advance option, which only appears when the stage's criteria are met. Despite this gating, users can still actively change their draft status through metadata edit, because different map drawing session might need slightly different criteria.  Existing maps already past criteria stay where they are; gating affects forward moves only.
3. **Contiguity staleness**: contiguity only recomputes on save (query keys on `updated_at`). When there are unsaved changes, show the contiguity line grayed out with `(?/n)` instead of a stale count, ideally with an "updates on save" note. A related change: tune down the loadness of the current continguity notification, make the unsaved prompt less alarming, and a styled line of text near the section title of "validity check".
4. Keep the branch's contextual hint buttons within their stages (paint by counties / find unassigned areas in Get started; find disconnected fragments, population tooltip, demographic map in Refine and validate).
5. Same visibility rules as the branch: hidden in Super Draw, COI mode, and when not editing. Collapsible, not dismissible.
[...]

PR #683, opened 2026-07-31
Workflow buckets layout

### LINKED ISSUE #675: Launch 01 - Reorganize editor sidebar into workflow tabs (Population / Stats / Map Layers)
## Context

The editor sidebar currently renders five stacked accordion cards (District overview, Demographics, Elections, Validity check, Boundaries and areas). The `feat/draw-mode-simplified` branch prototyped a cleaner three-tab layout that we want to adopt — but that branch mixes in unrelated work (district meters redesign, brush controls, lock-painted mode), so **do not merge or work from it directly**. Reimplement only the tab restructure on `dev`, using the branch as reference.

## Requirements

1. Replace the stacked accordion in the sidebar with three tabs:
   - **Population** — `PopulationPanel` (districts mode only)
   - **Map Layers** — Boundaries and areas (`OverlaysPanel`), Demographics heatmap controls, Elections heatmap controls (districts mode only), and Map options (the current Visual settings popover content, `ToolSettings`)
   - **Stats** — Validity check (`MapValidation`), Demographics table, Elections table (whole tab districts-mode only)
   - Within Map Layers and Stats, sections are quiet collapsible headers (open by default), not bordered cards — see `DataLayerSection` in the reference branch.
2. Tab styling should match the static site's `SecondaryNav` treatment (centered text links, active = bold + districtrBlue, hover underline) rather than default Radix `Tabs.List` chrome.
3. In **Super Draw mode only**, add an icon-based toggle near the tabs that reverts the sidebar to the current stacked-panels layout. The choice persists across sessions (zustand `persist`).
4. COI (communities) mode keeps its current filtering behavior: districts-only tabs/sections are hidden, and a mode switch that hides the active tab falls back to the first visible one.
5. The mobile tab view derives its panel list from the same section registry (`SECTIONS` → `DataPanelUtils`) and must stay working.

## Suggested approach

- Reference implementation: `git show feat/draw-mode-simplified:app/src/app/components/sidebar/DataCards.tsx` — the `SECTIONS` registry, `DataLayersPanel`, `EvaluationPanel`, and `DataLayerSection` components map directly onto this issue (rename Data Layers → Map Layers, Evaluation → Stats).
- Main file: `app/src/app/components/sidebar/DataCards.tsx`. Keep the existing `AnimatedCollapse` helper.
- `ToolSettings` import for Map options: `app/src/app/components/Toolbar/Settings.tsx`; check what should happen to the existing Visual settings popover trigger once its contents move here.
- The branch also added `uiHintStore.sidebarTabRequest` for cross-panel tab jumps ("Find unassigned" → Stats tab). Issues 02 and 07 need this mechanism — port it as part of this issue.
- Stacked-panels toggle: new persisted flag in `mapControlsStore` (which already holds `sidebarPanels` accordion state — keep that state for the fallback layout); render the icon toggle adjacent to the tab list, gated on `useToolbarStore.superDraw`.
- Tab styles: borrow the class treatment from `app/src/app/components/Static/SecondaryNav.tsx` (those are `next/link`s, so extract styling only, not the component).

## Dependencies / sequencing

- **Land this first.** Issues 02, 05, 06, and 07 all touch files this issue restructures (`DataCards.tsx`, `Settings.tsx` contents, panel-opening hint actions).
- Issue 02 (helper box) and Issue 07 (population bars) target sidebar tabs from hint actions (`sidebarTabRequest`); they stub to accordion panels if this hasn't landed.
- Issue 05 removes three toggles from the Visual settings content that this issue relocates into Map Layers — coordinate so both removals happen against the same version.
- Issue 06 edits `withCoalition` usage inside `DataCards.tsx` — direct merge-conflict zone with this rewrite.
- The Table/Map segmented switchers on today's cards persist only in the stacked-cards fallback layout this issue keeps for Super Draw (see Issue 06, req 5).

## Acceptance criteria

- [x] Sidebar shows Population / Map Layers / Stats tabs with the section contents listed above; no stacked cards by default
- [x] Tabs visually match static-site SecondaryNav styling (bold + districtrBlue active state, hover underline)
- [x] Map options (visual settings) reachable inside Map Layers
- [x] In COI mode, only content appropriate to COI shows; no districts-only tabs; no crash when switching modes with a hidden tab active
- [x] In Super Draw, an icon toggle near the tabs switches back to the stacked-panels layout; setting survives reload; toggle absent in plain Draw
- [x] Mobile panel list still works
- [x] No other changes from `feat/draw-mode-simplified` (meters, brush controls, etc.) are pulled in

nofurtherinformation:
## Description
- Closes #675 

## Reviewers
- Primary: @fangge518 
- Secondary:

## Review required
- UX review
- Tab/card super draw logic


app/src/app/components/sidebar/DataCards.tsx
```diff
@@ -275,3 +282,230 @@ export const DataCards: React.FC = () => {
...
+const WORKFLOW_TABS: Array<{
+  key: WorkflowTabKey;
+  label: string;
+  content: React.ReactNode;
+  /** Hidden in communities (COI) mode. */
+  districtsOnly?: boolean;
+}> = [
+  {key: 'population', label: 'Population', content: <PopulationPanel />, districtsOnly: true},
+  {key: 'mapLayers', label: 'Map Layers', content: <MapLayersPanel />},
+  {key: 'stats', label: 'Stats', content: <StatsPanel />, districtsOnly: true},
```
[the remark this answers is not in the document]
nofurtherinformation: Fixed in 0c251bfc142c07eac23e8fa71fd657e525cb87fd

app/src/app/components/sidebar/WorkflowTabs.tsx
```diff
@@ -0,0 +1,224 @@
...
+  const clearSidebarTabRequest = useUiHintStore(state => state.clearSidebarTabRequest);
+  useEffect(() => {
+    if (sidebarTabRequest) {
+      setTab(sidebarTabRequest);
+      clearSidebarTabRequest();
+    }
+  }, [sidebarTabRequest, clearSidebarTabRequest]);
+  // A request issued while the tabs are unmounted (stacked layout, eval view)
+  // must not linger and fire a surprise jump on a later mount.
+  useEffect(() => clearSidebarTabRequest, [clearSidebarTabRequest]);
```
fangge518: I: The comment above this effect says a tab-jump request issued while the tabs are unmounted "must not linger and fire a surprise jump on a later mount." But this effect only clears the request in its cleanup (i.e., when `WorkflowTabs` unmounts) — it does nothing on mount. If a request arrives while the tabs are unmounted (stacked layout or eval view) and the tabs remount later, the *other* effect above it (whose dependency array includes `sidebarTabRequest`) fires on that mount exactly because the stale request is still truthy — producing the jump this effect was seemingly added to prevent. Not reachable today (`requestSidebarTab` has no caller yet in this PR), but #676/#681 are expected to wire one in, so this is worth fixing at the source before it ships load-bearing.

app/src/app/components/sidebar/WorkflowTabs.tsx
```diff
@@ -0,0 +1,224 @@
...
+import type {HelpTipKey} from '@components/HelpTip/helpTipContent';
+
+/** A quiet section header inside a workflow tab: no card chrome, just a
+ * heading row in the shared plane. Collapsible but open by default. */
+const TabSection: React.FC<{label: string; helpTip?: HelpTipKey; children: React.ReactNode}> = ({
+  label,
+  helpTip,
+  children,
+}) => {
+  const [open, setOpen] = useState(true);
```
fangge518: I: Each `TabSection` tracks `open` with local `useState(true)` (line 26). Since only the active tab's content is mounted (`{activeTab?.content}`), switching from e.g. Map Layers to Stats and back fully unmounts and remounts `MapLayersPanel` — resetting every section inside it back to open, discarding whatever the user just collapsed. The old accordion this replaces stored its open/closed state in `mapControlsStore.sidebarPanels`, which survived exactly this kind of remount. Defaulting to open matches the issue's spec; losing a deliberate collapse on every tab round-trip doesn't seem like it was a deliberate choice.

app/src/app/components/sidebar/DataCards.tsx
```diff
@@ -1,277 +1,44 @@
...
-        variant="surface"
+export const DataCards: React.FC = () => {
+  const superDraw = useToolbarStore(state => state.superDraw);
+  const stackedSidebar = useToolbarStore(state => state.stackedSidebar);
+  const setStackedSidebar = useToolbarStore(state => state.setStackedSidebar);
+  // The stacked preference is a Super Draw escape hatch; plain Draw always
+  // gets the workflow tabs, whatever a past Super Draw session persisted.
+  const stacked = superDraw && stackedSidebar;
+
+  const layoutToggle = superDraw ? (
```
fangge518, line 19: PP: In the tabbed layout, the toggle sits inline with the tab nav strip, sharing its bottom border. In the stacked layout, it's in its own row above `StackedPanels`, with no matching border — and `ToolbarInSidebar`'s Visual Settings popover appears only in this mode, adding an extra row above the toggle that the tabbed layout doesn't have. The net effect: switching between the two modes shifts the toggle's position and moves the visual divider between the toolbar area and the panel content. Suggest a fixed-position toggle, with Visual Settings moved to sit left of it on the same row in the stacked layout, so the divider between the upper (toolbar/settings/toggle) and lower (panel content) sections stays put across the switch.

app/src/app/components/sidebar/WorkflowTabs.tsx
```diff
@@ -0,0 +1,224 @@
...
+  // COI mode can leave a single visible tab; a one-tab strip is noise, but the
+  // Super Draw layout toggle must stay reachable.
+  const showStrip = visibleTabs.length > 1;
+
+  return (
+    <Flex direction="column" gap="2">
+      {showStrip ? (
+        <nav
+          aria-label="Sidebar panels"
+          className="border-b border-gray-200 py-2 bg-white overflow-x-auto"
```
fangge518: I: The sidebar scrolls internally (`Sidebar.tsx`'s `overflow-y-auto`/`ScrollArea`), and a long panel — the Stats tab's tables, in particular — can push the tab strip itself out of view, forcing a scroll back to the top to switch tabs. The static site's equivalent nav (`SecondaryNav.tsx`, `sticky top-16`) and the old Districtr header (`Header.tsx`, `sticky top-0`) both stay pinned during scroll — this nav's className (`border-b border-gray-200 py-2 bg-white overflow-x-auto`) otherwise closely mirrors `SecondaryNav`'s but drops the `sticky` positioning. Worth matching that precedent for consistency.

fangge518 (approved):
I've tried interacting with the workflow bucket and it is working as intended for the ordinary operations. Since this PR is what the other PR depends on, I will set a much higher threshold for blocking it, we can add other polish to it later when we work on other issues. But if any UI feels off, please feel free to tweak it until it feels right. Three I items and one PP item worth fixing before other work: a tab-jump guard that doesn't actually prevent the "surprise jump" it's meant to, and section-collapse state that resets every time you leave and return to a tab.

app/src/app/components/sidebar/WorkflowTabs.tsx
nofurtherinformation: Cleaned up in 637f1dbf7cf65827b7788592472425a491b18914

app/src/app/components/sidebar/WorkflowTabs.tsx
nofurtherinformation: Reverted to the zustand store based approach in a0f0471b3174d9f72277262a5e419fb5e0035dce -- this was definitely a negative UX reversion

app/src/app/components/sidebar/DataCards.tsx
nofurtherinformation: I've tweaked this a bit, and we'll also continue to iterate with upcoming issues

PR #685, opened 2026-07-31
Helper box

### LINKED ISSUE #676: Launch 02 - Draft-status-driven helper box (Get started → Refine and validate → Advanced)

nofurtherinformation:
## Description
- Closes #676 

## Reviewers
- Primary: @fangge518 
- Secondary:

## Review required
- UX meets expectations
- Metadata business logic is sound
- No incorrect/incomplete data is displayed

### DIFF
--- a/app/src/app/components/Topbar/MapStatus.tsx
+++ b/app/src/app/components/Topbar/MapStatus.tsx
@@ -62,10 +62,10 @@ export const MapStatusButtons: React.FC<{
   draftStatus: DraftStatus | null;
   onChange: (draftStatus: DraftStatus) => Promise<void>;
 }> = ({draftStatus, onChange}) => {
-  // Forward moves are earned: statuses whose criteria aren't met yet can't be
-  // selected (backward moves always can). Same criteria the helper box shows.
-  // This Radix version's SegmentedControl.Item has no disabled prop, so lock
-  // by ignoring the selection and dimming the item.
+  // Forward moves are earned: statuses whose criteria aren't met can't be
+  // selected (backward moves always can). This Radix version's
+  // SegmentedControl.Item has no disabled prop, so lock by ignoring the
+  // selection and dimming the item.
   const {statusLocked} = useDraftStatusCriteria();
   return (
     <SegmentedControl.Root
--- a/app/src/app/components/Topbar/MapStatus.tsx
+++ b/app/src/app/components/Topbar/MapStatus.tsx
@@ -10,7 +10,6 @@ import {
 } from '@constants/document/draftStatus';
 import {InProgressIcon, ScratchWorkIcon, ReadyIcon} from './Icons';
 import {ACCESS_STATES} from '@constants/document/state';
-import {useDraftStatusCriteria} from '@/app/hooks/useDraftStatusCriteria';
 
 const statusIcons: Record<DraftStatus, React.FC> = {
   [DRAFT_STATUSES.SCRATCH]: ScratchWorkIcon,
@@ -58,46 +57,22 @@ export const MapStatus: React.FC<{
   );
 };
 
+/** Free choice by design: the draft status is the user's own label for their
+ * plan. The only earned move is the helper box's advance button. */
 export const MapStatusButtons: React.FC<{
   draftStatus: DraftStatus | null;
   onChange: (draftStatus: DraftStatus) => Promise<void>;
 }> = ({draftStatus, onChange}) => {
-  // Forward moves are earned: statuses whose criteria aren't met can't be
-  // selected (backward moves always can). This Radix version's
-  // SegmentedControl.Item has no disabled prop, so lock by ignoring the
-  // selection and dimming the item.
-  const {statusLocked} = useDraftStatusCriteria();
   return (
-    <SegmentedControl.Root
-      value={draftStatus as string}
-      onValueChange={value => {
-        if (!statusLocked(value as DraftStatus)) onChange(value as DraftStatus);
-      }}
-      size="2"
-    >
-      {DRAFT_STATUS_ORDER.map(status => {
-        const locked = statusLocked(status);
-        const item = (
-          <SegmentedControl.Item
-            key={status}
-            value={status}
-            aria-disabled={locked}
-            className={locked ? 'opacity-50 !cursor-not-allowed' : ''}
-          >
-            <Flex direction="row" gap="2" align="center" justify="start">
-              {statusIcons[status]({})}
-              <Text>{DRAFT_STATUS_TEXT[status]}</Text>
-            </Flex>
-          </SegmentedControl.Item>
-        );
-        return locked ? (
-          <Tooltip key={status} content="Complete the helper checklist to unlock this status">
-            {item}
-          </Tooltip>
-        ) : (
-          item
-        );
-      })}
+    <SegmentedControl.Root value={draftStatus as string} onValueChange={onChange} size="2">
+      {DRAFT_STATUS_ORDER.map(status => (
+        <SegmentedControl.Item key={status} value={status}>
+          <Flex direction="row" gap="2" align="center" justify="start">
+            {statusIcons[status]({})}
+            <Text>{DRAFT_STATUS_TEXT[status]}</Text>
+          </Flex>
+        </SegmentedControl.Item>
+      ))}
     </SegmentedControl.Root>
   );
 };
--- a/app/src/app/components/Topbar/MapStatus.tsx
+++ b/app/src/app/components/Topbar/MapStatus.tsx
@@ -81,6 +81,7 @@ export const MapStatusButtons: React.FC<{
           <SegmentedControl.Item
             key={status}
             value={status}
+            aria-disabled={locked}
             className={locked ? 'opacity-50 !cursor-not-allowed' : ''}
           >
             <Flex direction="row" gap="2" align="center" justify="start">
--- a/app/src/app/components/Topbar/MapStatus.tsx
+++ b/app/src/app/components/Topbar/MapStatus.tsx
@@ -10,6 +10,7 @@ import {
 } from '@constants/document/draftStatus';
 import {InProgressIcon, ScratchWorkIcon, ReadyIcon} from './Icons';
 import {ACCESS_STATES} from '@constants/document/state';
+import {useDraftStatusCriteria} from '@/app/hooks/useDraftStatusCriteria';
 
 const statusIcons: Record<DraftStatus, React.FC> = {
   [DRAFT_STATUSES.SCRATCH]: ScratchWorkIcon,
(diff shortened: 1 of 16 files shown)


app/src/app/components/sidebar/MapValidation/MapValidation.tsx
```diff
@@ -27,12 +28,33 @@ export const MapValidation = () => {
...
+  // Opening the check (or swapping panels) silently saves pending edits so
+  // the results reflect the current map — helper-box jumps land on fresh
+  // numbers, without the map-lock overlay or saved toast (this also runs when
+  // the Stats tab merely opens with the section expanded). Deliberately keyed
+  // on the panel, not isOutdated: painting while the panel is open must not
+  // trigger a save per stroke.
+  useEffect(() => {
+    if (isOutdated) handlePutAssignments(false, {silent: true});
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [activePanel]);
```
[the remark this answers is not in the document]
nofurtherinformation: Resolved in 3d9c2f9228f09c0e46276319fa03464593703631

app/src/app/components/sidebar/Sidebar.tsx
```diff
@@ -103,6 +104,7 @@ export default function SidebarComponent() {
       </div>
       <Flex direction="column" gap="3" className="size-full">
         <div className="flex flex-col gap-3 pr-3">
+          <DraftStatusHelper />
```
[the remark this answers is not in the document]
nofurtherinformation: Added in 3d9c2f9228f09c0e46276319fa03464593703631

app/src/app/components/sidebar/MapValidation/MapValidation.tsx
```diff
@@ -27,12 +28,36 @@ export const MapValidation = () => {
...
+  // the results reflect the current map — helper-box jumps land on fresh
+  // numbers, without the map-lock overlay or saved toast (this also runs when
+  // the Stats tab merely opens with the section expanded). Also keyed on the
+  // IDB document's arrival: it loads async, so the mount-time run sees
+  // isOutdated=false and would otherwise miss the opening save. Deliberately
+  // not keyed on isOutdated itself: painting while the panel is open must not
+  // trigger a save per stroke.
+  const idbLoaded = !!idbDocument;
+  useEffect(() => {
+    if (isOutdated) handlePutAssignments(false, {silent: true});
```
fangge518, line 63: I: The new effect (`if (isOutdated) handlePutAssignments(false, {silent: true})`, keyed on `activePanel`/`idbLoaded`) fires whenever this panel mounts or its sub-tab changes, with no check of the viewer's access level. `MapValidation` sits under the Stats tab, which any read-only/shared-link viewer can open — not gated behind editing the way the draft-status helper box is. If `isOutdated` happens to be true for such a viewer, this triggers a PUT-assignments attempt on their behalf. The backend doesn't enforce edit access on that endpoint either, so at best it's a harmless no-op write; at worst (a stale local timestamp vs. a concurrently-edited document) it could surface the save-conflict UI to someone who never asked to save anything.

app/src/app/components/sidebar/DraftStatusHelper.tsx
```diff
@@ -0,0 +1,457 @@
...
+        unassigned !== undefined && unassigned > 0
+          ? [
+              !inBlockView &&
+              unassignedRatio !== undefined &&
+              unassignedRatio > ROUGH_DRAW_UNASSIGNED_RATIO
+                ? {
+                    label: 'Paint by counties to roughly draw districts',
+                    onClick: handleCountyBrushHint,
+                  }
+                : {label: 'Find unassigned areas', onClick: () => openValidation('Completeness')},
```
fangge518, line 217: I: When nearing the end of "Get started," the "Find unassigned areas" hint jumps straight to the Stats tab's Completeness section. The destination is correct, but the jump itself is a single instant cut — a user might not immediately register how they got there. Similarly for the contiguity check in the second stage. Worth sequencing the jump instead of firing it all at once: pulse the "Stats" tab label briefly, then switch to it and pulse "Completeness" (or the section header), then land on the existing highlight for the completeness content itself. The `flash`/`flashTarget` mechanism in `uiHintStore.ts` already supports pulsing arbitrary targets, so this would be a sequencing change on top of what's already there rather than new plumbing.

app/src/app/components/Topbar/MapStatus.tsx
```diff
@@ -57,6 +57,8 @@ export const MapStatus: React.FC<{
   );
 };
 
+/** Free choice by design: the draft status is the user's own label for their
```
fangge518, line 60: PP: When the helper box moves `draft_status` forward or back (advance button, step-back link), the topbar's status icon (`MapStatus.tsx`'s `IconButton`) changes silently. Worth reusing the same pulse-highlight already built for hint targets (`uiHintStore.ts`'s `flash`/`flashTarget`, `.ui-flash`) on that icon when the change lands — just highlighting the icon itself should be enough, and it doubles as a teaching moment for users who haven't noticed it's clickable.

app/src/app/components/sidebar/DraftStatusHelper.tsx
```diff
@@ -0,0 +1,457 @@
...
+      : currentStatus === DRAFT_STATUSES.IN_PROGRESS
+        ? 1
+        : 2;
+  const criteriaStage = !scratchDone ? 0 : !inProgressDone ? 1 : 2;
+  const displayStage = Math.min(statusStage, criteriaStage);
+  const regressed = displayStage < statusStage;
+  const previousStatus: DraftStatus =
+    statusStage === 2 ? DRAFT_STATUSES.IN_PROGRESS : DRAFT_STATUSES.SCRATCH;
+
+  const title = ['Get started', 'Refine and validate', 'Advanced'][displayStage];
```
fangge518: PP: Once a plan reaches Ready to share and the user collapses the card after exploring its pointer links, returning later shows only a collapsed header reading "Advanced" — with no visible content, that label alone may not clearly convey what's inside. Consider relabeling this stage "Helpful Links" (or similar) so the collapsed header stays self-explanatory on its own.

fangge518 (commented):
Three I and one PP below — nothing blocking.

app/src/app/components/sidebar/MapValidation/MapValidation.tsx
nofurtherinformation: Good catch - access guarded in d319e17b8bdc76e1c7aa7e1ebbab81a6157782cc

app/src/app/components/sidebar/DraftStatusHelper.tsx
nofurtherinformation: Added a nicer sequence of jumps and flash hints in d1bee8695553246a670a9f4181c8f614b8f784d3. I suspect we could refine this more, but this should give us a reasonable starting point

app/src/app/components/Topbar/MapStatus.tsx
nofurtherinformation: Agreed -- added a status badge flash in d319e17b8bdc76e1c7aa7e1ebbab81a6157782cc and refined the flash in d1bee8695553246a670a9f4181c8f614b8f784d3

app/src/app/components/sidebar/DraftStatusHelper.tsx
nofurtherinformation: I landed on "Advanced plan evaluation" -- I think most users _could_ safely ignore the features here, but we could iterate on this text

PR #688, opened 2026-07-31
Segment control migration

### LINKED ISSUE #680: Launch 06 - Choropleth controls: radio buttons, no icons, no presets; coalition to Super Draw only; data citations
## Context

The demographic/election map-layer panel (`MapPanel.tsx`) uses icon-adorned segmented controls for display mode, plus a second segmented control of overlay presets (Overlay / Show Thematic Map / Show Districts) with an `x`-key cycle hotkey. Fine-grained opacity sliders exist but are Super-Draw-only. The coalition builder is exposed in plain Draw. No data-source citations appear anywhere.

## Requirements

1. **Display mode control** (None / Comparison / Overlay / Sized circles): remove the icons and swap the segmented control for radio buttons. (Comparison stays Super-Draw-only as today.) Still need to explore different options for displaying layer control to reduce clutter within each accordion panel.
2. **Remove the overlay presets and hotkey entirely** (all modes — Draw and Super Draw): delete the Overlay / Show Thematic Map / Show Districts segmented control, the `x`-cycle keyboard listener, and its tooltip. Replace with the overlay-opacity slider in Draw mode; Super Draw keeps its existing two sliders (overlay + districts layer).
3. **Coalition is Super Draw only**: hide the "Create a coalition" expander and the SummaryPanel coalition section in plain Draw (COI unchanged). No coalition entry in the choropleth variable list in plain Draw either, unless a coalition was already configured in Super Draw — engineer's judgment on whether existing coalition variables stay selectable; must not crash if a saved map has one active.
4. **Data citations** (hardcoded for now, all maps same source): small gray text under the demographics/elections tables and under the choropleth legend:
   - Demographics: "Source: U.S. Census Bureau, 2020 Decennial Census (P.L. 94-171)"
   - Elections: "Source: Dave's Redistricting App"
5. The Table/Map switchers on the demographics/elections cards are **not** part of this swap — they largely disappear with Issue 01's tabs, but persist in the stacked-cards fallback layout.

## Suggested approach

- `app/src/app/components/sidebar/MapPanel.tsx`: `mapDisplayModes` → drop `icon`; render `RadioGroup` instead of `SegmentedControl`.
- Delete `getOpacityStates`, the presets block, and the `x`-hotkey effect. Check `overlayMemory` / Visual-settings toggle interactions in `Settings.tsx` — they read `showPaintedDistricts`/`overlayOpacity`, which the presets used to set; make sure removing presets leaves sane values (`showPaintedDistricts: true`, slider-driven opacity).
- Move the overlay-opacity slider out of the `superDraw` guard; keep the districts-layer slider Super-Draw-only.
- Coalition gating: `withCoalition` in `DataCards.tsx` and the coalition section in `SummaryPanel.tsx` get a `superDraw` condition (`useToolbarStore`); same for `coalitionOptionFor` in the variable list if hiding it in Draw.
- Citations: a tiny `<Text size="1" color="gray">` component with the two hardcoded strings, rendered at the bottom of the evaluation-table section and after the legend in `MapPanel`; keyed off `columnGroup === 'VOTERHISTORY'`.

## Dependencies / sequencing

- **After Issue 01**: the coalition gating touches `DataCards.tsx`, which Issue 01 rewrites (`withCoalition` moves into its `TabbedSummaryPanel`/registry). Landing this first means redoing the gating in the rewrite.
- `MapPanel.tsx` changes are independent of everything else.

## Acceptance criteria

- [ ] Display mode is icon-free radio buttons; Comparison only in Super Draw
- [ ] No overlay presets anywhere; pressing `x` does nothing; no dangling listener
- [ ] Draw: single overlay-opacity slider works; Super Draw: both sliders work
- [ ] Plain Draw shows no coalition builder/expander; Super Draw unchanged; maps with an existing coalition don't break in Draw
- [ ] Census citation under demographics table + legend; DRA citation under elections table + legend
- [ ] Table/Map card switchers untouched by this issue

nofurtherinformation:
## Description
- Closes #680 

## Reviewers
- Primary: @fangge518 
- Secondary:

## Checklist

## Review required
- UX review primarily
- Code here is simple refactors and layout

### DIFF
--- a/app/src/app/components/Shared/PovSwitcher.tsx
+++ b/app/src/app/components/Shared/PovSwitcher.tsx
@@ -1,5 +1,5 @@
 'use client';
-import {Flex, SegmentedControl, Text} from '@radix-ui/themes';
+import {Flex, RadioGroup, Text} from '@radix-ui/themes';
 
 /** Partisan point of view for election tables/metrics. */
 export type Pov = 'dem' | 'rep';
@@ -14,9 +14,11 @@ export const PovSwitcher: React.FC<{
     <Text size={labelSize} color="gray">
       Point of view
     </Text>
-    <SegmentedControl.Root size="1" value={pov} onValueChange={v => setPov(v as Pov)}>
-      <SegmentedControl.Item value="dem">Democratic</SegmentedControl.Item>
-      <SegmentedControl.Item value="rep">Republican</SegmentedControl.Item>
-    </SegmentedControl.Root>
+    <RadioGroup.Root size="1" value={pov} onValueChange={v => setPov(v as Pov)}>
+      <Flex direction="row" align="center" gapX="3" wrap="wrap">
+        <RadioGroup.Item value="dem">Democratic</RadioGroup.Item>
+        <RadioGroup.Item value="rep">Republican</RadioGroup.Item>
+      </Flex>
+    </RadioGroup.Root>
   </Flex>
 );
--- a/app/src/app/components/Toolbar/Settings.tsx
+++ b/app/src/app/components/Toolbar/Settings.tsx
@@ -1,13 +1,5 @@
 import React from 'react';
-import {
-  Heading,
-  CheckboxGroup,
-  Flex,
-  Button,
-  Text,
-  Select,
-  SegmentedControl,
-} from '@radix-ui/themes';
+import {Heading, CheckboxGroup, Flex, Button, Text, Select, RadioCards} from '@radix-ui/themes';
 import {type BasemapId, BASEMAP_IDS} from '@/app/constants/map/layerStyle';
 import {useFeatureFlagStore} from '@store/featureFlagStore';
 import {useMapStore} from '@store/mapStore';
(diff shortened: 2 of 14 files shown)


fangge518:
I have interacted with the new UI layout for a bit. I really like how clean the draw mode's stats table currently is, which I take is the most important aim of this PR. Here are some general thoughts on the other parts of the new UI, perhaps less urgent then cleaning up the draw mode:

General issues:
1. The coalition builder was in current dev's Draw mode of COI maps, but not in this PR's head. Coalition builder is probably more useful for a community mapper than a districter, so we might want to keep this in the Draw mode of COI maps.

2. The style of the current tab selectors are really sharp. My first reaction when seeing these was not to interpret them as tab labels, but after clicking on them several times I can easily form this association. For internal consistency, would it be sensible to use a similar style as our current tab selectors used elsewhere in the site, such as "Population/Stats/Map layers" or the tab selectors of our secondary navigation menu?

3. Some layers settings that are a bit repetitive across different panels election/demographics and across different tabs within those panels, especially when they are stacked just near each other in the "Map Layers" of the workflow buckets. While we are still exploring, here is another proposal: I think we can unify all these options into one section, perhaps called "Additional data layer", there are four selections stacked here. Data meaning: demographic or election; Data display mode: None/Comparison/Overlay/Sized Circles; Demographic variable: Black/White/...; Election variable: 2016 Pres/2020 Pres..  Additional settings, including demographic layer opacity, election layer opacity, district layer opacity and data format can be put into the current gear-shaped settings menu. The gradient legends seem to belong to the corner of the map window better as these need to be compared more readily with the map color, and are useful even when the users switch to other Populations or Stats. These four selections will always be visible when "Additional data layer" is expanded, but are conditionally greyed out (e.g., other three settings greyed out when Data display is None and election variable greyed out when Data meaning is demographic.)

Nits:
1. This might be a preexisting issue: the settings panel for toggling "Show data as percent" is titled "Choropleth map settings" even when it is controlling the sized circle data. Perhaps "Additional layer setting" or simply remove the title?

4. Currently Superdraw-->Stats--> demographics (in either layout)--> settings near Table summary type: the options are still slider style toggle. Should we change this to radio buttons to be consistent?

nofurtherinformation:
Thanks for this feedback! I've integrated a few updates to further clarify, and I think this is at a good starting point.

PR #689, opened 2026-07-31
Add "Disallow Paint-over" brush toggle

### LINKED ISSUE #677: Launch 03 - "Disallow paint over" toggle
## Context

`feat/draw-mode-simplified` added a "Lock painted" toggle: a boolean brush filter that protects **all** assigned areas — paint only lands on unassigned land, erasing still works. We want this on `dev`, extracted from the branch (don't merge the branch itself).

Separately, dev already has district-level locks (`mapOptions.lockPaintedAreas`, per-district padlocks in the Population panel + `ZoneLockPicker`), which protect *specific chosen districts*. Two similarly-named "lock" features would be confusing, so the district-level feature is renamed to **Freeze**.

## Requirements

1. **Extract the toggle** from the branch and rename it **"Disallow paint-over"**:
   - This label is a bit passive, so we still need to discuss the exact name further. Alternatives: Allow paint-over, Paint empty area only, etc.
   - Text-only button — no lock (or any) icon.
   - Rendered with the brush controls (visible when the brush tool is active), districts mode + editing only, disabled on read-only access.
   - Behavior: with the toggle on, the brush cannot repaint any assigned area; the eraser (and block painting inside broken units) is unaffected.

## Suggested approach

- Extraction points from the branch (`git show feat/draw-mode-simplified:<path>`):
  - `app/src/app/components/Toolbar/LockPaintedToggle.tsx` — new component (drop the icons, update label/tooltip)
  - `app/src/app/store/types.ts` — `lockAssignedAreas?: boolean` on MapOptions (consider naming it `disallowPaintOver` since it's new here)
  - `app/src/app/store/mapControlsStore.tsx` — default `false`
  - `app/src/app/utils/map/filterFeatures.ts` — the branch's brush-only filter block (guards: districts mode + `ACTIVE_TOOLS.BRUSH`, filters `f => !zoneAssignments.get(id)`)
  - `BrushControls.tsx` — render site

## Dependencies / sequencing

- **Before Issue 05**: its right-column key options include this toggle (Issue 05 leaves a slot if this hasn't landed). Both edit `BrushControls.tsx`.

## Acceptance criteria

- [ ] "Disallow paint over" text-only toggle appears with brush controls in districts edit mode; hidden in COI/view modes; disabled when read-only
- [ ] Toggle on: brushing over any assigned area does nothing; brushing unassigned works; eraser works regardless
- [ ] Toggle state coexists correctly with district locks (both filters can be active)
- [ ] Per-district lock and lock-all still function unchanged behaviorally

fangge518:
Closes #677. Adds a session-only toggle that, while on, restricts the brush to unassigned areas — no district's already-painted area can be repainted, whichever district it belongs to.

**Fix:** a new `disallowPaintOver` flag on `mapOptions` (default off, no persistence — matches sibling `mapOptions` fields), checked in `filterFeatures.ts` only when the active tool is the brush and the map is in districts mode. It composes with the existing per-district `lockPaintedAreas` filter rather than replacing it — both can be active together. The eraser and block-painting inside broken units go through a different tool value and are untouched.

The toggle itself is a checkbox+label, matching `PaintByCounty`'s pattern except smaller and without a card to emphasize it. Hidden entirely in COI/community mode, and disabled for read-only viewers.

### DIFF
--- a/app/src/app/components/HelpTip/helpTipContent.ts
+++ b/app/src/app/components/HelpTip/helpTipContent.ts
@@ -60,6 +60,12 @@ export const helpTipContent = {
     videoFile: 'county_brush.webm',
     guideAnchor: 'drawing-the-districts',
   },
+  // No video for this one yet — text-only, unlike the other brush tips.
+  disallowPaintOver: {
+    title: 'Disallow paint over',
+    text: 'Toggle this to protect every district you\'ve already drawn — the brush will only add unassigned units, leaving painted areas untouched. Erasing still works.',
+    guideAnchor: 'drawing-the-districts',
+  },
   switchDistrict: {
     title: 'Switching districts',
     text: 'Click a color to switch which district you’re painting. For plans with many districts, use the dropdown to find the one you want.',
--- a/app/src/app/components/HelpTip/helpTipContent.ts
+++ b/app/src/app/components/HelpTip/helpTipContent.ts
@@ -63,7 +63,7 @@ export const helpTipContent = {
   // No video for this one yet — text-only, unlike the other brush tips.
   disallowPaintOver: {
     title: 'Disallow paint over',
-    text: 'When toggled on, the brush will only add unassigned units, leaving painted areas untouched.',
+    text: 'When enabled, the brush will only add unassigned units, leaving painted areas untouched.',
     guideAnchor: 'drawing-the-districts',
   },
   switchDistrict: {
--- a/app/src/app/components/HelpTip/helpTipContent.ts
+++ b/app/src/app/components/HelpTip/helpTipContent.ts
@@ -63,7 +63,7 @@ export const helpTipContent = {
   // No video for this one yet — text-only, unlike the other brush tips.
   disallowPaintOver: {
     title: 'Disallow paint over',
-    text: 'Toggle this to protect every district you\'ve already drawn — the brush will only add unassigned units, leaving painted areas untouched. Erasing still works.',
+    text: 'When toggled on, the brush will only add unassigned units, leaving painted areas untouched.',
     guideAnchor: 'drawing-the-districts',
   },
   switchDistrict: {
--- a/app/src/app/components/Toolbar/DisallowPaintOver.tsx
+++ b/app/src/app/components/Toolbar/DisallowPaintOver.tsx
@@ -1,0 +1,48 @@
+import {Card, Text} from '@radix-ui/themes';
+import {useMapStore} from '@/app/store/mapStore';
+import {useMapControlsStore} from '@/app/store/mapControlsStore';
+import {ACCESS_STATES} from '@constants/document/state';
+import {HelpTip, HELP_TIP_HOVER_DELAY} from '@components/HelpTip/HelpTip';
+
+export default function DisallowPaintOver() {
+  const disallowPaintOver = useMapControlsStore(state => state.mapOptions.disallowPaintOver);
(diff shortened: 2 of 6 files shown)


app/src/app/components/Toolbar/DisallowPaintOver.tsx
```diff
@@ -0,0 +1,39 @@
...
+    <HelpTip tip="disallowPaintOver" openDelay={HELP_TIP_HOVER_DELAY}>
+      <Text as="label" size="1" className="cursor-pointer select-none">
+        <Flex gap="1" align="center">
+          <Checkbox
+            size="1"
+            checked={!!disallowPaintOver}
+            onCheckedChange={handleToggle}
+            disabled={disabled}
+          />
+          Disallow Paint-over
```
nofurtherinformation:
```suggestion
          Disallow paint-over
```

app/src/app/components/Toolbar/DisallowPaintOver.tsx
```diff
@@ -0,0 +1,39 @@
...
+      disallowPaintOver: !disallowPaintOver,
+    });
+  };
+
+  return (
+    <HelpTip tip="disallowPaintOver" openDelay={HELP_TIP_HOVER_DELAY}>
+      <Text as="label" size="1" className="cursor-pointer select-none">
+        <Flex gap="1" align="center">
+          <Checkbox
+            size="1"
```
nofurtherinformation:
```suggestion
            size="2"
```

app/src/app/components/Toolbar/DisallowPaintOver.tsx
```diff
@@ -0,0 +1,39 @@
...
+  const handleToggle = () => {
+    if (disabled) return;
+    setMapOptions({
+      disallowPaintOver: !disallowPaintOver,
+    });
+  };
+
+  return (
+    <HelpTip tip="disallowPaintOver" openDelay={HELP_TIP_HOVER_DELAY}>
+      <Text as="label" size="1" className="cursor-pointer select-none">
```
nofurtherinformation:
```suggestion
      <Text as="label" size="2" className="cursor-pointer select-none">
```

app/src/app/components/HelpTip/helpTipContent.ts
```diff
@@ -60,6 +60,12 @@ export const helpTipContent = {
     videoFile: 'county_brush.webm',
     guideAnchor: 'drawing-the-districts',
   },
+  // No video for this one yet — text-only, unlike the other brush tips.
+  disallowPaintOver: {
+    title: 'Disallow paint over',
+    text: 'When toggled on, the brush will only add unassigned units, leaving painted areas untouched.',
```
nofurtherinformation: Nit: Since we aren't using a toggle, we can use slightly more general language

```suggestion
    text: 'When enabled, the brush will only add unassigned units, leaving painted areas untouched.',
```

nofurtherinformation (approved):
Works as expected and code is clean! 

A few small nits / clean up -- I think size 2 might be better.

One bigger thing to think about, but I think is OK for now: if this is enabled, break mode is essentially disabled when you break into a set of blocks. But, this is a super draw interaction and easily remedied by disabling 'Disallow paint-over'

fangge518:
Reply to Dylan's suggestions: all the nits are applied. I haven't touched the disallow paint-over button when breaking, since we might want to keep this option on especially when breaking an empty block, but I agree that having this check on when breaking a painted unit might be confusing and may need a more intelligent solution.

PR #690, opened 2026-07-31
County brush: multi-county selection, default-on, layout

### LINKED ISSUE #678: Launch 04 - County brush: multi-county selection, brush-size aware, on by default for new multi-county maps
## Context

The county brush currently queries the `counties_fill` layer at the exact cursor point, takes only the first county hit, and ignores brush size entirely (`getFeaturesIntersectingCounties.ts`). Each mousemove also re-runs a full arquero table scan per county (`demographyService.getFiltered(fips)`), which is wasteful when repeated. It defaults to off, and its checkbox sits to the right of the brush-size slider.

## Requirements

1. **Brush-size-aware multi-county selection**: the brush size defines a capture box/radius around the cursor (same `boxAroundPoint` convention as `getFeaturesInBbox`); **every** county intersecting it is painted in full. Sweeping with a large brush can capture several counties in one event.
2. **Performant**: no visible lag while sweeping across multiple counties. Repeatedly resolving the same county must not re-scan the table.
3. **Default-on for new maps**: county brush starts enabled when a user opens a **newly created** map whose region spans **more than one county**. Existing/returning maps keep the current default (off). If the user turns it off, it stays off for that session.
4. **Districts mode only** (unchanged from current gating; still hidden for LOCAL map types and while breaking a unit into blocks).
5. **Layout**: move the county brush control to the **left** of the brush-size slider in `BrushControls`.

## Suggested approach

- `app/src/app/utils/map/getFeaturesIntersectingCounties.ts`: replace `queryRenderedFeatures(e.point, …)` with the bbox from `boxAroundPoint(e, brushSize)` (`bboxAroundPoint.ts`); collect the **distinct** `STATEFP + COUNTYFP` values from all returned county features, concatenate their feature lists, and run one `filterFeatures` pass over the deduped union (`fastUniqBy` is already used elsewhere for this).
- Performance: memoize `getFiltered(fips)` results in `demographyService` (a `Map<fips, features>` cleared whenever the table reloads). Also skip recomputation when the set of counties under the brush hasn't changed since the last event.
- Default-on: on document load, if the map is newly created (arriving from map creation flow) and the parent layer spans >1 distinct county FIPS (derivable from the loaded demography table paths, or the counties source), set `paintByCounty: true` and the county paint function. Wire wherever new-document initialization happens in `mapStore`/`mapControlsStore`; session-only, no persistence.
- Layout: swap the order of `BrushSizeSelector` and `PaintByCounty` in `BrushControls.tsx` and re-check the `mt-3` alignment hack.

## Dependencies / sequencing

- **Before Issue 05** (or fold the layout change into it): Issue 05 rebuilds `BrushControls` into a two-column scaffold whose left column orders county brush before the slider. If Issue 05 lands first, requirement 5 here is already satisfied — drop it and only ship the behavior changes. Both editing `BrushControls.tsx` concurrently will conflict.
- Otherwise independent.

## Acceptance criteria

- [ ] With a large brush, one stroke near county borders paints all intersected counties; small brush behaves like today (single county under cursor)
- [ ] No dropped frames / event-loop stalls while sweeping across counties on a statewide map (verify with a large state, e.g. TX)
- [ ] Same county re-entered repeatedly does not re-run the table filter (cache hit)
- [ ] Newly created multi-county map: county brush checked on first load; single-county map: unchecked; reopened existing map: unchecked
- [ ] User unchecking it mid-session is respected until reload
- [ ] County brush control renders left of the brush-size slider
- [ ] Locked/frozen districts and disallow-paint-over filtering still respected (`filterFeatures` still applied)

fangge518:
Closes #678. The county brush only painted whichever county sat under the exact cursor point, so a brush that straddled a county line silently dropped the far side.

**Fix:** `getFeaturesIntersectingCounties` now boxes the cursor by brush size (the same convention `getFeaturesInBbox` already uses), queries every county intersecting that box, and unions their block features through one `filterFeatures` pass. Unioning counties means re-querying demography per county on every mousemove, so `demographyService.getFiltered` now caches its result per county id (invalidated whenever the underlying table reloads), and the paint function itself skips recomputation when the brushed county set hasn't changed since the last mousemove.

Also: `paintByCounty` now defaults on for a blank, multi-county map — off for DC (the only one of districtr's 52 map-module jurisdictions that's a single county-equivalent) and off for any map the user has already started painting, so opening an in-progress map never silently flips this under them. Computed once per document load in `useDocumentWithSync`, from the document's `statefps` and whether its loaded assignments are empty. Still gated by the `paintCounties` feature flag, still a session-only preference — unchecking it holds for the rest of the session. The county-brush toggle now sits to the left of the brush-size slider.

## Tested

- [x] Brushing across a county line paints both counties' blocks, not just one
- [x] A blank multi-county map (e.g. a fresh Texas map) loads with county brush checked
- [x] A blank DC map loads with county brush unchecked
- [x] Reopening a map with existing assignments loads with county brush unchecked
- [x] Unchecking county brush holds for the rest of the session

### DIFF
--- a/app/src/app/hooks/useDocumentWithSync.tsx
+++ b/app/src/app/hooks/useDocumentWithSync.tsx
@@ -14,6 +14,19 @@ import {MAP_MODES} from '@constants/map/mode';
 import {MAP_TYPES} from '@constants/document/types';
 import {APP_LOADING_STATES} from '@constants/document/state';
 
+// DC is the only one of districtr's 52 map-module jurisdictions (50 states + DC + PR)
+// that's a single county-equivalent (itself) — every other state/PR has multiple
+// counties/municipios, so "multi-county" is the correct default everywhere else.
+const DC_STATEFP = '11';
+
+/** County brush defaults on for a blank multi-county map, off for DC (nothing to
+ * multi-select) and off for a map the user has already started painting — flipping
+ * it out from under an in-progress edit would be surprising, not helpful. */
+function defaultPaintByCounty(statefps: string[] | undefined, hasAssignments: boolean): boolean {
+  if (hasAssignments) return false;
+  return !statefps?.includes(DC_STATEFP);
+}
+
 interface UseDocumentWithSyncOptions {
   document_id: string | null | undefined;
   enabled?: boolean;
@@ -37,6 +50,7 @@ export function useDocumentWithSync({
   const setAppLoadingState = useMapStore(state => state.setAppLoadingState);
   const setLoadingState = useMapStore(state => state.setLoadingState);
   const mapMode = useMapControlsStore(state => state.mapMode);
+  const setMapOptions = useMapControlsStore(state => state.setMapOptions);
   const ingestDistrictFromDocument = useAssignmentsStore(state => state.ingestFromDocument);
   const ingestCoiFromDocument = useCoiAssignmentsStore(state => state.ingestFromDocument);
   const districtResolveConflict = useAssignmentsStore(state => state.resolveConflict);
@@ -139,9 +153,21 @@ export function useDocumentWithSync({
         if (isCommunityDocument) {
           const data = formatCoiAssignmentsFromDocument(result.response.assignments);
           ingestCoiFromDocument(data, result.response.document, result.response.hasLocalEdits);
+          setMapOptions({
+            paintByCounty: defaultPaintByCounty(
+              result.response.document.statefps,
+              data.communityAssignments.size > 0
+            ),
+          });
         } else {
           const data = formatAssignmentsFromDocument(result.response.assignments);
           ingestDistrictFromDocument(data, result.response.document, result.response.hasLocalEdits);
+          setMapOptions({
+            paintByCounty: defaultPaintByCounty(
+              result.response.document.statefps,
+              data.zoneAssignments.size > 0
+            ),
+          });
         }
         // Set overlays from document response
         setMapDocument(result.response.document);
--- a/app/src/app/hooks/useDocumentWithSync.tsx
+++ b/app/src/app/hooks/useDocumentWithSync.tsx
@@ -14,19 +14,6 @@ import {MAP_MODES} from '@constants/map/mode';
 import {MAP_TYPES} from '@constants/document/types';
 import {APP_LOADING_STATES} from '@constants/document/state';
 
-// DC is the only one of districtr's 52 map-module jurisdictions (50 states + DC + PR)
-// that's a single county-equivalent (itself) — every other state/PR has multiple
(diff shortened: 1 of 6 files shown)


app/src/app/hooks/useDocumentWithSync.tsx
```diff
@@ -14,6 +14,19 @@ import {MAP_MODES} from '@constants/map/mode';
 import {MAP_TYPES} from '@constants/document/types';
 import {APP_LOADING_STATES} from '@constants/document/state';
 
+// DC is the only one of districtr's 52 map-module jurisdictions (50 states + DC + PR)
+// that's a single county-equivalent (itself) — every other state/PR has multiple
+// counties/municipios, so "multi-county" is the correct default everywhere else.
+const DC_STATEFP = '11';
```
nofurtherinformation: H: There are some other edge cases where there will only be one county -- eg. Indiana local towns: https://beta.districtr.org/map/48474

It might make sense to on demographic data population data load do a quick sanity check of the demographic store IDs (if not too heavy):
- Grab all the ID columns
- Slice to the first 5 digits of the GEOIDs
- Count the number of unique county FIPS codes and if >1, enable county brush

This could also be done in the parquet worker as it builds the columnar dataset

app/src/app/utils/map/getFeaturesIntersectingCounties.ts
```diff
@@ -5,12 +5,25 @@ import {
...
 import {demographyService} from '../demography/demographyService';
 
+/**
+ * Module-scoped memo of the last computed result, keyed by the sorted set of
+ * county FIPS codes under the brush. Consecutive mousemove events over the
+ * same set of counties (the common case while dragging) skip the
+ * demography lookup + filterFeatures pass entirely.
+ */
+let lastCountyKey: string | null = null;
+let lastResult: MapGeoJSONFeature[] | undefined;
```
nofurtherinformation, line 20: Nice work: This implementation has good perf and works acceptably even under worst case conditions (eg. Texas, 4x CPU slowdown)! 

I'm not able to reasonably come up with edge cases that would cause this to be stale (eg stale locked districts) and bypass filters. And, even if so, the failure case is a quick undo action.

nofurtherinformation (changes requested):
Nice work! Very concise approach and works well. One moderate suggestion on safeguarding against single-county maps, but otherwise LGTM!

app/src/app/hooks/useDocumentWithSync.tsx
fangge518: Good catch! I've implemented the more thorough check for one county. Instead of counting all distinct counties, it breaks whenever it finds the second distinct counties. This runs extremely fast even for the largest states (<1ms).

PR #691, opened 2026-07-31
Constant two-column layout for tool controls

### LINKED ISSUE #679: Launch 05 - Tool controls: two-column layout, stable height, unified help tip
## Context

The area under the tool buttons (`ToolControls`) mounts a different component per tool — pan shows nothing, brush/eraser show `BrushControls`, break shows controls only in block view, inspector its own panel. Height changes on every tool switch, causing layout jumps. Display toggles that matter while painting (district numbers, population tooltip) are buried in the Visual settings popover.

## Requirements

1. **T-shaped Two-column layout** under the toolbar when the sidebar is wide enough (container query on the resizable sidebar; stack to one column when narrow):
   - **Top**: Always one column, county brush, brush size slider  (county brush left of the slider, per Issue 04)
   - **Left**: District (zone) picker, this column should be ~2/3 of the total width and display color picker pips in a small grid as needed
   - **Right**: key option toggles — District numbers (`showZoneNumbers`), Population tooltip (`showPopulationTooltip`), Disallow paint over (from Issue 03; leave a slot until it lands)
2. **Remove those toggles from the Visual settings popover** — they live only in the new right column (no duplication).
3. **Constant vertical spacing across move / paint / erase / break**: the full scaffold renders for all four tools; irrelevant controls are disabled/grayed, not unmounted:
   - Move (pan): slider + district picker disabled; key options remain usable
   - Erase: district picker disabled (districts mode); slider active
   - Break before entering block view: same scaffold as paint
   - **Inspector keeps its own distinct layout** (allowed to differ in height)
4. **Unified help tip**: move/paint/erase/break/ inspector share **one** combined `HelpTip` on hover anywhere over those three buttons in draw mode and in super-draw mode; undo/redo keep their individual tips.

## Suggested approach

- `app/src/app/components/Toolbar/ToolControls/ToolControls.tsx`: collapse the per-tool config for pan/brush/eraser/shatter into one shared scaffold component with a `disabled` map derived from `activeTool` (keep `InspectorControls` as its own branch). Getting rid of the conditional mounting is what kills the height jumps — verify no hotkey/listener assumptions break (`ZonePicker` etc. currently assume they only mount when applicable).
- Layout: the sidebar already sets `containerType: 'inline-size'` (`Sidebar.tsx`), so use a `@container` CSS breakpoint for the column split.
- Key options: extract the three checkbox items from `app/src/app/components/Toolbar/Settings.tsx` into a small shared component rendered in the right column; delete them from the popover's checkbox group.
- Unified tip: wrap the pan/brush/eraser button trio in a single `HelpTip` for draw mode and wrap all five main tools in a single `HelpTip` for superdraw mode.  (the undo/redo pair already does exactly this — same pattern). Drop the separate pan/brush/eraser entries. Make dummy videos with the correct video name on S3 (e.g., by copying from other clips), all clips will be updated and overridden before launch, so having the correct video name is good for now.

## Dependencies / sequencing

- **After Issue 03** (Disallow paint over provides the third right-column toggle) and **after Issue 04** (county brush ordering; if this issue lands first, Issue 04 drops its layout requirement). All three edit `BrushControls.tsx`.
- **Coordinate with Issue 01**: Issue 01 moves the remaining Visual settings content into the Map Layers tab. The toggle removal here must land against whichever version of that content exists — agree on order so the two edits to `Settings.tsx` don't collide.

## Acceptance criteria

- [ ] Wide sidebar: T shaped layout with brush slider and county brush toggle above, then controls in two columns with ~2/3 - 1/3 width layout (paint controls left, key options right); narrow sidebar: stacked, nothing clipped
- [ ] Switching between move/paint/erase/break causes no vertical shift of the sidebar content below; disabled controls are visibly grayed and inert
- [ ] District numbers / Population tooltip toggles work from the right column and no longer appear in Visual settings
- [ ] Inspector still shows its own controls
- [ ] Hovering any of move/paint/erase shows the single combined tip; break, inspector, undo/redo tips unchanged
- [ ] Tool hotkeys and eraser/zone-picker behavior in COI mode unchanged

### LINKED ISSUE #682: Launch 08 - HelpTip: rename video link to "Quick demonstration"
## Context

The hover HelpTip shows a link that opens the video clip modal, currently labeled "See demonstration ▸" (`app/src/app/components/HelpTip/HelpTip.tsx`, ~line 254).

## Requirements

Change the link text to **"Quick demonstration ▸"** (keep the arrow glyph and behavior unchanged).

## Suggested approach

One-line string change in `HelpTip.tsx`. Check for a matching aria-label or test assertion referencing the old text.

## Dependencies / sequencing

None. (Issue 05 adds a new combined help tip entry but doesn't touch this string.)

## Acceptance criteria

- [ ] HelpTip video link reads "Quick demonstration ▸"; modal opens as before
- [ ] No stale "See demonstration" strings remain (grep)

fangge518:
Closes #679, closes #682. `ToolControls.tsx` used to mount a different component tree per active tool — nothing for pan, `BrushControls` for paint/erase, and (before entering block view) a literal empty render for break, on the reasoning that there's nothing to paint yet. Each behavior lived in its own branch, and controls like `ZonePicker` were written assuming they'd only ever mount when applicable.

**Fix:** one `ToolControlsScaffold` now mounts unconditionally for pan, paint, erase, and break (both before and after entering block view) — same structure every time, a `disabled` calculation per tool grays out the pieces that don't apply (pan disables the brush-size slider and district/zone picker; erase disables the picker only in Districts mode; break disables both before a unit is broken, matching paint's full controls once it is). Inspector keeps its own separate layout, untouched. The scaffold lays out as a top row (county-brush checkbox left of the brush-size slider, matching #678's ordering), then a `@container`-queried two-column row below it — district/zone picker on the left, three key-option toggles (district numbers, population tooltip, and a disabled placeholder for #677's "disallow paint over") on the right, stacking to one column under a narrow sidebar.

Since the zone/district picker now mounts even when visually disabled, its digit-key hotkey (previously gated only by whether the component mounted at all) needed its own gate: `ColorPicker` and `CoiPicker` take a `disabled` prop that suppresses the keydown handler, threaded down from the scaffold.

The district-numbers and population-tooltip toggles moved out of `Settings.tsx`'s Visual Settings popover entirely, into a `KeyOptionToggles` component rendered only in the scaffold's right column. Nothing loses access to them: the toolbar (and this scaffold with it) is always mounted above the sidebar, independent of whether the sidebar itself shows the tabbed layout or the stacked-panels/Super Draw fallback, so the popover never needed its own copy.

Also merges the five per-tool `HelpTip` entries (pan, paint, erase, break, inspector) into two shared ones — `drawToolsCombination` (pan/paint/erase) and `superdrawToolsCombination` (pan/paint/erase/break/inspector) — using the same one-entry/many-triggers pattern already used for undo/redo. Each tool button still renders its own `HelpTip` instance but overrides the shared entry's text to empty, leaving only the demonstration-video link.

Folds in #682: the demonstration link's label changes from "See demonstration ▸" to "Quick demonstration ▸", fixed across every tip. `HelpTipEntry` gains an optional `linkSuffix` — plain, non-clickable text trailing after the link on the same line — so only "Quick demonstration ▸" itself is ever part of the click target, never a longer descriptive phrase. The two tool-group combos read "Quick demonstration ▸ on how to combine tools efficiently"; undo/redo reads "Quick demonstration ▸ on how to undo/redo". Undo/redo's hover card now also drops its description sentence in plain Draw mode (the same text="" suppression the tool-group combos already used), leaving just the link — the sentence still appears in the expanded video modal's description, it just no longer duplicates above the link in the hover card.

County Brush (`PaintByCounty.tsx`) joins the same combined-tools demonstration too, dropping its own dedicated `countyBrush` HelpTip entry/video in favor of `drawToolsCombination`/`superdrawToolsCombination` — same text-suppression pattern as the other buttons. Its "unavailable while breaking a unit into blocks" state still overrides the hover text, and now explicitly hides the link via a new `HelpTip` prop, `hideLink`, instead of the old mechanism (any non-empty override text silently hid the link). That mechanism change is also what lets undo/redo's Super Draw shortcut text keep the demonstration link visible alongside it — previously any override text hid the link unconditionally, which was wrong for undo/redo (the shortcuts supplement the same video, they aren't a different situation) even though it was right for County Brush's break-mode message.

County Brush and the Population tooltip toggle are both now also disabled during the Move (pan) tool, matching the brush-size slider and zone picker — neither does anything while just panning.

## Tested

- [X] Switching between pan/paint/erase/break shows the same layout with only the relevant controls enabled
- [X] Break shows disabled controls before entering block view, and live ones after
- [X] District numbers / population tooltip toggles appear only in the scaffold's right column — not in the Visual Settings popover, not in the Map Layers tab's "Map options" section
- [X] Hovering pan/paint/erase (draw) or the super-draw tool group shows only "Quick demonstration ▸ on how to combine tools efficiently", no separate description text, and only "Quick demonstration ▸" itself is clickable
- [X] Hovering undo/redo in plain Draw mode shows only "Quick demonstration ▸ on how to undo/redo"; in Super Draw mode it shows the chorded shortcuts as text, with the same link still visible below them
- [X] Hovering County Brush shows the same combined "Quick demonstration ▸ on how to combine tools efficiently" link as the other basic tools, except while breaking a unit into blocks, where it shows "Unavailable while breaking a unit into blocks" with no link
- [X] Two-column layout collapses to one column when the sidebar is narrow
- [X] County Brush and Population tooltip are both grayed out and non-interactive during the Move (pan) tool; District numbers stays enabled
- [X] `bun run ts` — clean


app/src/app/components/Toolbar/DistrictZonePicker.tsx
nofurtherinformation: I: Tangential to this PR, let's only show the District count in Super Draw (since only then the number of districts is modifiable)

app/src/app/components/Toolbar/ToolControls/ToolControlsScaffold.tsx
```diff
@@ -0,0 +1,95 @@
...
+      </Flex>
+
+      <TwoColumnGrid>
+        <Box style={disabledSectionStyle(zonePickerDisabled)}>
+          <ZonePicker disabled={zonePickerDisabled} />
+        </Box>
+        <KeyOptionToggles />
+      </TwoColumnGrid>
+
+      {paintConstraint && (
```
nofurtherinformation: Q: Tangential to this PR, but should we refactor this to follow the same pattern as the break pill on the map?

app/src/app/components/Toolbar/ToolControls/ToolControlsScaffold.tsx
```diff
@@ -0,0 +1,95 @@
...
+        <Box className="flex-grow" style={{flexGrow: 1, ...disabledSectionStyle(sliderDisabled)}}>
+          <BrushSizeSelector />
+        </Box>
+      </Flex>
+
+      <TwoColumnGrid>
+        <Box style={disabledSectionStyle(zonePickerDisabled)}>
+          <ZonePicker disabled={zonePickerDisabled} />
+        </Box>
+        <KeyOptionToggles />
```
nofurtherinformation, line 78: Q: In Super Draw, what do you think about moving the sidebar layout toggle and visual settings button (when using stacked panels) to the bottom of this area? 

That might condense the space slightly more

app/src/app/components/HelpTip/helpTipContent.ts
```diff
@@ -72,19 +69,19 @@ export const helpTipContent = {
...
     title: 'District overview',
-    text: 'Click here to expand the district overview, listing each drawn district — click a number there to select it and switch the brush to that color, and see each district’s population against the ideal target.',
-    videoFile: 'district_overview.webm',
-    guideAnchor: 'district-overview',
+    text: 'Expand here to monitor the district populations as you draw.',
   },
   districtLock: {
     title: 'Locking districts',
-    text: 'Toggle the lock icon next to a district to protect it from being painted over while you work on other districts.',
+    text: 'Toggle it protect this district from being painted over or erased.',
```
nofurtherinformation:
```suggestion
    text: 'Toggle it to protect this district from being painted over or erased.',
```

nofurtherinformation (approved):
Looks good! One typo and a couple tangents we could clean up while working on these parts of the app, but nothing blocking.

app/src/app/components/Toolbar/ToolControls/ToolControlsScaffold.tsx
fangge518: I think this is the right direction, but might need some further attention on how it interact with inspector, and how the toggle fits in the default layout without visual setting. Let's do this in a future change.

app/src/app/components/Toolbar/ToolControls/ToolControlsScaffold.tsx
fangge518: Done.

app/src/app/components/Toolbar/DistrictZonePicker.tsx
fangge518: This is added.

app/src/app/components/HelpTip/helpTipContent.ts
fangge518: Fixed.

PR #692, opened 2026-07-31
Fix stale completeness check via React Query, keyed like Contiguity

fangge518:
The Stats tab's Completeness check (unassigned-area finder) could show leftover results from an earlier fetch even after a map was fully assigned. `unassignedFeatureBboxes` only refreshed on three ad hoc triggers: initial mount (guarded by a `hasFoundUnassigned` flag that lives in a store outliving the component, so it never re-fires on remount), a `document_id` change, or an `updated_at` change tracked via a `useRef` that resets fresh on every remount — so it couldn't detect a save that landed while the component was unmounted. Reopening the panel, or returning to it after a save happened elsewhere, could leave stale unassigned-area chips on an actually-complete map with no signal they might be wrong.

The sibling Contiguity check doesn't have this problem: it uses `@tanstack/react-query`, keyed on `[document_id, updated_at]`, so a key change refetches regardless of mount history.

**Fix:** Move the unassigned-features fetch onto the same `useQuery` pattern, keyed identically (`['UnassignedFeatures', documentIdParam, updated_at]`). `unassignedFeatures.ts` is trimmed to just the `selectedIndex` UI state the query doesn't own — the fetch, cache, and loading state (`hasFoundUnassigned`/`isLoading`) now live entirely in the query.

## Tested

- [x] Type check (`bun run ts`)
- [x] Assign all population, save, open Stats → Completeness — shows "No unassigned areas found"
- [x] Un-assign an area, save while the panel stays open — chip list picks up the change without leaving and returning
- [x] Re-assign it, save, navigate away and back to Completeness — shows the current (complete) state on remount (the case that was broken)
- [x] Read-only shared-link view — the `public_id`-based fetch path still resolves

### DIFF
--- a/app/src/app/components/sidebar/MapValidation/ZoomToUnassigned.tsx
+++ b/app/src/app/components/sidebar/MapValidation/ZoomToUnassigned.tsx
@@ -1,56 +1,50 @@
 import {useSummaryStats} from '@/app/hooks/useSummaryStats';
 import {useMapStore} from '@/app/store/mapStore';
 import {useMapControlsStore} from '@/app/store/mapControlsStore';
+import {useAssignmentsStore} from '@/app/store/assignmentsStore';
 import {useUnassignFeaturesStore} from '@/app/store/unassignedFeatures';
+import GeometryWorker from '@/app/utils/GeometryWorker';
 import {formatNumber} from '@/app/utils/numbers';
 import {Checkbox, Flex, Text} from '@radix-ui/themes';
-import React, {useEffect, useRef} from 'react';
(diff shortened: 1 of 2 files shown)


nofurtherinformation (approved):
Let's wait to merge this until after the launch feedback, but-- LGTM!

Nice integration of react query and net negative LOC :) 

I had a reaction against react query a while back, we had tried to use it for everything including mutations, and it was pretty fussy with the level of complexity in some of our server interactions. But, this usage makes sense and is very clean. Thank you!

PR #694, opened 2026-08-03
Feat/launch staging internal misc

nofurtherinformation:
* Center population bars horizontally when over max width
* Overlay slider clips the right edge
* Flip unassigned areas / population preview text
* Get started
   * “Move to” instead of “mark”
   * Mark as in progress could be bigger/more prominent
   * Add icon indicating state for move to based on existing 
   * Add save now top level scent for contiguity as needed
* Make manual save not flash the screen, just use the autosave behavior but with dfferent text


fangge518 (approved):
Everything is working as intended. And the code looks good to me.

Two nitpicks:
1. In the HelpBox, I prefer to have "Save now" and "Find disconnected components" shown in one sentence: "link{Save now} and link{find disconnected components}". I hope this would suggest to user that these are two steps of one operation. There is also a risk that user might think clicking it once is enough and confused after clicking "Save now" that nothing salient is occurring.

2. I think "Move to In Progress" is syntactically as unclear as "Mark as In Progress". How about "Mark the plan as "In Progress"" and "Mark the plan as "Ready to Share""?

PR #695, opened 2026-08-03
Post-#691/#690 fixes: key-option toggles, county brush, help-tip copy

fangge518:
Small fixes found while reviewing the #690/#691 batch after merge.

- The three key-option checkboxes ("Only paint unassigned areas", "Show district numbers", "Show population on hover") were visually misaligned and vaguely labeled. Relabeled and consolidated into one place (previously "Disallow paint-over" duplicated a dead-code path in Inspector's own controls that never actually rendered).
- County brush could show checked on a blank multi-county map (e.g. Montana) while still only painting single units — its on/off flag and its actual paint behavior were separate state that only a manual toggle kept in sync. Now synced automatically.
- County brush stayed clickable on single-county and local maps even though toggling it did nothing there; now disabled outright. Connecticut is disabled outright too, regardless of county count — its county boundaries no longer correspond to the FIPS codes districtr's own map data uses.
- The draft-status helper's "Show population tooltips as you paint" hint pointed at a sidebar tab section that no longer holds that toggle; now points at the toggle's real location.
- Removed a duplicate "release paint mask" button left over in Inspector's controls — the floating map pill already covers that state everywhere.
- Trimmed the demographics/elections help-tip copy: both mentioned on-map configuration that doesn't exist in the sidebar section where they actually show.
- The Demographics/Elections help tips were shared, word-for-word, between the Map Layers tab (a toggle list) and the Stats tab (a data table) — same text, same demonstration link, in two contexts that don't need the same thing. Split into separate tips: Map Layers keeps the link, drops the text; Stats keeps the text, drops the link. The stacked layout's own combined Demographics/Elections section (which merges what these two tabs split apart) still gets both.


nofurtherinformation (approved):
LGTM!

We _could_ opt to hide the county brush entirely, but I like this approach to start

PR #708, opened 2026-08-05
Rewrite guide page for the tabbed sidebar

fangge518:
The guide's "Main Tools" and "Data Panels" sections still described the old sidebar accordion layout — panel names, a "toggle the lock icon" workflow, and a combined Demographics/Elections panel — none of which match the current tabbed sidebar (Population/Stats/Map Layers tabs, a "Forbid paint-over" checkbox instead of a lock icon).

**Fix:** rewrote both sections to match the tabbed sidebar — "Data Panels" is now "Data Tabs" (Population, Stats, Map Layers), with the validation/table content moved under Stats and the display-mode/map-display toggle controls consolidated under Map Layers. Recorded four new clips for the restructured sections and re-recorded the guide's existing clips for a consistent look. Synced the in-page table of contents and every subheading's anchor id to match, and updated `helpTipContent.ts`'s `guideAnchor`/`videoFile` fields so the toolbar's own "View in guide" and "Quick demonstration" links stay consistent with the rewrite.

## Tested

- [x] Guide page renders with the rewritten content, plays the new clips, and its TOC/anchor links land on the right sections (local manual check)


nofurtherinformation (approved):
LGTM -- I think this is a good update and makes the guide more accurate!

